PR #2020, opened 2021-04-05
test(Select): refactor non native select tests

teimurjan:
[FX-1687](https://toptal-core.atlassian.net/browse/FX-1687)

### Description

### How to test

### Screenshots

### Review

<details>
<br />

List of available commands:

</details>


packages/picasso/src/YearSelect/__snapshots__/test.tsx.snap
```diff
@@ -111,6 +115,7 @@ exports[`YearSelect render in descending order 1`] = `
       </div>
     </li>
     <li
+      aria-checked="false"
```
OleksandrNechai, line 118: What caused this to appear?

teimurjan: I added this attribute to `Select` so it's comfortable to query the checked value in tests.

packages/picasso/src/NonNativeSelect/test.tsx
```diff
@@ -97,31 +102,12 @@ describe('NonNativeSelect', () => {
...
-
-    const selectInput = getByPlaceholderText(placeholder)
-
-    fireEvent.focus(selectInput)
-
-    expect(queryByRole('menu')).not.toBeInTheDocument()
-  })
-
-  it('should open menu when click on select if there is NO search', () => {
+  it('opens menu on select click', () => {
```
OleksandrNechai, line 105: Where is a click? :-)

cypress/component/Select.spec.tsx
```diff
@@ -122,35 +86,169 @@ const OPTIONS = [
...
-  it('focuses Select with and without a search', () => {
-    mount(<SelectSearchBehaviourExample />)
+  it('renders', () => {
+    mount(
+      <TestingPicasso>
+        <TestSelect />
+      </TestingPicasso>
+    )
+
+    cy.get('body').happoScreenshot()
```
denieler, line 103: why do we use `body` selector all the time to make a screenshot? isn't it faster to compare smaller screenshots?

cypress/component/Select.spec.tsx
```diff
@@ -160,11 +258,16 @@ describe('Select', () => {
       .and('match', /true/)
   })
 
-  it('renders as non-transparent on a colorful background', () => {
+  it('sets background correctly', () => {
```
denieler, line 261:
```suggestion
  it('sets background correctly to various select states', () => {
```

packages/picasso/src/NonNativeSelectOption/NonNativeSelectOption.tsx
```diff
@@ -31,6 +31,7 @@ const NonNativeSelectOption = React.memo(
     return (
       <MenuItem
         role='option'
+        aria-checked={selected}
```
denieler, line 34: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role, https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaChecked

it says `aria-checked` is a checkbox role. We already have `aria-selected` here

packages/picasso/src/NonNativeSelectOption/test.tsx
```diff
@@ -40,4 +40,42 @@ describe('NonNativeSelectOption', () => {
...
+      selected: true,
+      highlighted: true,
+      option: { text: 'Option #1', value: 'option1' }
+    })
+
+    const option = getByRole('option')
+
+    expect(option.getAttribute('aria-checked')).toEqual('true')
+    expect(option.getAttribute('aria-selected')).toEqual('true')
+    expect(option.getAttribute('value')).toEqual('option1')
```
denieler, line 55: do we really need those `expects` if we want to use snapshot in this case?

packages/picasso/src/Select/story/index.jsx
```diff
@@ -32,116 +32,44 @@ page.createTabChapter('Props').addComponentDocs({
 
 page
   .createChapter()
-  .addExample('Select/story/Default.example.tsx', {
-    title: 'Default',
-    effect: async (testPage, makeScreenshot) => {
-      const hideInputCaretStyle = `
-        input {
-          caret-color: transparent !important;
```
denieler, line 40: don't happo tests have issue with the input caret?

OleksandrNechai (approved):
LGTM! :smiley: 

I must express concern but this is just my experience, not a best practice or truth.

All software design is based on trying to depend on something that changes less often. I believe tests should also depend on something that rarely changes.

In our case, the most stable thing is `Select`'s interface. And I would test only against it and avoid testing some parts of Select or any other implementation details. Because this reduces our ability to maneuver/refactor later.

packages/picasso/src/Select/story/index.jsx
```diff
@@ -32,116 +32,44 @@ page.createTabChapter('Props').addComponentDocs({
...
-        const mouseOverEvent = new MouseEvent('mouseover', {
-          view: window,
-          bubbles: true,
-          cancelable: true
-        })
-
-        lastHalfVisibleOption.dispatchEvent(mouseOverEvent)
-      })
-
-      // the options list should slightly scroll to show the hovered option
```
denieler, line 68: I haven't seen the test, which would replicate this one removed. But I think this is important to have

packages/picasso/src/YearSelect/__snapshots__/test.tsx.snap
OleksandrNechai: I hoped you did this because you cared about accessibility... What happened with `testid` approach?

packages/picasso/src/Select/story/index.jsx
teimurjan: Nope

cypress/component/Select.spec.tsx
OleksandrNechai: How about `looks fine on a colorful background`? :smile:

packages/picasso/src/NonNativeSelectOption/test.tsx
teimurjan: We discussed it earlier when I was working on `Accordion` tests. As far as I remember, we decided to combine such expectations with snapshots to have a clearer view of the element's attributes.

packages/picasso/src/NonNativeSelectOption/NonNativeSelectOption.tsx
teimurjan: You're right but `aria-selected` shows the highlighted status of an option. Should we use the data attribute for that?

cypress/component/Select.spec.tsx
teimurjan: It looks like it's a standard in our codebase. If you render the only element in `body`, you receive a small screenshot with this element without any extra space.

packages/picasso/src/NonNativeSelectOption/test.tsx
denieler: if you, as a creator of the test, trust the test more - 👍

packages/picasso/src/YearSelect/__snapshots__/test.tsx.snap
teimurjan: I need to query a selected option. `testid` can't be dynamic.

packages/picasso/src/NonNativeSelect/test.tsx
teimurjan: Here: https://github.com/toptal/picasso/blob/1483c8fc9a499d59d259284ae9f744252df3f8ba/packages/picasso/src/NonNativeSelect/test.tsx#L115

OleksandrNechai: :man_facepalming:

packages/picasso/src/NonNativeSelectOption/NonNativeSelectOption.tsx
denieler: I think `aria-selected` already do that job, isn't it? we need to read probably docs and understand how to use it better if you are saying that we mark now as selected only highlighted option. I'm sure docs describe this behaviour

cypress/component/Select.spec.tsx
denieler: My thoughts behind it, that if we Happo will be too expensive for us and we will need to replace it - it might be very difficult to replace correctly `body` selectors with a new setup

packages/picasso/src/NonNativeSelectOption/NonNativeSelectOption.tsx
teimurjan: I will check it out.

cypress/component/Select.spec.tsx
```diff
@@ -78,13 +78,18 @@ const SelectNativeExample = () => {
   )
 }
 
-const OPTIONS = [
-  { value: '1', text: 'Option 1' },
-  { value: '2', text: 'Option 2' },
-  { value: '3', text: 'Option 3' },
-  { value: '4', text: 'Option 4' },
-  { value: '5', text: 'Option 5' }
-]
```
OleksandrNechai, line 123: This was easier to understand to me. What if you set this array and then `getOptions` would take first `n` from this array?

cypress/component/Select.spec.tsx
```diff
@@ -86,85 +78,237 @@ const SelectNativeExample = () => {
...
-    </Container>
-  </TestingPicasso>
-)
+const getOptions = (count: number) =>
+  Array(count)
+    .fill(0)
+    .map((_, index) => ({
+      value: (index + 1).toString(),
+      text: `Option ${index + 1}`
+    }))
```
denieler, line 87: nit: what about `[...Array(count).keys().map(i => `Option ${i}`)]`? 😁

cypress/component/Select.spec.tsx
```diff
@@ -86,85 +78,237 @@ const SelectNativeExample = () => {
...
+      const option = window.document.querySelector('[role="option"][value="8"]')
+      const mouseOverEvent = new MouseEvent('mouseover', {
+        view: window,
+        bubbles: true,
+        cancelable: true
+      })
+
+      option?.dispatchEvent(mouseOverEvent)
+
+      return cy.get('body').happoScreenshot()
```
denieler, line 157: I think we need more hints to test what it is about. The previous test even opening after half a year or even later we still understand what was going on there, but with this one, it might be problematic.

For example:
```
[role="option"][value="8"]
```
is not clear why.
Also, the name of the test don't really describe what is going on inside it

cypress/component/Select.spec.tsx
teimurjan: I understand your warning, but I don't see difficulties replacing selectors when we are over the limit. I wouldn't say I like premature optimization, so I propose to see what's next before changing anything.

cypress/component/Select.spec.tsx
teimurjan: @OleksandrNechai IMO `sets background correctly to various select states` has more meaning. 🙂

packages/picasso/src/NonNativeSelectOption/NonNativeSelectOption.tsx
teimurjan: @denieler, I got rid of `aria-checked` because there is no matching attribute for an option that is the value of select.

packages/picasso/src/Select/story/index.jsx
teimurjan: ✅

cypress/component/Select.spec.tsx
teimurjan: 👍

cypress/component/Select.spec.tsx
teimurjan: > This was easier to understand to me. What if you set this array and then getOptions would take first n from this array?

I'll use @OleksandrNechai's proposal to improve the readability if you're OK with that. 🙂

cypress/component/Select.spec.tsx
teimurjan: Refactored the test. Check it out, please. 🙂

cypress/component/Select.spec.tsx
denieler: in each test then we would need to replace selectors. Replacing them to the same thing - not difficult. But if you would need to identify and selector wrapper individually per each test in more than 100 test files - this will be very difficult task

cypress/component/Select.spec.tsx
```diff
@@ -86,85 +78,239 @@ const SelectNativeExample = () => {
...
+      </TestingPicasso>
+    )
+
+    openSelect()
+
+    // cy.get & cy.trigger scroll to the element so custom scrolling behaviour can't be tested
+    // using window to avoid cypress auto scrolling
+
+    // eslint-disable-next-line promise/catch-or-return
+    cy.window().then(window => {
```
denieler, line 143: can't we `await` here the `window` object? so the test will remain without nesting levels?

teimurjan: It's not a `Promise`. It's `Cypress.Chainable`, so we can't

packages/picasso/src/NonNativeSelect/__snapshots__/test.tsx.snap
```diff
@@ -311,25 +131,19 @@ exports[`NonNativeSelect should render description 1`] = `
           <div
             class="PicassoContainer-rightxsmallMargin PicassoContainer-flex PicassoContainer-inline PicassoMenuItem-iconContainer"
           />
-          <span
-            class="PicassoMenuItem-stringContent PicassoMenuItem-stringContentMedium"
+          <div
+            data-testid="custom-option-3"
```
denieler, line 135: where are those `custom-option-x` coming from?

teimurjan: https://github.com/toptal/picasso/blob/b8616d08a97e024d9bf5cd3aa74c5ab93d261516/packages/picasso/src/NonNativeSelect/test.tsx#L326

cypress/component/Select.spec.tsx
denieler: 😞

denieler (approved):

cypress/component/Select.spec.tsx
teimurjan: Happo charges for a screenshot (no matter is it body or a single element). Knowing that I don't think we should change anything. At least in this PR.

denieler: yes, but as I said I meant the future, when Happo might become too expensive for us or will not fit our needs regarding functionality or anything else

PR #2048, opened 2021-04-26
feat(Tag): add styles for tag content overflow

s0ber:
### Description

The `min-width: 0` rule allows the content not to stretch beyond the width of the parent component. Without this rule we have no way to prevent the tag from stretching (and though applying additional layout-based styles like `text-overflow: ellipsis` or using the `TypographyOverflow` component).

### Screenshots

**Before**
Снимок экрана 2021-04-26 в 19 37 57

**After**
Снимок экрана 2021-04-26 в 19 37 48

### Review

<details>
<br />

List of available commands:

</details>


packages/picasso/src/Tag/styles.ts
```diff
@@ -5,7 +5,8 @@ import '../Chip/styles'
 export default ({ palette }: Theme) =>
   createStyles({
     root: {
-      fontSize: '1rem'
+      fontSize: '1rem',
+      maxWidth: '100%'
```
s0ber, line 9: I can remove this rule as it can be applied on a client side.

packages/picasso/src/Tag/styles.ts
```diff
@@ -30,7 +31,8 @@ export default ({ palette }: Theme) =>
     },
     innerLabel: {
       fontSize: '0.75rem',
-      fontWeight: 600
+      fontWeight: 600,
+      minWidth: 0
```
s0ber, line 35: Alternatively we can use `width: 100%`

packages/picasso/src/Tag/styles.ts
havenchyk: @s0ber could you please clarify why is this needed?

havenchyk:
Two points: 

- please explain why max-width on the root is needed
- I think this example is not okay, what do you think?
image

packages/picasso/src/Tag/styles.ts
havenchyk: I think it will be a breaking change. Can you add 'fullWidth' prop instead?

packages/picasso/src/Tag/styles.ts
havenchyk: minWidth is fine

havenchyk (changes requested):
Please remove width: 100% from the root and add a new property instead on Tag to handle such case.

s0ber:
@havenchyk it's not `width: 100%`, but `max-width: 100%`, it will prevent the component from stretching past the parent component boundaries. But I can remove it completely and add it on a client side as a `css` prop.

s0ber:
@havenchyk thank you for taking a look! I'll push a fix!

havenchyk (approved):
My bad, misread the code :D

packages/picasso/src/Tag/styles.ts
havenchyk: this is fine

s0ber:
@havenchyk I think I'll leave the `max-width` property because we use it almost everywhere where we use tags. It's hard to imagine the case where we would like the tag to stretch past the layout boundary.

packages/picasso/src/Tag/story/Overflown.example.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Container, Tag } from '@toptal/picasso'
+import { TypographyOverflow } from '@toptal/picasso-lab'
+
+const Example = () => (
+  <Container style={{ width: '500px' }}>
+    <Tag variant='grey'>
+      <TypographyOverflow inline weight='semibold'>
+        Loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong
+      </TypographyOverflow>
```
OleksandrNechai, line 10: To me, his example does not demonstrate the capabilities of the component but rather tests its boundaries. Therefore the better place for this example would be in Cypress tests + Happo.

packages/picasso/src/Tag/styles.ts
OleksandrNechai: @havenchyk are you sure that messing with the width will not cause problems with reusing this component?

havenchyk: I checked at codesandbox and it worked fine by me. Any specific example of when it can cause the problem?

OleksandrNechai: Well, it will work. But is it reusable? Is it a concern of a component to set its width? This [presentation](https://www.youtube.com/watch?v=XR6eM_5pAb0) always inspires me, I find ideas very appealing.

OleksandrNechai (changes requested):
I reject because of example. I think it should be rather test, overflow does not really demonstrate features of the component

s0ber:
@OleksandrNechai do you have any other visual tests in Picasso apart the ones that are checking the storybooks?

OleksandrNechai:
@s0ber hey! Yes. We use Happo. Have a look at Happo tested example here https://github.com/toptal/picasso/blob/master/cypress/component/Accordion.spec.tsx

OleksandrNechai (approved):

PR #2054, opened 2021-04-28
feat: [FX-1872] Add showCustom for useNotification

denieler:
[FX-1872]

### Description

Add the opportunity to show Custom Notifications via the `useNotifications` hook.

<details>
<br />

List of available commands:

</details>

[FX-1872]: https://toptal-core.atlassian.net/browse/FX-1872


packages/picasso/src/Notification/Notification.tsx
```diff
@@ -47,7 +47,7 @@ export interface PrivateProps
   elevated?: boolean
 }
 
-export type PublicProps = Omit<PrivateProps, 'variant' | 'icon'>
+export type PublicProps = Omit<PrivateProps, 'elevated' | 'icon'>
```
denieler, line 50: it looks like the creator of this `PublicProps` wanted to have 3 props to be private: `variant`, `icon`, and `elevated` based on https://github.com/toptal/picasso/blob/230ab5f1b5a89250b68fe2ecc851c2541f675c7f/packages/picasso/src/Notification/Notification.tsx#L34. But `elevated` was not added by some reason. 

I don't see how `Notification` component can be used on isolation without the `variant` prop 🤷‍♂️ If `variant` prop is private we probably should make `Notification` component as internal in Picasso only. So I suggest to open `variant` prop (and it is needed for my example 😃)

packages/picasso/src/Notification/Notification.tsx
```diff
@@ -118,9 +118,8 @@ const useStyles = makeStyles<Theme>(styles, {
   name: 'Notification'
 })
 
-// eslint-disable-next-line react/display-name
 export const Notification = forwardRef<HTMLElement, PrivateProps>(
-  function Notification(props, ref) {
+  function Notification (props, ref) {
```
denieler, line 122: autofix by prettier probably

packages/picasso/src/utils/Notifications/story/index.jsx
```diff
@@ -100,6 +100,16 @@ Each of them has this list of props:
         })
       }
     })
+    .addExample('utils/Notifications/story/Custom.example.tsx', {
+      title: 'Custom',
+      effect: async (testPage, makeScreenshot) => {
+        testPage.click('[data-testid="trigger"]')
+        await testPage.waitFor(100)
```
denieler, line 107: just took the same approach we had before... later probably we will replace them with Cypress + Happo

denieler:
Not sure why `__image_snapshots__/notifications-stream-options-effect-2-snap.png` file got regenerated. Visually there are no changes, looks like

packages/picasso/src/utils/Notifications/use-notifications.tsx
```diff
@@ -94,10 +88,32 @@ export const useNotifications = () => {
     [closeSnackbar, enqueueSnackbar]
   )
 
+  const getPicassoNotification = useCallback(
```
alexanderpolyankin, line 91: nit: `getStyledNotifcation` / `getDefaultNotification`?

teimurjan (approved):

packages/picasso/src/utils/Notifications/use-notifications.tsx
```diff
@@ -76,16 +75,11 @@ export const useNotifications = () => {
...
-          <StyledNotification
-            content={content}
-            icon={icon}
-            key={key}
-            variant={variant}
-            onClose={closeNotification}
-          />
-        ),
+        content: (key: string) =>
+          React.cloneElement(notificationComponent, {
```
havenchyk, line 79: Isn't it a notificationElement in this case?

packages/picasso/src/Notification/Notification.tsx
havenchyk: Please leave a comment about the reason to keep one and not another type

havenchyk (approved):

packages/picasso/src/utils/Notifications/use-notifications.tsx
denieler: renamed to `notificationElement` 👍

packages/picasso/src/Notification/Notification.tsx
denieler: I moved and changed a bit this comment:

```
/** `elevated` and `icon` props are omitted from the public declaration, since they're only for internal use */
```

to the declaration of `PublicProps`. It should be enough I think 👍

PR #2070, opened 2021-05-13
chore: add color utils docs section

denieler:
[FX-1906](https://toptal-core.atlassian.net/browse/FX-1906)

### Description

Add a section for Color utils.

Screenshot 2021-05-13 at 11 48 44

<details>
<br />

List of available commands:

</details>

### DIFF
--- a/packages/picasso/src/utils/Utils/story/Colors.example.tsx
+++ b/packages/picasso/src/utils/Utils/story/Colors.example.tsx
@@ -1,0 +1,31 @@
+import { Table } from '@toptal/picasso'
+import React from 'react'
+
+const Example = () => {
+  const data = {
+    'colorUtils.alpha': `Sets the alpha channel for the color and returns the hex color format`,
+    'colorUtils.lighten': `Sets the lightness of the color and returns the hex color format`,
+    'colorUtils.darken': `Sets the darkness of the color and returns the hex color format`
+  }
+
+  return (
+    <Table>
+      <Table.Head>
+        <Table.Row>
+          <Table.Cell>Name</Table.Cell>
+          <Table.Cell>Description</Table.Cell>
+        </Table.Row>
+      </Table.Head>
+      <Table.Body>
+        {Object.entries(data).map(([name, description], index) => (
+          <Table.Row key={index}>
+            <Table.Cell>{name}</Table.Cell>
+            <Table.Cell>{description}</Table.Cell>
+          </Table.Row>
+        ))}
+      </Table.Body>
+    </Table>
+  )
+}
+
+export default Example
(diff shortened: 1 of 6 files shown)


packages/picasso/src/utils/Utils/story/Colors.example.tsx
```diff
@@ -0,0 +1,31 @@
+import { Table } from '@toptal/picasso'
+import React from 'react'
+
+const Example = () => {
+  const data = {
```
teimurjan, line 5: What do you think about extracting it out of the component's scope?

OleksandrNechai, line 5: I would inline this variable

OleksandrNechai (approved):

packages/picasso/src/utils/Utils/story/Colors.example.tsx
denieler: I did it by analogy with https://github.com/toptal/picasso/blob/9c31dabb9bf0b9e34e8ac36b6add29237d43fb23/packages/picasso/src/utils/Utils/story/Generic.example.tsx#L5 and other utils. If we want to change - we would need to change everywhere, but I don't see huge reason to do that, so I would probably leave it as it is, ok?

packages/picasso/src/utils/Utils/story/Colors.example.tsx
```diff
@@ -0,0 +1,31 @@
...
+    <Table>
+      <Table.Head>
+        <Table.Row>
+          <Table.Cell>Name</Table.Cell>
+          <Table.Cell>Description</Table.Cell>
+        </Table.Row>
+      </Table.Head>
+      <Table.Body>
+        {Object.entries(data).map(([name, description], index) => (
+          <Table.Row key={index}>
```
teimurjan, line 21: We have a rule not to use indices as keys. Let's use `name` as a key.

teimurjan (approved):

packages/picasso/src/utils/Utils/story/Colors.example.tsx
OleksandrNechai: Yes, its rather nit, just this `data` and maps over it makes code a bit more complicated for not reason :-)

packages/picasso/src/utils/Utils/story/Colors.example.tsx
denieler: yep, thank you 🙇‍♂️ the change is done 👍

packages/picasso/src/utils/Utils/story/Colors.example.tsx
denieler: done 👍

PR #2067, opened 2021-05-12
feat(FileList): [TEA-2545] Implement internal file list component

besirovic:
[TEA-2545]

[DESIGN](https://share.goabstract.com/2556a122-6443-4546-96db-3d40c37b2a0d?mode=build&sha=0aeb05e01da837c228249c74cc79ec21280f8b6d)

### Description

Implement internal `FileList` component needed for `FileInput` component. Component will be added to `FileInput` as part of  [TEA-2546](https://toptal-core.atlassian.net/browse/TEA-2546).

> NOTE
>
> `FileList` example inside `FileInput` section is added just for testing purpose and will be removed as part of  [TEA-2546](https://toptal-core.atlassian.net/browse/TEA-2546)

### How to test

- In storybook, open `FileInput` component and find `FileList` section on the bottom
- Check if list of files is implemented according design

### Screenshots

Screenshot 2021-05-12 at 02 14 06

### Review

- [x] Read [CONTRIBUTING.md](https://github.com/toptal/picasso/blob/master/CONTRIBUTING.md) and [Component API principles](https://github.com/toptal/picasso/blob/master/docs/api-principles.md)
- [x] Annotate all `props` in component with documentation
- [x] Create `examples` for component
- [ ] Ensure that deployed demo has expected results and good examples
- [x] Ensure that tests pass by running `yarn test`
- [ ] Ensure that visuals tests pass by running `yarn test:visual`. If not - check the documentation [how to fix visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md#fixing-broken-visual-tests-inside-a-pr)
- [ ] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).

<details>
<summary>PR commands</summary>
<br />

List of available commands:

- `@toptal-bot run all` - Run whole pipeline
- `@toptal-bot run build` - Check build
- `@toptal-bot run visual` - Run visual tests
- `@toptal-bot run deploy:documentation` - Deploy documentation
- `@toptal-bot run package:alpha-release` - Release alpha version

</details>

[TEA-2545]: https://toptal-core.atlassian.net/browse/TEA-2545

PR #2082, opened 2021-05-21
feat(FileList): Revert [TEA-2545] Implement internal file list component

havenchyk:
Reverts toptal/picasso#2067


havenchyk:
This is needed because this change introduced a breaking change in picasso, we decided to follow the way of fixing it:

- revert current PR
- include changes into the next major version

So backward compatibility is not broken, users who don't want to upgrade to v9 are in a safe position

PR #2088, opened 2021-05-26
feat(ci): create workflow for gha stats

IvanShamatov:
[CI-490]

### Description

This is a workflow which is triggered by other workflow's completion event. It sends request to Anvil to fetch stats.

[CI-490]: https://toptal-core.atlassian.net/browse/CI-490

### DIFF
--- a/.github/workflows/report_gha_stats_to_anvil.yml
+++ b/.github/workflows/report_gha_stats_to_anvil.yml
@@ -1,0 +1,20 @@
+name: 'Report GHA stats to Anvil'
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: ["Visual Testing", "Request review on PRs", "Consulting the Oracle", "Automatic Rebase"]
+    types: 
+      - completed
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


conf (approved):
LGTM, please, fix the failing checks and this is good to go.

.github/workflows/report_gha_stats_to_anvil.yml
```diff
@@ -0,0 +1,20 @@
+name: 'Report GHA stats to Anvil'
+on:
+  workflow_dispatch:
+  workflow_run:
+    workflows: ["Visual Testing", "Request review on PRs", "Consulting the Oracle", "Automatic Rebase"]
```
OleksandrNechai, line 5: What if tomorrow we add a new action and forget to add it to this list. What happens?

OleksandrNechai (approved):
LGTM, did not test.

.github/workflows/report_gha_stats_to_anvil.yml
IvanShamatov: Not new action, but workflow, and yes, there is no solution for that right now unfortunately

PR #2092, opened 2021-06-02
fix(Slider): overlapping labels on range slider

krasai:
[SPC-1098]

### Description

The labels for range sliders are overlapping themself when they are close to each other. We need to change tooltip placement when labels intersect. 

### How to test

- Go to the `Slider` story: `/?path=/story/components-slider--slider`
- Test Range Slider. 

### Screenshots

|Kapture 2021-06-02 at 15 05 09|Kapture 2021-06-02 at 14 59 41|

### Review

<details>
<br />

List of available commands:

</details>

[SPC-1098]: https://toptal-core.atlassian.net/browse/SPC-1098

### DIFF
--- a/packages/picasso/src/Slider/Slider.tsx
+++ b/packages/picasso/src/Slider/Slider.tsx
@@ -3,7 +3,8 @@ import React, {
   ChangeEvent,
   ComponentProps,
   useRef,
-  useState
+  useState,
+  useMemo
 } from 'react'
 import { makeStyles, Theme } from '@material-ui/core/styles'
 import MUISlider, { ValueLabelProps } from '@material-ui/core/Slider'
@@ -62,13 +63,19 @@ type ValueLabelComponentProps = ValueLabelProps & {
   index: number
 }
 
-const DefaultTooltip = (
-  isTooltipAlwaysVisible: boolean,
-  disablePortal?: boolean,
-  compact?: boolean,
+type DefaultTooltipProps = {
+  isTooltipAlwaysVisible: boolean
+  disablePortal?: boolean
+  compact?: boolean
   isTooltipReversed?: boolean
-  // eslint-disable-next-line max-params
-): React.FunctionComponent<ValueLabelComponentProps> => ({
+}
+
+const DefaultTooltip = ({
+  isTooltipAlwaysVisible,
+  disablePortal,
+  compact,
+  isTooltipReversed
+}: DefaultTooltipProps): React.FunctionComponent<ValueLabelComponentProps> => ({
   children,
   open,
   value,
@@ -141,13 +148,23 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider (
   const isTooltipAlwaysVisible = tooltip === 'on'
   const isThumbHidden =
     hideThumbOnEmpty && (typeof value === 'undefined' || value === null)
-  const ValueLabelComponent = (UserDefinedTooltip ||
-    DefaultTooltip(
-      isTooltipAlwaysVisible,
-      disablePortal,
+  const ValueLabelComponent = useMemo(
+    () =>
+      (UserDefinedTooltip ||
+        DefaultTooltip({
+          isTooltipAlwaysVisible,
+          disablePortal,
+          compact,
+          isTooltipReversed
+        })) as typeof UserDefinedTooltip,
+    [
+      UserDefinedTooltip,
       compact,
+      disablePortal,
+      isTooltipAlwaysVisible,
       isTooltipReversed
-    )) as typeof UserDefinedTooltip
+    ]
+  )
 
   const watchTooltipsPlacement = () => {
     if (sliderRef.current) {
@@ -209,7 +226,7 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider (
         }}
         ValueLabelComponent={
           // From Workaround for https://github.com/mui-org/material-ui/issues/21889
-          ValueLabelComponent as React.ElementType<ValueLabelProps>
+          (ValueLabelComponent as unknown) as React.ElementType<ValueLabelProps>
         }
         valueLabelFormat={tooltipFormat}
         valueLabelDisplay={tooltip}
--- a/packages/picasso/src/Slider/Slider.tsx
+++ b/packages/picasso/src/Slider/Slider.tsx
@@ -68,13 +68,15 @@ type DefaultTooltipProps = {
   disablePortal?: boolean
   compact?: boolean
   isTooltipReversed?: boolean
+  isRange?: boolean
 }
 
 const DefaultTooltip = ({
   isTooltipAlwaysVisible,
   disablePortal,
   compact,
-  isTooltipReversed
+  isTooltipReversed,
+  isRange
 }: DefaultTooltipProps): React.FunctionComponent<ValueLabelComponentProps> => ({
   children,
   open,
@@ -87,15 +89,15 @@ const DefaultTooltip = ({
   }
 
   const placement = () => {
-    if (!isTooltipAlwaysVisible) {
-      return 'top'
+    if (!isRange) {
+      return 'right'
     }
 
     if (isTooltipReversed) {
-      return index === 0 ? 'left' : 'right'
+      return index === 0 ? 'top-end' : 'top-start'
     }
 
-    return index === 0 ? 'right' : 'left'
+    return 'top'
   }
 
   return (
@@ -112,7 +114,7 @@ const DefaultTooltip = ({
   )
 }
 
-export const Slider = forwardRef<HTMLElement, Props>(function Slider (
+export const Slider = forwardRef<HTMLElement, Props>(function Slider(
   props,
   ref
 ) {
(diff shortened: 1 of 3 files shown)


packages/picasso/src/Slider/Slider.tsx
```diff
@@ -52,27 +59,43 @@ export interface Props extends ComponentProps<typeof MUISlider> {
...
+  index: number
 }
 
 const DefaultTooltip = (
   isTooltipAlwaysVisible: boolean,
   disablePortal?: boolean,
-  compact?: boolean
+  compact?: boolean,
+  isTooltipReversed?: boolean
+  // eslint-disable-next-line max-params
```
hmschreiner, line 70: NIT: change `DefaultTooltip` to receive an object as a param so you can get rid of the `eslint-disable-next-line max-params` comment:

```javascript
  const DefaultTooltip = ({
    isTooltipAlwaysVisible: boolean,
    disablePortal?: boolean,
    compact?: boolean,
    isTooltipReversed?: boolean
  }): FC<ValueLabelComponentProps> => ({
  ...
```

elviocb: if the interface changes this will become a breaking change, not sure if it's worth only to get rid of eslint comment

s0ber:
The placement works as expected, but the tooltips are blinking when being dragged. It looks like a glitch.

krasai:
@s0ber this problem existed before. We pass the props to the tooltip via callback, and the tooltip component is re-rendered on every change. Anyway, I wrapped it inside useMemo, and the issue is less visible now.

OleksandrNechai (changes requested):
@krasai I see the following problems with this solution:
* the problem is not solved, we still have overlaps but on the boundaries: 
Peek 2021-06-07 12-56
* this switching of both tooltips doesn't look nice to me
Peek 2021-06-07 12-58
WDYT about the following next steps?
* Go consult with designers on how tooltips should behave. With the current design, overlapping seems unavoidable. Hence I don't see why we should complicate code to avoid it.
* Discuss what to do next having the designer's input. If they are fine with the current implementation, I would always render tooltips by default like this: 
image
and only switch when tooltip begins to overlap with component boundary on the left or the right. This will probably generate less switching and simpler implementation.

krasai:
@OleksandrNechai 
After the discussion on the # -base-core channel, I implemented it according to the provided projects: https://toptal-core.atlassian.net/browse/CSPR-293
I think sometimes overlapping is unavoidable. If there is no space on the right side we should display it a bit closer to the left side.

OleksandrNechai:
@krasai hey! I have some feedback from Milos Milikic

> I found a couple of things that are not the same as in the designs:
> * know size should be 11px (currently it’s 9px)
> * Filled line should be blue (currently it’s gray on some examples)

Do you think it is related to the current PR?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
-    )) as typeof UserDefinedTooltip
+      isTooltipAlwaysVisible,
+      isTooltipReversed,
+      isRange
+    ]
+  )
+
+  const watchTooltipsPlacement = () => {
+    if (sliderRef.current) {
+      const tooltips: NodeListOf<HTMLElement> = sliderRef.current.querySelectorAll(
```
OleksandrNechai, line 176: I am concerned with this approach of getting references for tooltips. What happens if MUI changes roles?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
       disablePortal,
-      compact
-    )) as typeof UserDefinedTooltip
+      isTooltipAlwaysVisible,
+      isTooltipReversed,
+      isRange
+    ]
+  )
+
+  const watchTooltipsPlacement = () => {
```
OleksandrNechai, line 174: From the code of the function, I find its name confusing. It just does a bunch of calculations, what does it watch? BTW can we use Intersection Observer API instead? Like switch placement, if tooltips intersect?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -52,27 +60,51 @@ export interface Props extends ComponentProps<typeof MUISlider> {
...
   value,
-  valueLabelDisplay
+  valueLabelDisplay,
+  index
 }) => {
   if (valueLabelDisplay === 'off') {
     return children
   }
 
+  const placement = () => {
```
OleksandrNechai, line 91: Looks like it could be just a variable not a function. Or maybe this function can be taken out of the component so we avoid creating it each render?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
     unmarkTrack,
     ...classes
   } = useStyles()
+  const sliderRef = useCombinedRefs<HTMLElement>(ref, useRef<HTMLElement>(null))
+  const [isTooltipReversed, setIsTooltipReversed] = useState(false)
```
OleksandrNechai, line 148: Maybe `flipped`? :)

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
+  const ValueLabelComponent = useMemo(
+    () =>
+      (UserDefinedTooltip ||
+        DefaultTooltip({
+          isTooltipAlwaysVisible,
+          disablePortal,
+          compact,
+          isTooltipReversed,
+          isRange
+        })) as typeof UserDefinedTooltip,
```
OleksandrNechai, line 163: Can we avoid `as typeof UserDefinedTooltip` and instead tune `DefaultTooltip` typing?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
+  const watchTooltipsPlacement = () => {
+    if (sliderRef.current) {
+      const tooltips: NodeListOf<HTMLElement> = sliderRef.current.querySelectorAll(
+        '[role="tooltip"]'
+      )
+      const sliders: NodeListOf<HTMLElement> = sliderRef.current.querySelectorAll(
+        '[role="slider"]'
+      )
+
+      if (sliders.length === 2 && tooltips.length === 2) {
```
OleksandrNechai, line 183: How about reusing `isRange`?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
+        const minDistance =
+          Array.from(tooltips).reduce((acc, cur) => {
```
OleksandrNechai, line 185: Why reduce? :) Maybe we could do just map(...).sum()?
Where this min distance algo comes from?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,21 +144,72 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
+          setIsTooltipReversed(true)
+        }
+      }
+    }
+  }
+
+  const handleChange = (event: React.ChangeEvent<{}>, newValue: Value) => {
+    onChange?.(event, newValue)
+
+    if (isRange && disablePortal) {
```
OleksandrNechai, line 203: why is this only working with disabled portals? I find its weird to have people remember to pass this prop so their tooltips do not overlap

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -149,11 +232,11 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
         }}
         ValueLabelComponent={
           // From Workaround for https://github.com/mui-org/material-ui/issues/21889
-          ValueLabelComponent as React.ElementType<ValueLabelProps>
+          (ValueLabelComponent as unknown) as React.ElementType<ValueLabelProps>
```
OleksandrNechai, line 235: Eh? :)

OleksandrNechai (commented):
First of all, thank you @krasai for the contribution to Picasso! We really appreciate it! :bow: 

And sorry about a very late review, we are a bit understaffed now and vacation season so :man_shrugging: 

Now, I will be clear, I find the solution hacky. I do realize that there might be no elegant way to do things. Maybe we will need to negotiate with designers again if there is no way we can improve it. So can we somehow not rely on selecting DOM elements using `mui` markup. If they change it, our component breaks. Can we maybe use an intersection observer?  See other questions above.

packages/picasso/src/Slider/Slider.tsx
OleksandrNechai: This could be a performance hit to run dom select on each change? :thinking:

packages/picasso/src/Slider/Slider.tsx
OleksandrNechai: or tooltips[0].offsetWidth + tooltips[1].offsetWidth :-D

teimurjan:
I created an alternative solution by using refs instead of `document.querySelector`. It's not clean because I wanted to show the approach: https://github.com/toptal/picasso/compare/SPC-1098-overlapping-labels-on-range-filters-2

krasai:
Closing it, the work is continued on https://github.com/toptal/picasso/pull/2108

PR #2093, opened 2021-06-03
fix(Drawer): fix incorrect z-index of the Drawer

dmaklygin1:
### Description

The Drawer's appearance covers the appearance of the `Modal` window, however `zIndex` of the Drawer and `Modal` must be 1200 and 1300 accordingly.

The problem seems to be on the UI-Material side because their Drawer's zIndex also equals 1300 (https://material-ui.com/components/drawers/#api)
I'm going to create a PR to UI Material repo as well.

### Screenshots

before:

https://user-images.githubusercontent.com/76203853/120647367-33938d80-c483-11eb-9999-77da0c298aa7.mp4

after:

https://user-images.githubusercontent.com/76203853/120647368-33938d80-c483-11eb-815e-bc803e8470b0.mp4

### Review

<details>
<br />

List of available commands:

</details>

### DIFF
--- a/packages/picasso-lab/src/Drawer/Drawer.tsx
+++ b/packages/picasso-lab/src/Drawer/Drawer.tsx
@@ -1,5 +1,5 @@
 import MUIDrawer from '@material-ui/core/Drawer'
-import { makeStyles, Theme } from '@material-ui/core/styles'
+import { makeStyles, Theme, useTheme } from '@material-ui/core/styles'
 import cx from 'classnames'
 import { Button, Container } from '@toptal/picasso'
 import { BaseProps, useDrawer, usePicassoRoot } from '@toptal/picasso-shared'
@@ -44,6 +44,7 @@ export const Drawer: FunctionComponent<Props> = props => {
   } = props
   const classes = useStyles()
   const { setHasDrawer } = useDrawer()
+  const theme = useTheme()
   const container = usePicassoRoot()
 
   useLayoutEffect(() => {
@@ -69,6 +70,7 @@ export const Drawer: FunctionComponent<Props> = props => {
       onClose={handleOnClose}
       disablePortal={disablePortal}
       container={container}
+      ModalProps={{ style: { zIndex: theme.zIndex.drawer } }}
     >
       <Container
         flex


packages/picasso-lab/src/Drawer/Drawer.tsx
```diff
@@ -69,6 +70,7 @@ export const Drawer: FunctionComponent<Props> = props => {
       onClose={handleOnClose}
       disablePortal={disablePortal}
       container={container}
+      ModalProps={{ style: { zIndex: theme.zIndex.drawer } }}
```
piotrd, line 73: We need to ask @toptal/frontend-experience-eng guys, but I think there should be some test for this.

piotrd (approved):

packages/picasso-lab/src/Drawer/Drawer.tsx
dmaklygin1: hm, what kind of test is going to be?
I can see only the integration test here, to test how the Drawer is shown along with the Modal window.

denieler: please don't use inline styles 🙇‍♂️ it should be possible to pass a className at least

OleksandrNechai: yes, integration one would be awesome, done with cypress+happo. For example like here https://github.com/toptal/picasso/blob/master/cypress/component/Accordion.spec.tsx

dmaklygin1: I'll check whether it's possible to pass a className.

ertrzyiks: It should be possible to set z-index using overrides like: 
```
PicassoProvider.override(() => ({
  MuiDrawer: {
    paper: {
      maxWidth: '100%'
    },
    modal: {
      zIndex: 1200
    }
  }
}))
```

but MUIModal applies zIndex using inline styles, so such override is overwritten anyway.

denieler: we need to understand why it's overridden from the default zIndex, which specified by MUI

dmaklygin1: in MUI, the Drawer also has a 1300 zIndex.
Why it works in MUI and does not work in Picasso? Because both Drawer and Modal are rendered in the same node and Modal is rendered after the Drawer.
in Picasso, we may render Drawer and Modal in separate nodes.

dmaklygin1: The issue is definitely on the UI side. And seems, it will be fixed, but it will be fixed on version 5 (https://github.com/mui-org/material-ui/issues/26580), however, the Picasso is still on 4.

So, either we need to wait for Picasso upgrade or push the fix to the current version.

ertrzyiks: If this issue is blocking for top team we could apply this fix with a comment to remove it after update to MUI 5. As the original incorrect zindex is applied using inline styles, we are forced to use inline styles or !important anyway.

ertrzyiks (approved):
It's a good workaround until the issue will be fixed in MUI. Let's add missing visual tests to cover this case with a screenshot. I suggest to comment the approach as a temporary while the issue is being addressed in 3rd party.

packages/picasso-lab/src/Drawer/Drawer.tsx
denieler: if I'm not mistaken they said in https://github.com/mui-org/material-ui/issues/26580 that the issue can be still reproduced in v5. So upgrading MUI to v5 will not fix it.

ertrzyiks: Afaik the policy of MUI library is that they now fix bugs only in v5. Some crucial are probably backported to v4, but I would expect the fix for z-index to land in v5 only.

denieler: yeah, it's definitely a way to go later - to upgrade MUI in Picasso 👍 when MUI v5 will be stable

ertrzyiks: I found the note confirming my blurry memory about material ui update policy in their readme file.
```
All ongoing work has moved to v5. The development of v4 is limited to important bug fixes, security patches and easing the upgrade path to v5.
```

OleksandrNechai (approved):

PR #2102, opened 2021-06-16
feat(tooltip): show tooltip on both label and control

MrBra1nwash:
[FX-1874](https://toptal-core.atlassian.net/browse/FX-1874)

### Description

The tooltip for the radio and checkbox with the label was positioned incorrectly and was not visible if we hover over the label. It was always centered relatively on radio or checkbox. Now it is either in the center of the button or in the center of the button and label. 
The problems that I solved here:
1. Pass only one `ref` to the child component. In case we have `label` pass `ref` to the `label`, otherwise, pass it to the `button`(`checkbox` or `radio`)
2. Pass listeners (`onMouseLeave`, `onMouseOver`) from `Tooltip` to the `label`
3. Created two examples in `Radio` and `Checkbox` components instead of `Tooltip`, because changes were done in these components 

### How to test

- Go to Radio and Checkbox story
- Scroll down to the `With tooltip` example
- Tooltip should be centered

### Screenshots
Screenshot 2021-06-16 at 17 25 23
Screenshot 2021-06-16 at 17 25 16

### Review

<details>
<br />

List of available commands:

</details>

### DIFF
--- a/cypress/component/Tooltip.spec.tsx
+++ b/cypress/component/Tooltip.spec.tsx
@@ -3,12 +3,14 @@ import { mount } from '@cypress/react'
 import {
   Autocomplete,
   Button,
+  Checkbox,
   Container,
   Dropdown,
   Grid,
   Link,
   Menu,
   Modal,
+  Radio,
   Tooltip,
   TooltipProps,
   Typography
@@ -165,6 +167,30 @@ const LinkTooltipExample = () => {
   )
 }
 
+const CheckboxTooltipExample = () => {
+  const tooltipContent = <span data-testid='tooltip-content'>Content</span>
+
+  return (
+    <TestingPicasso>
+      <Tooltip content={tooltipContent} interactive>
+        <Checkbox label='Checkbox' data-testid='tooltip-trigger' />
+      </Tooltip>
+    </TestingPicasso>
+  )
+}
+
+const RadioTooltipExample = () => {
+  const tooltipContent = <span data-testid='tooltip-content'>Content</span>
+
+  return (
+    <TestingPicasso>
+      <Tooltip content={tooltipContent} interactive>
+        <Radio label='Radio' data-testid='trigger' />
+      </Tooltip>
+    </TestingPicasso>
+  )
+}
+
 const AutocompleteTooltipExample = () => {
   const [value, setValue] = useState('')
 
@@ -284,6 +310,30 @@ describe('Tooltip', () => {
     cy.get('[data-testid="tooltip-content"').should('be.visible')
 
     cy.get('[data-testid="tooltip-trigger"').click()
+    cy.get('[data-testid="tooltip-content"').should('not.be.visible')
+  })
+
+  it('renders on hover, and hides on click for Checkbox', () => {
+    mount(<CheckboxTooltipExample />)
+
+    cy.get('[data-testid="tooltip-content"').should('not.exist')
+    cy.get('[data-testid="tooltip-trigger"').realHover()
+
+    cy.get('[data-testid="tooltip-content"').should('be.visible')
+    cy.get('body').happoScreenshot()
+    cy.get('[data-testid="tooltip-trigger"').click()
+    cy.get('[data-testid="tooltip-content"').should('not.be.visible')
+  })
+
+  it('renders on hover, and hides on click for Radio', () => {
+    mount(<RadioTooltipExample />)
+
+    cy.get('[data-testid="tooltip-content"').should('not.exist')
+    cy.get('[data-testid="trigger"').realHover()
+
+    cy.get('[data-testid="tooltip-content"').should('be.visible')
+    cy.get('body').happoScreenshot()
+    cy.get('[data-testid="trigger"').click()
     cy.get('[data-testid="tooltip-content"').should('not.be.visible')
   })
 
--- a/packages/picasso/src/Checkbox/Checkbox.tsx
+++ b/packages/picasso/src/Checkbox/Checkbox.tsx
@@ -7,7 +7,7 @@ import {
   TextLabelProps
 } from '@toptal/picasso-shared'
 import cx from 'classnames'
-import React, { forwardRef, ReactNode } from 'react'
+import React, { ComponentProps, forwardRef, ReactNode } from 'react'
 
 import { RequiredDecoration } from '../FormLabel'
 import CheckboxGroup from '../CheckboxGroup'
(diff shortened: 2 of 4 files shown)


dmaklygin1:
I've noticed, that in the disabled state a tooltip still appears.
Picasso | Checkbox 2021-06-17 15-42-45

packages/picasso/src/Checkbox/Checkbox.tsx
```diff
@@ -46,76 +46,98 @@ export interface Props
...
+
+    const classes = useStyles()
+    const rootClasses = {
+      root: classes.root,
+      disabled: classes.disabled
+    }
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { color, ...checkboxAttributes } = rest
+
+    console.log('checkboxAttributes', checkboxAttributes)
```
dmaklygin1, line 83: console

MrBra1nwash: 🙈 Thanks, fixed.

MrBra1nwash:
> I've noticed, that in the disabled state a tooltip still appears.
> Picasso | Checkbox 2021-06-17 15-42-45

I think it is expected behavior because we can use it when we want to tell the user why it is disabled. Whether the tooltip should be displayed is up to the consumer of the component

packages/picasso/src/Checkbox/Checkbox.tsx
```diff
@@ -46,76 +46,97 @@ export interface Props
...
-    ...rest
-  } = props
+const CheckboxWrapper: FC<{ className: string }> = ({
+  className,
+  children
+}) => (
+  <Container as='span' flex inline className={className}>
+    {children}
+  </Container>
+)
```
dmaklygin1, line 56: hm, I guess it's a minor update. `CheckboxWrapper` is used twice and called with the same set of props. I would keep `<Container as='span' flex inline className={classes. checkboxWrapper}>`

dmaklygin1: I mean, caching to variables is a good approach. But here, it's confusing a little bit.

dmaklygin1:
> > I've noticed, that in the disabled state a tooltip still appears.
> > Picasso | Checkbox 2021-06-17 15-42-45
> 
> I think it is expected behavior because we can use it when we want to tell the user why it is disabled. Whether the tooltip should be displayed is up to the consumer of the component

maybe, but it's not shown in the current production version.

packages/picasso/src/Checkbox/Checkbox.tsx
MrBra1nwash: Sorry, I didn't get your point. What exactly is confusing? If I remove `CheckboxWrapper` i have to use `<Container as='span' flex inline className={className}>` in two places. And if we want to make changes in `<Container as='span' flex inline className={className}>` we have to make them in two places instead of one.

MrBra1nwash:
> maybe, but it's not shown in the current production version.

Do you want me to add two more stories with a disabled checkbox and radio with a tooltip? `Tooltip` is just a wrapper around any component. Component wrapped in the `Tooltip` can have `50` states (e.g. disabled, active, different variants etc.). I don't think that we need to show a separate story for every state + tooltip

sashuk:
@MrBra1nwash I think @dmaklygin means that the disabled checkbox should not show a tooltip just like it does not show the tooltip in `master`, as there is a risk that it might break things if somebody relies on this (that disabled checkboxes do not show tooltips).

sashuk (approved):
Tested on temploy, works great! The disabled checkbox issue should be sorted out, I think, but besides that everything is fine

packages/picasso/src/Checkbox/story/index.jsx
```diff
@@ -55,3 +55,4 @@ page
   })
   .addExample('Checkbox/story/Required.example.tsx', 'Required')
   .addExample('Checkbox/story/CustomLabel.example.tsx', 'Custom label')
+  .addExample('Checkbox/story/WithTooltip.example.tsx', 'With tooltip')
```
denieler, line 58: we probably need some effect to be added to make a screenshot with the tooltip

packages/picasso/src/Checkbox/Checkbox.tsx
```diff
@@ -46,76 +46,97 @@ export interface Props
...
+      id,
+      indeterminate,
+      onChange,
+      value,
+      focusVisibleClassName: classes.focused
+    }
+
+    if (!label) {
+      return (
+        <CheckboxWrapper className={classes.checkboxWrapper}>
```
denieler, line 104: why we couldn't save the form of code?
```
const muiCheckbox = (
    <Container as='span' flex inline className={classes.checkboxWrapper}>
      <MUICheckbox
      ...
```
I don't see the difference why we've split into 2 places with:
```
<CheckboxWrapper className={classes.checkboxWrapper}>
            <MUICheckbox
```

packages/picasso/src/Checkbox/Checkbox.tsx
dmaklygin1: We create an additional component just for caching 3 props: as, flex and inline.
I would use:
```
<span className={classes.checkboxWrapper}>
  <MUICheckbox ... />
</span>
```

and `flex` and `inline` props we should put to the definition of the `classes.checkboxWrapper`

dmaklygin1: it looks lighter

dmaklygin1 (commented):
only two things:
1. some optimisation in `CheckboxWrapper`
2. doubts concerning a `disabled` state

packages/picasso/src/Checkbox/Checkbox.tsx
MrBra1nwash: Because if there is no label we need to pass `ref` to a `MUICheckbox`. If we have a `label` we need to pass `ref` to `FormControlLabel`

MrBra1nwash:
> @MrBra1nwash I think @dmaklygin means that the disabled checkbox should not show a tooltip just like it does not show the tooltip in `master`, as there is a risk that it might break things if somebody relies on this (that disabled checkboxes do not show tooltips).
Thank you for the clarification :) But it is actually a bug in Picasso. Tooltip doesn't work with Checkbox or Radio. If someone relies on the bug we don't need to adjust to it. If Picasso owners consider it as a breaking change I can make this update major

https://user-images.githubusercontent.com/16545305/122569344-52814a80-d053-11eb-8e21-caeb031a5cc4.mov

packages/picasso/src/Checkbox/story/index.jsx
MrBra1nwash: I could add an `open` prop to the `Tooltip` to the story to not complicate things too much. WDYT?

packages/picasso/src/Checkbox/Checkbox.tsx
MrBra1nwash: With this approach, we will rely on the implementation details of the `Container` component. If any of the `as='span' flex inline` props would change their behavior we must replicate it in `classes.checkboxWrapper`.

packages/picasso/src/Checkbox/Checkbox.tsx
denieler: why then not make it explicit?
```
<MUICheckbox
  ref={label ?  undefined : ref}
  ...
<FormControlLabel
  ref={label ? ref : undefined}
```
?

packages/picasso/src/Checkbox/story/index.jsx
denieler: I would not make an additional example for this. If you want to make a screenshot - I would use the Cypress test for it.

packages/picasso/src/Checkbox/Checkbox.tsx
MrBra1nwash: Daniel's tips solved the problem :) I removed `CheckboxWrapper`, please check

packages/picasso/src/Checkbox/story/index.jsx
MrBra1nwash: Removed and added cypress tests. Thank you!

packages/picasso/src/Checkbox/Checkbox.tsx
MrBra1nwash: I have no idea why I didn't do that right away 🙈 Thank you, fixed!

cypress/component/Tooltip.spec.tsx
```diff
@@ -287,6 +313,30 @@ describe('Tooltip', () => {
     cy.get('[data-testid="tooltip-content"').should('not.be.visible')
   })
 
+  it('renders on hover, and hides on click for Checkbox', () => {
+    mount(<CheckboxTooltipExample />)
+
+    cy.get('[data-testid="tooltip-content"').should('not.exist')
+    cy.get('[data-testid="tooltip-trigger"').realHover()
+
+    cy.get('[data-testid="tooltip-content"').should('be.visible')
```
denieler, line 322: but don't you need a screenshot? I thought the real issue is that tooltip appearing not aligned
```
cy.get('body').happoScreenshot()
```

MrBra1nwash: I didn't know about, thank you, fixed. @denieler please see this discussion https://github.com/toptal/picasso/pull/2102#issuecomment-864046221

packages/picasso/src/Checkbox/Checkbox.tsx
```diff
@@ -46,76 +46,84 @@ export interface Props
...
-    disabled,
-    requiredDecoration,
-    onChange,
-    value,
-    checked,
-    indeterminate,
-    titleCase,
-    ...rest
-  } = props
+export const Checkbox = forwardRef<HTMLButtonElement | HTMLLabelElement, Props>(
```
denieler, line 49: this type looks not right `HTMLButtonElement | HTMLLabelElement`, firstly - because it can be confusing for the component user why ref can have 2 different types and there is no information from outside when it will receive one of another without looking into the component itself. And secondly, because later we convert the type to `HTMLButtonElement` ref only - https://github.com/toptal/picasso/blob/9c0701584d8dfabb36db985d30b821c60867577c/packages/picasso/src/Checkbox/Checkbox.tsx#L126

denieler (approved):

denieler:
> > @MrBra1nwash I think @dmaklygin means that the disabled checkbox should not show a tooltip just like it does not show the tooltip in `master`, as there is a risk that it might break things if somebody relies on this (that disabled checkboxes do not show tooltips).
> 
> Thank you for the clarification :) But it is actually a bug in Picasso. Tooltip doesn't work with Checkbox or Radio. In my PR I fixed this bug. If someone relies on the bug we don't need to adjust to it. If Picasso owners consider it as a breaking change I can make this update major
> 
>  Screen.Recording.2021-06-18.at.16.34.17.mov

This change would not require people to fix somehow the component, tests also most likely would not fail. So I would release this as a non-breaking change. People will just receive a new functionality, a new feature. So Minor update looks valid to me 👍

MrBra1nwash:
> This change would not require people to fix somehow the component, tests also most likely would not fail. So I would release this as a non-breaking change. People will just receive a new functionality, a new feature. So Minor update looks valid to me 👍

Thanks. Changed from `fix` to `feat`

packages/picasso/src/Checkbox/Checkbox.tsx
MrBra1nwash: The [follow up](https://toptal-core.atlassian.net/browse/SP-1518) ticket is created

PR #2104, opened 2021-06-22
feat: [FX-1956] Add picasso-provider package

denieler:
[FX-1956]

### Description

Detachment of the global Picasso provider and context to the separate `@toptal/picasso-provider` package.

Because this functionality almost never changes in Picasso lifetime it looks like it makes sense to detach this as a separate package, which would simplify the version management of Picasso in monorepos and micro-frontends architectures (each MFE can have its own version of Picasso components, while all will be using the same version of picasso-provider).

<details>
<br />

List of available commands:

</details>

[FX-1956]: https://toptal-core.atlassian.net/browse/FX-1956


denieler:

denieler:

denieler:

denieler:

.eslintrc
```diff
@@ -70,6 +70,22 @@
...
+        "no-restricted-imports": ["error", "@toptal/picasso-provider"]
+      }
+    },
+    // don't allow picasso-shared external imports inside picasso-provider
+    {
+      "files": ["packages/picasso-provider/src/**"],
+      "excludedFiles": ["*.example.jsx", "*.example.tsx"],
+      "rules": {
+        "no-restricted-imports": ["error", "@toptal/picasso-shared"]
+      }
```
denieler, line 88: there shouldn't be any references from provider to shared. Because we use provider in shared.

packages/picasso-provider/bin/build.js
```diff
@@ -6,7 +6,7 @@ const sh = require('shelljs')
 const { log, safeExec } = require('../../../bin/utils')
 
 // typescript doesn't copy binaries, so we need to do it manually
-function copyIcons() {
+function copyIcons () {
```
denieler, line 9: linter/prettier

packages/picasso/src/index.ts
```diff
@@ -137,3 +137,17 @@ export { default as Breadcrumbs } from './Breadcrumbs'
...
+Picasso root component has been moved to @toptal/picasso-provider package.
+
+Please use:
+  import Picasso from '@toptal/picasso-provider'
+
+instead.
+  `)
+}
+
+export default PicassoFallback
```
denieler, line 153: I've added this fallback component to raise an error when Picasso is used in the old way (from `@toptal/picasso` package). Because by default vscode don't raise any errors, because default imported Picasso looks like can be used in a way `Picasso.Edit24` for icons.

packages/picasso/src/index.ts
```diff
@@ -137,3 +137,17 @@ export { default as Breadcrumbs } from './Breadcrumbs'
 export type { BreadcrumbsProps } from './Breadcrumbs'
 
 export * from './Icon'
+
+// TODO: Remove in @toptal/picasso@11.x.x version
+const PicassoFallback = () => {
```
teimurjan, line 142: 👍

teimurjan (approved):

packages/picasso/package.json
```diff
@@ -27,6 +27,7 @@
   },
   "dependencies": {
     "@material-ui/core": "4.11.0",
+    "@toptal/picasso-provider": "^0.0.1",
```
vedrani, line 30: Should picasso-provider drive notistack version? 
We could get in lot of problems if those 2 diverge because we will lose notification functionality.

packages/picasso-provider/src/Picasso/NotificationsProvider/NotificationsProvider.tsx
```diff
@@ -3,7 +3,7 @@ import cx from 'classnames'
 import { SnackbarProvider } from 'notistack'
 import React, { FunctionComponent } from 'react'
 
-import { useDrawer, usePageTopBar } from '../Picasso'
+import { useDrawer, usePageTopBar } from '../RootContext'
```
vedrani, line 6: Seems we cannot move this to Page as I suggested, because it's used on higher level.

Side question, do we ensure drawer and pageTopBar flags are reseted if we route to Page that doesn't have Drawer or Topbar?

packages/picasso-provider/src/Picasso/Picasso.tsx
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable complexity */
 import {
-  createMuiTheme,
   MuiThemeProvider,
```
vedrani, line 3: I think this will create problems for us, because it's same problem but on MUI level, with MUI provider `MuiThemeProvider` and `StylesProvider`. Different versions of MUI in Picasso-provider and Picasso will create problem of MUI accessing `theme` from React context.

Basically we can have Picasso-provider but afaik it doesn't solve problem of having multiple versions of Picasso in MFE.

vedrani (changes requested):
I have doubts that this will solve problem of multiple Picasso versions due to MUI provider and context. 

I would like to see how can we ensure we at least use same version of notistack and more importantly MUI.

packages/picasso-provider/src/Picasso/NotificationsProvider/NotificationsProvider.tsx
denieler: > Side question, do we ensure drawer and pageTopBar flags are reseted if we route to Page that doesn't have Drawer or Topbar?

I haven't checked, but I'm pretty sure they are refreshed. At least we never had such complains before

packages/picasso/package.json
denieler: done 👍

packages/picasso-provider/src/Picasso/Picasso.tsx
denieler: moved dependencies from `@topta/picasso` package to `@toptal/picasso-provider`

denieler:

denieler:
**Title:**

feat: [FX-1956] Add picasso-provider package

BREAKING CHANGE: We moved the root Picasso component to the separate package - `@toptal/picasso-provider`. This should help us in managing multiple different versions of Picasso package (`@toptal/picasso`) for sub-applications, while the main host application would be able to share Picasso Context via the locked version of `@toptal/picasso-provider` package.

To migrate to the new version you need to change

```
import Picasso from '@toptal/picasso'
```

to

```
import Picasso from '@toptal/picasso-provider'
```

packages/picasso-provider/package.json
```diff
@@ -20,5 +20,9 @@
   "bugs": {
     "url": "https://github.com/toptal/picasso/issues"
   },
-  "dependencies": {}
+  "dependencies": {
```
vedrani, line 23: Sorry, should we put `^`?

vedrani (approved):

packages/picasso-provider/package.json
denieler: I would keep fixed versions to avoid possible bugs, we had pretty critical recently with `notistack` bumping patch version 😞

PR #2108, opened 2021-06-30
fix(Slider): overlapping labels on range slider

teimurjan:
[SPC-1098]

### Description

The labels for range sliders are overlapping each others when too close. We need to change tooltip placement when labels intersect. 

Designs for new Range Slider: https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/master/commits/20c8113e84edea686ef2649d328d248e963e78c5/files/564adc73-0216-46a6-98e5-70ff39d5c47d/layers/4661426D-F2D3-4E56-81C6-0BAB4792872A

### How to test

- Go to the `Slider` story: `/?path=/story/components-slider--slider`
- Test Range Slider. 

### Screenshots

### Review

<details>
<br />

List of available commands:

</details>

[SPC-1098]: https://toptal-core.atlassian.net/browse/SPC-1098


packages/picasso/src/SliderValueLabel/styles.ts
```diff
@@ -0,0 +1,13 @@
+import { createStyles } from '@material-ui/core/styles'
+
+export default () =>
+  createStyles({
+    popper: {
+      '&[x-placement*="top-start"]': {
+        left: '-10px !important'
```
OleksandrNechai, line 7: why 10px, not some other number? :)

OleksandrNechai: I think there should be ways to avoid `!important`

packages/picasso/src/Slider/SliderContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+type ValueLabel = { tooltipWidth: number; thumb: HTMLElement }
+
+const SliderContextProvider = ({ children }: ProviderProps) => {
+  const [valueLabels, setValueLabels] = useState<(ValueLabel | null)[]>([
+    null,
+    null
+  ])
+  const [hasTooltipOverlow, setHasTooltipOverflow] = useState(false)
+
+  const checkTooltipsOverflow = useCallback(() => {
```
OleksandrNechai, line 37: Maybe `checkTooltipsOverlap` would be a better name?

OleksandrNechai (commented):
This is a bit complicated solution, no doubts about it. However, since the complexity is encapsulated in the component and API does not change, I think we can go forward with it. 

Designers were ok to give up with tooltips, but they preferred to keep them, so let's make them happy! :rocket: 

I think we need more cypress tests with happo screenshots covering how tooltips look like in different positions on the slider. Also, please rename the PR, checks will fail if It is not done in conventional commits way.

packages/picasso/src/SliderValueLabel/styles.ts
krasai: Unfortunately the `left` property is added as inline style by Material UI. It's going to be hard to overwrite it.

OleksandrNechai (approved):
LGTM, the only problem is that HAPPO screenshots are no good. Maybe there is a way to do some kind of delay before taking a snapshot?

packages/picasso/src/SliderValueLabel/styles.ts
```diff
@@ -0,0 +1,13 @@
...
+      },
+      '&[x-placement*="top-end"]': {
+        left: '10px !important'
```
denieler, line 10: as we use everywhere 8x metrics (8px, 16px, etc.), `10px` looks like a value, which might not be aligned with the design. Where can we check it?

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -49,40 +62,27 @@ export interface Props extends ComponentProps<typeof MUISlider> {
...
+    () => (props: ValueLabelProps) => (
+      <SliderValueLabel
+        {...props}
+        isTooltipAlwaysVisible={isTooltipAlwaysVisible}
+        disablePortal={disablePortal}
+        compact={compact}
+        isRange={isRange}
+      />
+    ),
+    [isTooltipAlwaysVisible, disablePortal, compact, isRange]
```
denieler, line 81: why do we need a separate hook for this component?

and there is something that looks not correct. Let me try to explain what I see:
1. we use `{...props}` on `SliderValueLabel` component
2. we specify all props `isTooltipAlwaysVisible, disablePortal, compact, isRange` in `useMemo`
3. we pass all props `isTooltipAlwaysVisible, disablePortal, compact, isRange` to the `SliderValueLabel` component

this means some other props can be also passed to this component, but they would not take any effect, because `useMemo` would not let it to be changed, because of `2.`

packages/picasso/src/Slider/Slider.tsx
```diff
@@ -112,50 +112,59 @@ export const Slider = forwardRef<HTMLElement, Props>(function Slider(
...
+              [hideThumb]: isThumbHidden
+            }),
+            markActive: cx(classes.markActive, {
+              [markInactive]: isThumbHidden || disableTrackHighlight
+            })
+          }}
+          ValueLabelComponent={ValueLabelComponent}
+          valueLabelFormat={tooltipFormat}
+          valueLabelDisplay={tooltip}
+          onChange={handleChange}
```
denieler, line 164: why `onChange={onChange}` was not good enough?

packages/picasso/src/Tooltip/Tooltip.tsx
```diff
@@ -167,16 +170,19 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
...
 const useStyles = makeStyles<Theme>(styles, { name: 'PicassoTooltip' })
 
-export const Tooltip: FunctionComponent<Props> = props => {
+export const Tooltip = forwardRef<unknown, Props>((props, ref) => {
   const {
     content,
     children: originalChildren,
     placement,
     interactive,
+    classes: externalClasses,
```
denieler, line 185: we used to have only `className` prop to be used to extend the styling of the component. Mostly because having `classes` is not a common practice for the React components. So it would be great to keep this practice here and try to reuse `className`

packages/picasso/src/Slider/styles.ts
```diff
@@ -30,7 +30,7 @@ export default ({ palette }: Theme) =>
       opacity: 0.24
     },
     track: {
-      backgroundColor: palette.grey.main,
+      backgroundColor: palette.primary.main,
```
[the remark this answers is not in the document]
krasai: Yes, 
The designs can be found here: https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/master/commits/20c8113e84edea686ef2649d328d248e963e78c5/files/564adc73-0216-46a6-98e5-70ff39d5c47d/layers/4661426D-F2D3-4E56-81C6-0BAB4792872A
And, also I asked on slack if the track should be blue for all variants:
https://toptal-core.slack.com/archives/CCC3GP6CC/p1625838054009700

packages/picasso/src/SliderValueLabel/styles.ts
krasai: Changed to 8px.

packages/picasso/src/Slider/Slider.tsx
krasai: I forgot to revert this change back. Fixed.

packages/picasso/src/Tooltip/Tooltip.tsx
krasai: I tried to make it work with `className`, but I couldn't figure it out. I think the problem is that the className that we add to the tooltip is added to the tooltip trigger not to the tooltip content. 
https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tooltip/Tooltip.js#L611
And it would be hard to select the content since its rendered in completely different place, inside the portal.

packages/picasso/src/Slider/Slider.tsx
krasai: I'm not sure if I understand your question correctly. But we need to wrap the SliderValueLabel in useMemo to prevent rerendering of whole ValueLabel component on every `value` change (thumb shift). 

The Slider component accept only the definition of the component that needs to be rendered as a Value Label, (the rendering is done on material ui side) and we need to pass some additional props to it (like disablePortal or isRange). 
The additional props are passed by Material-UI during the rendering phrase. From what I see eg. the `value` are updating correctly.

krasai:
> LGTM, the only problem is that HAPPO screenshots are no good. Maybe there is a way to do some kind of delay before taking a snapshot?

I tried with different time (wait()) but with no luck :(

packages/picasso/src/Tooltip/Tooltip.tsx
krasai: I removed this part. There is no need to pass the CSS anymore.

packages/picasso/src/SliderValueLabel/SliderValueLabel.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    return 'top'
+  }
+
+  const handleTooltipRef = (tooltipElement: HTMLDivElement) => {
+    if (tooltipElement && thumbRef.current) {
+      registerValueLabel(index, tooltipElement, thumbRef.current)
+    }
+  }
```
denieler, line 53: what is the purpose of having 2 refs: `ref`, `tooltipRef` (as props for `Tooltip` component), but using in both of them the same `thumbRef.current` ref value? is it done to detect the point when the tooltip element is shown?

packages/picasso/src/SliderValueLabel/SliderValueLabel.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return (
+    <Tooltip
+      ref={thumbRef}
+      tooltipRef={handleTooltipRef}
+      content={value}
+      open={open || valueLabelDisplay === 'on'}
+      placement={placement()}
```
denieler, line 61:
```suggestion
      placement={getPlacement()}
```

denieler (approved):

packages/picasso/src/SliderValueLabel/SliderValueLabel.tsx
OleksandrNechai: Oh, come one :smile:

denieler: sorry, just looks bad as a component function 🤷 😃

PR #2115, opened 2021-07-08
fix(TagSelector): compare by value

george-aidonidis:
[FX-1995]

### Description

When setting initial values on `TagSelector` those (initial values) are not filtered out from the available choices.
[Sandbox example](https://codesandbox.io/s/tender-meadow-jjdol?file=/src/App.tsx)
This is happening because `isIncluded` function is comparing items by reference.

### How to test

- TagSelector component should work exactly the same as before

### Review

<details>
<br />

List of available commands:

</details>

[FX-1995]: https://toptal-core.atlassian.net/browse/FX-1995

### DIFF
--- a/packages/picasso/src/TagSelector/TagSelector.tsx
+++ b/packages/picasso/src/TagSelector/TagSelector.tsx
@@ -22,8 +22,11 @@ export interface Item extends AutocompleteItem {
 
 const EMPTY_INPUT_VALUE = ''
 
-const isIncluded = (items: Item[], currentItem: Item) =>
-  items.some(item => item === currentItem)
+export const isIncluded = (items: Item[], currentItem: Item) =>
+  items.some(
+    ({ text, value }) =>
+      text === currentItem.text && value === currentItem.value
+  )
 
 const getItemText = (item: Item | null) =>
   (item && item.text) || EMPTY_INPUT_VALUE
--- a/packages/picasso/src/TagSelector/TagSelector.tsx
+++ b/packages/picasso/src/TagSelector/TagSelector.tsx
@@ -23,10 +23,7 @@ export interface Item extends AutocompleteItem {
 const EMPTY_INPUT_VALUE = ''
 
 export const isIncluded = (items: Item[], currentItem: Item) =>
-  items.some(
-    ({ text, value }) =>
-      text === currentItem.text && value === currentItem.value
-  )
+  items.some(({ value }) => value === currentItem.value)
 
 const getItemText = (item: Item | null) =>
   (item && item.text) || EMPTY_INPUT_VALUE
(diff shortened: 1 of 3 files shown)


denieler (approved):

packages/picasso/src/TagSelector/TagSelector.tsx
```diff
@@ -22,8 +22,11 @@ export interface Item extends AutocompleteItem {
 
 const EMPTY_INPUT_VALUE = ''
 
-const isIncluded = (items: Item[], currentItem: Item) =>
-  items.some(item => item === currentItem)
+export const isIncluded = (items: Item[], currentItem: Item) =>
+  items.some(
+    ({ text, value }) =>
+      text === currentItem.text && value === currentItem.value
```
teimurjan, line 28: Do we need to compare by `text`? `value` seems to be the identificator of an `Item` so it's the only thing we need

packages/picasso/src/TagSelector/test.tsx
```diff
@@ -125,3 +125,21 @@ describe('TagSelector', () => {
     expect(onChange).toHaveBeenCalledWith([testOptions[0]])
   })
 })
+
+describe('#isIncluded', () => {
```
teimurjan, line 129:
```suggestion
describe('isIncluded', () => {
```

teimurjan: It's great that you've created a test for the util function. I think we also have to create a test for `TagSelector` itself to check that filtering works well in the situation where you've had the bug.

packages/picasso/src/TagSelector/TagSelector.tsx
george-aidonidis: Well, in theory we shouldn't but just wanted to be extra cautious. I don't think we will have a case where 2 objects can have same `value` but not `text`, unless you think otherwise

packages/picasso/src/TagSelector/test.tsx
george-aidonidis: Sure! Initially I tried to replicate the bug there but wasn't able to test it properly since it was complex to get the available options from the TagSelector, with RTL at least. That's why you only see the test for this simple function

I would love to see a test for the component itself, do you have any suggestion?

teimurjan: Added a Cypress test checking this functionality 🙂

teimurjan (approved):

packages/picasso/src/TagSelector/TagSelector.tsx
teimurjan: It's a nitpick for me, so feel free to do as you wish 🙂

george-aidonidis: Alright, let's remove it 😸. It doesn't seem plausible to have different text but same value

denieler (approved):

PR #2119, opened 2021-07-14
fix(Popper): fix popper shaking on each render

dmaklygin1:
[TOP-1523]

### Description

Sometimes Menu is shaking when a user is walking across a menu items. It happens due to creating a completely new instance of `Popper.js` on each render of the Picasso's `Popper`. To avoid this, we should avoid creating new instances of props that are passed to `MUI`'s `Popper`.

#### Before

https://monosnap.com/file/WRc2TdNnsGtlZwCXIWoUegTO2xuyvx

#### After

https://monosnap.com/file/JFlgVWiqBIsgu1zQmRARoRDNZaVQ0f

### Review

<details>
<br />

List of available commands:

</details>

[TOP-1523]: https://toptal-core.atlassian.net/browse/TOP-1523

### DIFF
--- a/packages/picasso/src/Popper/Popper.tsx
+++ b/packages/picasso/src/Popper/Popper.tsx
@@ -58,7 +58,7 @@ const getAnchorEl = (
   anchorEl: null | ReferenceObject | (() => ReferenceObject)
 ) => (typeof anchorEl === 'function' ? anchorEl() : anchorEl)
 
-const getPopperOptions = (popperOptions: PopperOptions) => ({
+export const getPopperOptions = (popperOptions: PopperOptions) => ({
   ...popperOptions,
 
   modifiers: {
@@ -131,6 +131,11 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper (props, ref) {
     }
   }, [isCompactLayout, open])
 
+  const memoizedPopperOptions = React.useMemo(
+    () => getPopperOptions(popperOptions),
+    [popperOptions]
+  )
+
   return (
     <MUIPopper
       open={open}
@@ -138,7 +143,7 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper (props, ref) {
       anchorEl={anchorEl}
       className={cx(classes.root, className)}
       popperRef={ref}
-      popperOptions={getPopperOptions(popperOptions)}
+      popperOptions={memoizedPopperOptions}
       disablePortal={disablePortal}
       style={{
         ...style,
--- a/packages/picasso/src/Popper/test.tsx
+++ b/packages/picasso/src/Popper/test.tsx
@@ -1,63 +1,159 @@
-import React, { useState, FC, ReactNode, forwardRef } from 'react'
-import Picasso from '@toptal/picasso-provider'
-import { render, act, fireEvent } from '@toptal/picasso/test-utils'
+import React from 'react'
+import MUIPopper from '@material-ui/core/Popper'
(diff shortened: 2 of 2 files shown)


packages/picasso/src/Popper/Popper.tsx
```diff
@@ -131,14 +131,19 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper (props, ref) {
     }
   }, [isCompactLayout, open])
 
+  const mergedPopperOptions = React.useMemo(
```
OleksandrNechai, line 134: why merged? Maybe `memoed`? 
```suggestion
  const memoedPopperOptions = React.useMemo(
```

packages/picasso/src/Popper/Popper.tsx
```diff
@@ -131,14 +131,19 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper (props, ref) {
...
+    () => getPopperOptions(popperOptions),
+    [popperOptions]
+  )
```
OleksandrNechai, line 137: Maybe it would be better to place this code closer to `getPopperOptions` definition?

dmaklygin1: what do you mean? `getPopperOptions` is defined outside of Popper component

packages/picasso/src/Popper/Popper.tsx
```diff
@@ -58,7 +58,7 @@ const getAnchorEl = (
   anchorEl: null | ReferenceObject | (() => ReferenceObject)
 ) => (typeof anchorEl === 'function' ? anchorEl() : anchorEl)
 
-const getPopperOptions = (popperOptions: PopperOptions) => ({
+export const getPopperOptions = (popperOptions: PopperOptions) => ({
```
denieler, line 61: what about adding `useMemo` here?

denieler (approved):

packages/picasso/src/Popper/Popper.tsx
dmaklygin1: it works, but I've got set of errors in the tests:

```

 Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
    See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.
```

packages/picasso/src/Popper/Popper.tsx
OleksandrNechai: Oh, ok :-)

packages/picasso/src/Popper/Popper.tsx
denieler: ah, yeah, but we can add here useMemo and move this function to the component maybe? or it will be not cool?

dmaklygin1: I don't think it's a good idea. The function will be created each time when Component is being rendered.

OleksandrNechai (approved):

packages/picasso/src/Popper/Popper.tsx
denieler: with useMemo shouldn't 🤷‍♂️ but the current solution also ok 👍

dmaklygin1: let's keep it then :)

PR #2122, opened 2021-07-27
feat(TreeView): [FX-2016] Horizontal layout, margin props & compact variant for TreeView

danganiev-tt:
[FX-2016]
[TOP-1535]

### Description

Added horizontal layout, vertical and horizontal margin props that allow to override default margins & compact variant of TreeView that allows for less bloated trees if tree has one or zero nodes with children per depth level

### How to test

'Horizontal layout' & 'Compacted tree' examples in TreeView storybook

### Screenshots

image

### Review

<details>
<br />

List of available commands:

</details>

[FX-2016]: https://toptal-core.atlassian.net/browse/FX-2016

[TOP-1535]: https://toptal-core.atlassian.net/browse/TOP-1535

### DIFF
--- a/packages/picasso/src/TreeView/PointLink/PointLink.tsx
+++ b/packages/picasso/src/TreeView/PointLink/PointLink.tsx
@@ -19,23 +19,26 @@ export const PointLink: FC<Props> = props => {
   const classes = useStyles()
   const path = useMemo(() => {
     const { source, target } = link
-    const sourceY =
-      source.y +
-      (direction === 'horizontal'
-        ? source.rect?.height / 2
-        : source.rect?.height || 0)
+    const sourceYDeltas = {
+      horizontal: source.rect?.height / 2,
+      vertical: source.rect?.height
+    }
+    const sourceY = source.y + (sourceYDeltas[direction] || 0)
     const sourceX = source.x + (source.rect?.width / 2 || 0)
 
-    return direction === 'horizontal'
-      ? `M${sourceX}, ${sourceY}
-            H${sourceX + horizontalMargin / 2}
-            V${target.y + DEFAULT_HEIGHT / 2}
-            H${target.x - DEFAULT_WIDTH / 2}`
-      : `M${source.x}, ${sourceY}
-               V${sourceY + verticalMargin / 2}
-               H${target.x}
-               V${target.y}`
-  }, [link, direction, verticalMargin])
+    const svgPaths = {
+      horizontal: `M${sourceX}, ${sourceY}
+        H${sourceX + horizontalMargin / 2}
+        V${target.y + DEFAULT_HEIGHT / 2}
+        H${target.x - DEFAULT_WIDTH / 2}`,
+      vertical: `M${source.x}, ${sourceY}
+        V${sourceY + verticalMargin / 2}
+        H${target.x}
+        V${target.y}`
+    }
+
+    return svgPaths[direction]
+  }, [link, direction, verticalMargin, horizontalMargin])
 
   return <path d={path} className={classes.pointLink} />
 }
--- a/packages/picasso/src/TreeView/PointLink/PointLink.tsx
+++ b/packages/picasso/src/TreeView/PointLink/PointLink.tsx
@@ -20,8 +20,8 @@ export const PointLink: FC<Props> = props => {
   const path = useMemo(() => {
     const { source, target } = link
     const sourceYDeltas = {
-      horizontal: source.rect?.height / 2,
-      vertical: source.rect?.height
+      horizontal: source.rect?.height / 2 || 0,
+      vertical: source.rect?.height || 0
     }
     const sourceY = source.y + (sourceYDeltas[direction] || 0)
     const sourceX = source.x + (source.rect?.width / 2 || 0)
--- a/packages/picasso/src/TreeView/PointLink/PointLink.tsx
+++ b/packages/picasso/src/TreeView/PointLink/PointLink.tsx
@@ -1,42 +1,41 @@
 import React, { FC, useMemo } from 'react'
 import { Theme, makeStyles } from '@material-ui/core/styles'
 
-import { DynamicPointLink } from '../types'
-import {
-  VERTICAL_MARGIN,
-  HORIZONTAL_MARGIN,
-  DEFAULT_HEIGHT,
-  DEFAULT_WIDTH
-} from '../variables'
+import { DirectionsType, DynamicPointLink } from '../types'
+import { DEFAULT_HEIGHT, DEFAULT_WIDTH } from '../variables'
 import styles from './styles'
 
 export interface Props {
   link: DynamicPointLink
-  isHorizontal?: boolean
+  direction: DirectionsType
+  verticalMargin: number
+  horizontalMargin: number
 }
 
 const useStyles = makeStyles<Theme>(styles, { name: 'PicassoPointLink' })
 
 export const PointLink: FC<Props> = props => {
-  const { link, isHorizontal } = props
+  const { link, direction, verticalMargin, horizontalMargin } = props
   const classes = useStyles()
   const path = useMemo(() => {
     const { source, target } = link
     const sourceY =
       source.y +
-      (isHorizontal ? source.rect?.height / 2 : source.rect?.height || 0)
+      (direction === 'horizontal'
+        ? source.rect?.height / 2
+        : source.rect?.height || 0)
     const sourceX = source.x + (source.rect?.width / 2 || 0)
 
-    return isHorizontal
+    return direction === 'horizontal'
       ? `M${sourceX}, ${sourceY}
-            H${sourceX + HORIZONTAL_MARGIN / 2}
+            H${sourceX + horizontalMargin / 2}
             V${target.y + DEFAULT_HEIGHT / 2}
             H${target.x - DEFAULT_WIDTH / 2}`
       : `M${source.x}, ${sourceY}
-               V${sourceY + VERTICAL_MARGIN / 2}
+               V${sourceY + verticalMargin / 2}
                H${target.x}
                V${target.y}`
-  }, [link, isHorizontal])
+  }, [link, direction, verticalMargin])
 
   return <path d={path} className={classes.pointLink} />
 }
--- a/packages/picasso/src/TreeView/PointLink/PointLink.tsx
+++ b/packages/picasso/src/TreeView/PointLink/PointLink.tsx
@@ -2,27 +2,41 @@ import React, { FC, useMemo } from 'react'
 import { Theme, makeStyles } from '@material-ui/core/styles'
 
 import { DynamicPointLink } from '../types'
-import { VERTICAL_MARGIN } from '../variables'
+import {
+  VERTICAL_MARGIN,
+  HORIZONTAL_MARGIN,
+  DEFAULT_HEIGHT,
+  DEFAULT_WIDTH
+} from '../variables'
 import styles from './styles'
 
 export interface Props {
   link: DynamicPointLink
+  isHorizontal?: boolean
 }
 
 const useStyles = makeStyles<Theme>(styles, { name: 'PicassoPointLink' })
 
 export const PointLink: FC<Props> = props => {
-  const { link } = props
+  const { link, isHorizontal } = props
   const classes = useStyles()
   const path = useMemo(() => {
     const { source, target } = link
-    const sourceY = source.y + (source.rect?.height || 0)
+    const sourceY =
+      source.y +
+      (isHorizontal ? source.rect?.height / 2 : source.rect?.height || 0)
+    const sourceX = source.x + (source.rect?.width / 2 || 0)
 
-    return `M${source.x}, ${sourceY}
-            V${sourceY + VERTICAL_MARGIN / 2}
-            H${target.x}
-            V${target.y}`
-  }, [link])
+    return isHorizontal
+      ? `M${sourceX}, ${sourceY}
+            H${sourceX + HORIZONTAL_MARGIN / 2}
+            V${target.y + DEFAULT_HEIGHT / 2}
+            H${target.x - DEFAULT_WIDTH / 2}`
+      : `M${source.x}, ${sourceY}
+               V${sourceY + VERTICAL_MARGIN / 2}
+               H${target.x}
+               V${target.y}`
+  }, [link, isHorizontal])
 
   return <path d={path} className={classes.pointLink} />
 }
--- a/packages/picasso/src/TreeView/TreeView.tsx
+++ b/packages/picasso/src/TreeView/TreeView.tsx
@@ -34,6 +34,8 @@ export interface Props {
   showZoom?: boolean
   /** Scales the current zoom transform by coefficient */
   scaleCoefficient?: number
+  /** Changes the layout of the tree from top-to-bottom to left-to-right */
+  isHorizontal?: boolean
 }
 
 const useStyles = makeStyles<Theme>(styles, { name: 'PicassoTreeView' })
@@ -46,11 +48,12 @@ export const TreeView = (props: Props) => {
     scaleExtent = DEFAULT_SCALE_EXTENT,
     initialScale = 1,
     scaleCoefficient = 0.5,
-    showZoom
+    showZoom,
+    isHorizontal
   } = props
   const classes = useStyles()
   const rootRef = createRef<SVGSVGElement>()
-  const { nodes, links, selectedNode } = useTree({ data })
+  const { nodes, links, selectedNode } = useTree({ data, isHorizontal })
 
   const center = useMemo<{ x: number; y: number } | undefined>(() => {
     if (!selectedNode) {
@@ -97,6 +100,7 @@ export const TreeView = (props: Props) => {
             <PointLink
               link={link}
               key={`link-${link.target.data.id}-${link.source.data.id}`}
+              isHorizontal={isHorizontal}
             />
           ))}
           {nodes.map(node => (
--- a/packages/picasso/src/TreeView/TreeView.tsx
+++ b/packages/picasso/src/TreeView/TreeView.tsx
@@ -17,7 +17,8 @@ import PointNode from './PointNode'
 import styles from './styles'
 import { useZoom } from './useZoom'
 import { DEFAULT_SCALE_EXTENT, DEFAULT_WIDTH } from './variables'
-import { TreeNodeInterface } from './types'
+import { DirectionsType, TreeNodeInterface, TreeViewVariant } from './types'
+import { getDefaultMargins } from './utils'
 
 export interface Props {
   /** Root node of the Tree */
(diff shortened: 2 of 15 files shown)


packages/picasso/src/TreeView/TreeView.tsx
```diff
@@ -34,6 +34,8 @@ export interface Props {
   showZoom?: boolean
   /** Scales the current zoom transform by coefficient */
   scaleCoefficient?: number
+  /** Changes the layout of the tree from top-to-bottom to left-to-right */
+  isHorizontal?: boolean
```
dmaklygin1, line 38: I would add a `direction` prop, instead of `isHorizontal` with two possible options: vertical, horizontal

dmaklygin1:
@toptal-bot run all

packages/picasso/src/TreeView/TreeView.tsx
danganiev-tt: Agreed, this would allow for better expandability later, if for example somebody would need something like inverse vertical layout (bottom to top)

packages/picasso/src/TreeView/PointLink/PointLink.tsx
```diff
@@ -1,28 +1,41 @@
...
-    const sourceY = source.y + (source.rect?.height || 0)
-
-    return `M${source.x}, ${sourceY}
-            V${sourceY + VERTICAL_MARGIN / 2}
-            H${target.x}
-            V${target.y}`
-  }, [link])
+    const sourceY =
+      source.y +
+      (direction === 'horizontal'
```
TomasSlama, line 24: U changed `isHorizontal` to `direction`, but many times you are using it still as isHorizontal.

This will of course work, but I personally like it prepared for an extension.

For example:
```js
const rectHeightByDirection = {
  horizontal: source.rect?.height / 2,
  vertical: source.rect?.height 
}
const source Y = source.y + (rectHeightByDirection[direction] || 0)
```

What do you think about it?

danganiev-tt: Sure, this is a better way. There might be similar code improvements with `variant` too.

danganiev-tt:
image

Found a bug, do not merge yet

packages/picasso/src/TreeView/story/Horizontal.example.tsx
```diff
@@ -0,0 +1,202 @@
...
+}
+
+const Example = () => (
+  <Container style={{ height: '25em' }}>
+    <TreeView
+      data={rootNode}
+      renderNode={renderNode}
+      initialScale={0.8}
+      verticalMargin={20}
+      direction='horizontal'
```
OleksandrNechai, line 197: What makes us to name prop "horizontal direction", but example to name "horizontal layout"?

OleksandrNechai (approved):
Could not understand most of the code. 
But I have now a strong feeling that this component should go to TopTeam repo if it is not used in other projects and is not part of the base library. WDYT @havenchyk

packages/picasso/src/TreeView/story/Compacted.example.tsx
```diff
@@ -0,0 +1,209 @@
+import React from 'react'
```
OleksandrNechai, line 1: To me, the name of the example `Compact` sounds better :-)

packages/picasso/src/TreeView/TreeView.tsx
```diff
@@ -34,6 +35,14 @@ export interface Props {
   showZoom?: boolean
   /** Scales the current zoom transform by coefficient */
   scaleCoefficient?: number
+  /** Determines the direction of the TreeView: vertical - top-to-bottom (default) or */
+  direction?: DirectionsType
+  /** Overrides default vertical margin - minimum vertical distance between nodes */
+  verticalMargin?: number
+  /** Overrides default horizontal margin - minimum horizontal distance between nodes */
+  horizontalMargin?: number
```
OleksandrNechai, line 43: Its too many properties to my taste. Maybe we should collect some of them into config object?

packages/picasso/src/TreeView/TreeView.tsx
```diff
@@ -46,11 +55,27 @@ export const TreeView = (props: Props) => {
     scaleExtent = DEFAULT_SCALE_EXTENT,
     initialScale = 1,
     scaleCoefficient = 0.5,
-    showZoom
+    showZoom,
+    direction = 'vertical',
+    verticalMargin,
+    horizontalMargin,
```
OleksandrNechai, line 61: No default values for these?

packages/picasso/src/TreeView/story/Compacted.example.tsx
OleksandrNechai: How compact differs from non-compact? To me, it looks like it's a matter of zooming to make the tree compact.

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,36 +31,115 @@ export interface UseTreeResponse {
...
-      node.children.reduce((acc, child) => {
-        calculateNodeXPosition(child, { leaves, nodeWidth })
-
-        acc += child.x
-
-        return acc
-      }, 0) / node.children.length
+type NodeAggregationType = 'siblings' | 'leaves'
+
+const calculateNodeXorYPosition = (isX: boolean) => {
```
OleksandrNechai, line 36: This function calculates nothing, just returns another function. What it exists for? It is called calculate? Is not `boolean` param a well-known antipattern?

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,36 +31,115 @@ export interface UseTreeResponse {
...
+    node: HierarchyPointNode<TreeNodeInterface>,
+    {
+      leaves,
+      nodeSizeAttr,
+      aggregationType = 'leaves'
+    }: {
+      leaves: HierarchyPointNode<TreeNodeInterface>[]
+      nodeSizeAttr: number
+      aggregationType: NodeAggregationType
+    }
```
OleksandrNechai, line 47: It would be easier to read if types were extracted

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,36 +31,115 @@ export interface UseTreeResponse {
...
+  ) => {
+    let index = 0
+    let position = 0
+    const xOrY = isX ? 'x' : 'y'
+
+    if (!node.children || !node.children.length) {
```
OleksandrNechai, line 53: I wonder, is not linter failing for this code? It is too large and too complicated. I think it should be refactored so someone could understand it. I have not seen 4 nested `if`s for quite some time already :sweat_smile:

packages/picasso/src/TreeView/utils/getDefaultMargins.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (verticalMargin) {
+    finalVMargin = verticalMargin
+  }
+
+  if (horizontalMargin) {
+    finalHMargin = horizontalMargin
+  }
+
+  return [finalVMargin, finalHMargin]
+}
```
OleksandrNechai, line 32: Is everything that this code does is `return [verticalMargin || defaultVMargin, horizontalMargin || defaultHMargin]`?

packages/picasso/src/TreeView/utils/getDefaultMargins.ts
```diff
@@ -0,0 +1,32 @@
...
+  DEFAULT_HORIZONTAL_MARGIN_V
+} from '../variables'
+import { DirectionsType } from '../types'
+
+const defaultMargins = {
+  horizontal: [DEFAULT_VERTICAL_MARGIN_H, DEFAULT_HORIZONTAL_MARGIN_H],
+  vertical: [DEFAULT_VERTICAL_MARGIN_V, DEFAULT_HORIZONTAL_MARGIN_V]
+}
+
+export const getDefaultMargins = (
```
OleksandrNechai, line 14: This function does not do what it says it does.

OleksandrNechai (changes requested):
After the second read, I realized the quality of the code is not right. It's too complex, very hard to read, badly structured and many functions are telling lies, meaning they don't do what they say they do.

havenchyk (approved):
Topteam is the only user of TreeView, so I'm fine with this change

packages/picasso/src/TreeView/story/Horizontal.example.tsx
danganiev-tt: Yes, I should choose one or the other and make the code and definitions more uniform

packages/picasso/src/TreeView/TreeView.tsx
danganiev-tt: Limiting to variant is not what I had in mind, since I have a use case where default margin had to be overridden, so I tried to implement the solution in a very universal way, where you could override any margin for any tree variant.

I agree about the collecting properties into a config option though. I should only collect my new options not to break current component interface

packages/picasso/src/TreeView/useTree.ts
danganiev-tt: It was not a well-known antipattern _to me_, but after some googling it is, and I'm starting to catch up :)

packages/picasso/src/TreeView/useTree.ts
danganiev-tt: I completely agree, it's probably almost incomprehensible in its complexity, I'll try to refactor and simplify it as much as I can

packages/picasso/src/TreeView/TreeView.tsx
danganiev-tt: I'll check if this, and the issues within the `packages/picasso/src/TreeView/utils/getDefaultMargins.ts` file can be somehow solved in combination

packages/picasso/src/TreeView/utils/getDefaultMargins.ts
danganiev-tt: I'll either rename the function or rethink it's code to match the name then

packages/picasso/src/TreeView/PointLink/PointLink.tsx
```diff
@@ -1,28 +1,44 @@
...
     const { source, target } = link
-    const sourceY = source.y + (source.rect?.height || 0)
-
-    return `M${source.x}, ${sourceY}
-            V${sourceY + VERTICAL_MARGIN / 2}
-            H${target.x}
-            V${target.y}`
-  }, [link])
+    const sourceYDeltas = {
+      horizontal: source.rect?.height / 2,
```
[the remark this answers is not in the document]
danganiev-tt: No, but I will

packages/picasso/src/TreeView/utils/getDefaultMargins.ts
danganiev-tt: I guess 😅

packages/picasso/src/TreeView/story/Compacted.example.tsx
danganiev-tt: Compact is the TreeView that is rendered in a specific way. It **only** works for trees in which only one node per depth has children, which also happens to be the only use case we have in Top Team 🙃 Here is the description I made in the `-frontend-exp-core` slack channel:

> Basically we have a tree in which only one node on each level of depth can have children. Current TreeView creates big space gaps around tree nodes which have children, to account for overall width/height (depending on layout) of the whole tree. This makes sense as it prevents nodes overlapping each other (maybe the picture at the bottom will help in understanding this better)
> But in our case we know for sure that overlapping won’t happen, and we do not want these gaps to exist. So I want to add an optional prop like allowOverlap?:boolean to the TreeView that will prevent addition of space gaps like the ones I described. 

Then @havenchyk suggested the `variant: compact` prop & name instead of `allowOverlap`, and I couldn't think of anything better, so I just went with it.

image

dmaklygin1:
> Topteam is the only user of TreeView, so I'm fine with this change

Nope, Staff portal also uses TreeView: https://github.com/toptal/staff-portal/blob/ebb05a0f5de92a0f69e6050edc3308b969bec85e/src/modules/summary-sidebar/components/StaffTreeModal/StaffTreeModal.tsx

dmaklygin1:
I was playing locally, and just assigned horizontal direction to the `With selected node` example:

```
directionOptions={{
  verticalMargin: 20,
  direction: 'horizontal'
}}
```


Picasso | TreeView 🔊 2021-09-01 12-04-01

packages/picasso/src/TreeView/TreeView.tsx
```diff
@@ -46,11 +64,32 @@ export const TreeView = (props: Props) => {
...
+    showZoom,
+    directionOptions
   } = props
+
+  const {
+    direction = 'vertical',
+    verticalMargin = DEFAULT_VERTICAL_MARGIN_V,
+    horizontalMargin = DEFAULT_HORIZONTAL_MARGIN_V,
+    variant = 'normal'
+  } = directionOptions || {}
```
dmaklygin1, line 76: wdyt about adding default values to the `defaultProps `?

dmaklygin1:
> I was playing locally, and just assigned horizontal direction to the `With selected node` example:
> 
> ```
> directionOptions={{
>   verticalMargin: 20,
>   direction: 'horizontal'
> }}
> ```
> 
> Picasso | TreeView 🔊 2021-09-01 12-04-01

isn't it because defaultHeight wasn't specified? if yes, then this is okay

dmaklygin1 (approved):

danganiev-tt:
@dmaklygin1 the default height should not be a problem, this looks like a bug. Although maybe adding default values to the `defaultProps` like you suggested might solve this. Anyway, I'll take a look

packages/picasso/src/TreeView/TreeView.tsx
danganiev-tt: I'm against it because this causes a lot of issues with Typescript due to `directionOptions` being a possible undefined type. Also, `defaultProps` will be eventually deprecated for functional components according to Dan Abramov https://twitter.com/dan_abramov/status/1133878326358171650?lang=en, so it's better to use object default values (but this also causes same Typescript issues here)

danganiev-tt:
@toptal-bot run all

danganiev-tt:
@toptal-bot run happo

danganiev-tt:

danganiev-tt:

danganiev-tt:

danganiev-tt:

danganiev-tt:

danganiev-tt:

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,52 +28,198 @@ export interface UseTreeResponse {
...
+
+      acc += child[coordinateType]
+
+      return acc
+    }, 0) / node.children.length
+
+  return position
+}
+
+const getPositionNoLeavesButChildren = (
```
havenchyk, line 101: this is still overcomplicated, but I don't see any better solution

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,52 +28,198 @@ export interface UseTreeResponse {
...
+  if (childWithChildrenIndex !== -1) {
+    calculateNodePosition({
+      ...options,
+      node: node.children[childWithChildrenIndex]
+    })
+  }
+
+  node.children.reduce((acc, child) => {
+    if (!child.children) {
+      calculateNodePosition({ ...options, node: child })
```
havenchyk, line 127: inside calculateNodePosition are again check on `node.children`, can it be simplified?

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,52 +28,198 @@ export interface UseTreeResponse {
...
+  )
+
+  if (childWithChildrenIndex !== -1) {
+    calculateNodePosition({
+      ...options,
+      node: node.children[childWithChildrenIndex]
+    })
+  }
+
+  node.children.reduce((acc, child) => {
```
havenchyk, line 125: I do not understand why do you use `reduce` here if you don't change `acc` at all. Can it be `forEach`?

packages/picasso/src/TreeView/useTree.ts
```diff
@@ -17,52 +28,198 @@ export interface UseTreeResponse {
...
+  const halfLength = node.children.length / 2
+
+  if (node.children.length % 2) {
+    position = node.children[Math.floor(halfLength)][coordinateType]
   } else {
-    node.x =
-      node.children.reduce((acc, child) => {
-        calculateNodeXPosition(child, { leaves, nodeWidth })
+    position =
+      (node.children[halfLength][coordinateType] +
```
havenchyk, line 139: nit: combination of middle, right, left words for variable name could help to understand this code better

packages/picasso/src/TreeView/TreeView.tsx
```diff
@@ -114,14 +179,6 @@ export const TreeView = (props: Props) => {
   )
 }
 
-TreeView.defaultProps = {
```
havenchyk, line 117: as I mentioned, we use defaultProps for storybook rendering. can you verify that with removing you didn't miss some important information on Props table?

packages/picasso/src/TreeView/TreeView.tsx
danganiev-tt: if there is a specific reason of using defaultProps then I'll backtrack the changes. Still the values of directionOptions need to be overridden inside the component even if I'll declare them in default props

packages/picasso/src/TreeView/useTree.ts
danganiev-tt: Probably just because I used reduce in `getPositionLeavesAndChildren`

packages/picasso/src/TreeView/useTree.ts
danganiev-tt: If you're talking about the `position` variable, then `rightPosition` and `leftPosition` would be wrong here in case of regular vertical tree (it should be `topPosition` and `bottomPosition` then).

packages/picasso/src/TreeView/story/Compact.example.tsx
```diff
@@ -0,0 +1,211 @@
...
+  TreeNodeInterface,
+  TreeNodeAvatar,
+  Container,
+  UserBadge
+} from '@toptal/picasso'
+import { HierarchyPointNode } from 'd3-hierarchy'
+import styled from 'styled-components'
+import { palette } from '@toptal/picasso/utils'
+
+const NodeContainer = styled<typeof Container>(Container)<{
```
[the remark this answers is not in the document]
danganiev-tt: I just copied all examples from` Default.example.tsx` 🤷🏻‍♂️. If we want to change this, it needs to be done in all of the examples I suppose

packages/picasso/src/TreeView/TreeView.tsx
danganiev-tt: Also I'll move the default props from function component to `defaultProps` to avoid duplication

danganiev-tt: Actually I had to duplicate `defaultProps` as object default values instead

OleksandrNechai (approved):

packages/picasso/src/TreeView/useTree.ts
danganiev-tt: This is done to not recalculate the whole tree for the child with children again, since it's done before.

TomasSlama (approved):

packages/picasso/src/TreeView/story/Compact.example.tsx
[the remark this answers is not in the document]
havenchyk: got it, keep it like this in this case

havenchyk (approved):
LGTM, thanks for your work!

PR #2141, opened 2021-08-19
style(Link): update to BASE 2.0 design

TomasSlama:
[FX-1963]

### Description

Adopt the Link component to the latest design.

changes in API:
 - replace prop `underline` with `textDecoration`

when we want some wrapping content to be clickable.
 - new prop disabled
   - `not-allowed` cursor on hover
   - removes `onClick` and `href`
   - adds `aria-disabled`  

### Screenshots

| image | image |

### Review

<details>
<br />

List of available commands:

</details>

[FX-1963]: https://toptal-core.atlassian.net/browse/FX-1963


packages/picasso/src/Link/Link.tsx
```diff
@@ -38,7 +38,7 @@ export type Props = BaseProps &
     children?: ReactNode
     /** Destination the link points to */
     href?: string
-    /** Controls when the link should have an underline */
+    /** Controls when the link should have an underline. Use 'none' only when Link is container. */
     underline?: UnderlineType
```
teimurjan, line 42: If it's a one-value prop, why not make it boolean (ex. `container?: boolean`)?

packages/picasso/src/Link/story/Underline.example.tsx
```diff
@@ -1,24 +1,34 @@
...
-      </Link>
-    </Container>
-    <Container inline>
-      <Link href='https://toptal.com' underline='always'>
-        Link with `underline: always`
-      </Link>
-    </Container>
-  </div>
-)
+const UnderlineExample = () => {
```
teimurjan, line 11: When I look at an example with **Underline** title I hope to see a link with an underline. Instead, I see a custom user badge link.

packages/picasso/src/Link/styles.ts
```diff
@@ -12,17 +12,31 @@ PicassoProvider.override(() => ({
 export default ({ typography, palette }: Theme) =>
   createStyles({
     root: {
+      textDecoration: 'underline',
       '&:focus': {
         outline: `1px dotted ${palette.blue.main}`
+      },
+      '&:hover': {
+        '&:not([aria-disabled="true"])': {
```
teimurjan, line 20:
```suggestion
        '&:not($disabled)': {
```

packages/picasso/src/Link/Link.tsx
teimurjan: Another idea is to move this prop to `VariantType`. (ex. `type VariantType = 'action' | 'anchor' | 'container'` or `type VariantType = 'action' | 'anchor' | 'wrapper'`)

teimurjan (commented):
1. Can you add a link to [Abstract]((https://share.goabstract.com/6b34cc9e-5316-4703-a7a0-d01d24b1ae17?mode=design&sha=c1f02ee345bc43e9617f5339aaa3eff1b2501d85)) on the Link's Storybook page, please?
2. Is `font-size` for link 14px? Looks like it's larger.

packages/picasso/src/Link/Link.tsx
TomasSlama: I wanted to keep it for back compatibility. If it would be a new prop I would definitely choose `variant`. But I was afraid that everybody who used this prop would have to make changes.

packages/picasso/src/Link/story/Underline.example.tsx
TomasSlama: I wanted to showcase that underline should be used for cases like this. `<Link underline>verify</Link>` is against base design.

packages/picasso/src/Link/Link.tsx
teimurjan: @TomasSlama hm, sounds reasonable. @havenchyk is it OK for you?

packages/picasso/src/Link/story/Underline.example.tsx
teimurjan: I'd rename it to `Container` and add a paragraph about the restrictions (whenever you need to use `Link` as a container, set `underline` to `none`).

packages/picasso/src/Link/Link.tsx
TomasSlama: @teimurjan maybe it would be great if We support both versions and somehow make the underline deprecated so we could remove it so we have the API right in the future.

packages/picasso/src/Link/styles.ts
TomasSlama: done

packages/picasso/src/Link/story/Underline.example.tsx
TomasSlama: done

packages/picasso/src/Link/styles.ts
```diff
@@ -13,13 +13,17 @@ export default ({ typography, palette }: Theme) =>
...
       '&:focus': {
         outline: `1px dotted ${palette.blue.main}`
       },
       '&:hover': {
-        '&:not([aria-disabled="true"])': {
+        '&:not($disabled)': {
           textDecoration: 'none'
         }
+      },
+      'div > &': {
```
teimurjan, line 25: Hmm, this selector doesn't look reliable to me. What if a link will be rendered in `Typography` with `as="div"`? IMO we should have an optional prop `size` with values `'default' | 'inherit'`.

TomasSlama: you are right, that looks much better. It should be breaking change probably. If you use the `Link` inside heading you have to set the `size` prop.

Now it makes sense to change the underline to `variant='container'` because we will have breaking change anyway.

havenchyk (changes requested):
image

Please, fix the type errors in development, squash commits locally, and mark them as a breaking change. Change the underline property type to boolean (maybe `noUnderline` is better or `showUnderline='false'`)

packages/picasso/src/Link/Link.tsx
```diff
@@ -82,19 +85,21 @@ export const Link: OverridableComponent<Props> = forwardRef<
     <MUILink
       {...nativeHTMLAttributes}
       ref={ref}
-      href={href}
-      target={target}
+      href={disabled ? undefined : href}
```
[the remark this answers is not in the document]
TomasSlama: I think so, if we would just use `pointerEvents: none`, users with basic knowledge of CSS can just turn it off in developer tools.

packages/picasso/src/Button/story/Augmentation.example.tsx
```diff
@@ -2,25 +2,27 @@ import React from 'react'
 import { Button, Link } from '@toptal/picasso'
 import { Settings16 } from '@toptal/picasso/Icon'
 
+const ActionLink = props => <Link {...props} variant='action' />
+
 const ButtonAugmentationExample = () => (
   <div>
-    <Button as={Link} href='/#home'>
+    <Button as={ActionLink} href='/#home'>
```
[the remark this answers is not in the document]
TomasSlama: sadly I can't, variant is used by Button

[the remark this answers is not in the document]
havenchyk: good point :)

packages/picasso/src/Link/Link.tsx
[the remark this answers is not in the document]
havenchyk: this is not a problem, but `pointersEvent: none` is worse solution to me by many reasons

packages/picasso/src/Link/Link.tsx
```diff
@@ -10,11 +10,12 @@ import cx from 'classnames'
...
 import { BaseProps, OverridableComponent } from '@toptal/picasso-shared'
 
+import { capitalize } from '../utils'
 import styles from './styles'
 
-type UnderlineType = 'none' | 'hover' | 'always'
-type VariantType = 'action' | 'anchor'
+type VariantType = 'action' | 'anchor' | 'container'
 type ColorType = 'white' | 'blue' | 'black'
+type SizeType = 'default' | 'inherit'
```
havenchyk, line 18: https://github.com/toptal/picasso/blob/master/docs/contribution/component-api.md#a-use-a-size-prop-with-an-enum-of-limited-values

packages/picasso/src/Link/Link.tsx
```diff
@@ -82,19 +89,22 @@ export const Link: OverridableComponent<Props> = forwardRef<
...
-      onClick={onClick}
+      onClick={disabled ? undefined : onClick}
       className={cx(classes.root, className, {
         [classes.action]: variant === 'action',
+        [classes.container]: variant === 'container',
         [classes.white]: color === 'white',
-        [classes.black]: color === 'black'
+        [classes.black]: color === 'black',
+        [classes.disabled]: disabled,
+        [classes[`size${size ? capitalize(size) : 'Default'}`]]: size
```
havenchyk, line 102: too dynamic 😄

havenchyk:
@teimurjan please take a look once again

packages/picasso/src/Link/Link.tsx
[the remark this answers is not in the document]
TomasSlama: If we want Link to be disabled, I think there should be no `href` or `onClick` available. So it actually does nothing. Don't you agree?

packages/picasso/src/Link/Link.tsx
TomasSlama: So If I understand correctly, I shouldn't use `size` prop because users are used to using values from SizeType enum, right?

So I should use some custom prop? Because values in SizeType don't make sense here. We want to give to users the choice to use default font size or inherit font size from a parent.

havenchyk: exactly, it's not about size, it's about font size, so font size can inherit or not

havenchyk (approved):

packages/picasso/src/Link/Link.tsx
```diff
@@ -38,8 +38,6 @@ export type Props = BaseProps &
     children?: ReactNode
     /** Destination the link points to */
     href?: string
-    /** Controls when the link should have an underline */
-    underline?: UnderlineType
```
denieler, line 42: looks like a breaking change, right?

denieler (commented):
Do you know why the change for file permissions has being made?

```
Notification/__image_snapshots__/notification-closable-notification-snap.png  100644 → 100755
```

packages/picasso/src/Link/Link.tsx
havenchyk: image
yes 😄

denieler: cool 👍

packages/picasso/src/Link/story/Container.example.tsx
```diff
@@ -0,0 +1,41 @@
...
+  Typography,
+  ChevronRight16
+} from '@toptal/picasso'
+
+const ContainerExample = () => {
+  return (
+    <>
+      <Container padded='medium' bottom='medium'>
+        <Typography>
+          Whenever you need to use Link as a container, use 'container' variant
```
denieler, line 16: this phrase is not clear to me 😞 what does it mean to use a Link as a container? is it about wrapping other components inside the link and behave as a link when clicking on this wrapped group of components?

TomasSlama: yeah, you got it right. Do you think we should rephrase it?

denieler: yeah, I would change it to something like "If you need to wrap a group of components or a single component into a Link - use the 'container' variant."

havenchyk (approved):

denieler (approved):
🎉 great work 💪 🙇‍♂️

TomasSlama:
@havenchyk cypress tests for Tooltip were not working for me locally. And one sometimes didn't work here. So I have updated those tests so they work every time.

havenchyk (approved):

PR #2148, opened 2021-09-01
chore: [FX-2064] update davinci

TomasSlama:
[FX-2064]

### Description

This is follow up task to creating new eslint rule in Davinci, where we forbid to use `@toptal/picasso-shared` in other projects. We need this rule to be disabled in Picasso.

With upgrade from v1 to v5, I had to fix codebase to satisfy rule `no-extraneous-dependencies`
  - every workspaces has its own dependencies
  - added [syncpack](https://github.com/JamieMason/syncpack) pre-commit hook to check if all dependencies across workspaces are in sync. It will list all unsync packages, we can sync them manually or by running `yarn syncpack:fix`

### How to test

- change version of React in only one package.json and try it to commit.
- tested only on Linux

### Screenshots
image

### Review

- n/a Annotate all `props` in component with documentation
- n/a Create `examples` for component
- n/a Ensure that deployed demo has expected results and good examples

<details>
<br />

List of available commands:

</details>

[FX-2064]: https://toptal-core.atlassian.net/browse/FX-2064


packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -160,7 +160,7 @@ const BarChart = <K extends string>({
               {data.map((entry, index) => {
                 const fill = getBarColor?.({ dataKey, entry, index })
 
-                return <Cell key={index} fill={fill} />
+                return <Cell key={`${entry.value}-${entry.name}`} fill={fill} />
```
vedrani, line 163: I have concern about it, what if values repeat?

packages/picasso-provider/package.json
```diff
@@ -20,9 +20,24 @@
...
+    "@material-ui/utils": "4.10.2",
+    "classnames": "^2.3.1",
     "notistack": "1.0.5",
     "react-helmet": "6.1.0"
+  },
+  "devDependencies": {
+    "@types/classnames": "^2.3.1",
+    "shelljs": "^0.8.4"
+  },
+  "resolutions": {
```
vedrani, line 38: Why do you need it?

packages/picasso/src/Dropdown/story/LongMenuList.example.tsx
```diff
@@ -33,8 +33,8 @@ const menuItems = [
 
 const menu = (
   <Menu>
-    {menuItems.map((itemName, index) => (
-      <Menu.Item onClick={handleClick} key={`menu-item-${index}`}>
+    {menuItems.map(itemName => (
+      <Menu.Item onClick={handleClick} key={itemName}>
```
vedrani, line 37: What if you have 2 items with same name?

packages/picasso-provider/package.json
TomasSlama: shelljs is used in `picasso-provider/bin/build.js` So the linter was throwing an error.

If you are asking about resolutions, problem is that `@material-ui/core` has `@material-ui/styles@^4.10.0`, so it actually installs a higher version than 4.10.0.  But with the higher version, I would have to do other changes in the codebase, because tests didn't pass. So I think it should be in a separate task.

packages/picasso/src/Dropdown/story/LongMenuList.example.tsx
TomasSlama: yeah, you are right, it will throw error in console to have uniq keys. But using index isn't ideal too, I'm thinking what would be the best approach.

packages/picasso-charts/src/BarChart/BarChart.tsx
TomasSlama: it is the same problem as in MenuList.

.eslintrc.js
```diff
@@ -24,6 +24,9 @@ const generateConfig = () =>
 
 module.exports = {
   extends: './node_modules/@toptal/davinci-syntax/src/configs/.eslintrc',
+  rules: {
+    '@toptal/davinci/no-private-package-imports': 'off'
```
havenchyk, line 28: shouldn't it go to `overrides` section only?

packages/topkit-analytics-charts/package.json
```diff
@@ -21,9 +21,19 @@
...
+    "@types/d3-array": "2.8.0",
+    "@types/d3": "^6.3.0",
     "date-fns": "^2.9.0"
+  },
+  "dependencies": {
+    "d3": "^6.6.2",
+    "d3-array": "^2.12.1"
+  },
+  "resolutions": {
+    "d3-array": "^2.12.1"
```
havenchyk, line 37: can you explain keeping resolutions here?

packages/shared/package.json
```diff
@@ -22,17 +22,17 @@
     "url": "https://github.com/toptal/picasso/issues"
   },
   "peerDependencies": {
-    "react": "^16.12.0",
-    "react-dom": "^16.12.0"
+    "react": "^17.0.1",
```
havenchyk, line 25: do we have to require 17 and not 16? did picasso users update react to 17?

packages/picasso/src/Tabs/story/ScrollButtons.example.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-array-index-key */
```
havenchyk, line 1: can we fix such problems instead of having eslint-disable?

packages/picasso/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-array-index-key */
```
havenchyk, line 1: is it possible to fix index here as well?

packages/picasso/src/Autocomplete/story/DynamicOptions.example.tsx
```diff
@@ -51,11 +51,11 @@ const Example = () => {
 
   const handleChange = (
     inputValue: string,
-    options: { isSelected: boolean }
+    handleOptions: { isSelected: boolean }
```
havenchyk, line 54: why so?

packages/picasso/package.json
```diff
@@ -23,38 +23,48 @@
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "peerDependencies": {
     "@material-ui/core": "4.11.0",
+    "@material-ui/styles": "4.10.0",
     "notistack": "1.0.5",
-    "react": "^16.12.0",
-    "react-dom": "^16.12.0",
+    "react": "^17.0.1",
```
havenchyk, line 28: did we decide to upgrade react?

packages/picasso-provider/package.json
havenchyk: I do not get the problem. Can you clarify why exactly you need to have versions specified in resolutions?  with yarn.lock we track which version of material-ui/styles is installed

package.json
```diff
@@ -124,7 +123,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged && (syncpack list-mismatches || (echo \"\n\nyou can use \\033[0;32myarn syncpack:fix \\033[0mto sync versions automatically or do it manually\n\n\" && return 1))"
```
havenchyk, line 126: `return 1`? 

also can't this command be a part of lint-staged? or do you want to run it for every changed file?

.eslintrc.js
TomasSlama: why do you think so? It seems cleaner this way to me.

TomasSlama: disabling it in general and only allow it for examples, since they are showcasing usage outside Picasso.

packages/picasso/package.json
TomasSlama: You are right, we did not, my bad. I will downgrade it.

package.json
TomasSlama: `return 1` - exit with error.

syncpack checks all package.json no matter in what files changes were made.

packages/picasso/src/Autocomplete/story/DynamicOptions.example.tsx
TomasSlama: there was eslint issue, options are already defined in upper scope

packages/picasso/src/CheckboxGroup/CheckboxGroup.tsx
TomasSlama: used "hack" `String(index)`

packages/shared/package.json
TomasSlama: fixed

packages/topkit-analytics-charts/package.json
TomasSlama: you are right, there is no need for resolutions.

packages/picasso-provider/package.json
vedrani: @TomasSlama I still think we need to check/discuss this. I really don't think resolution is solution. How much work is to do it without resolution?

packages/picasso/src/FileList/FileList.tsx
```diff
@@ -40,7 +40,7 @@ export const FileList = forwardRef<HTMLDivElement, Props>(function FileList (
           index={index}
           onRemove={onItemRemove}
           disabled={disabled}
-          key={`${fileUpload.file.name}-${index}`}
+          key={fileUpload.file.name}
```
vedrani, line 43: Do we have some other data field beside name? Maybe path? What if I want to upload 2 diff files with same name from different location?

packages/picasso/src/TypographyLoader/TypographyLoader.tsx
```diff
@@ -35,7 +35,7 @@ const Paragraph = ({ className, style = {} }: BaseProps) => (
 export const TypographyLoader = ({ className, rows = 1, style }: Props) => (
   <>
     {Array.from({ length: rows }).map((_, index) => (
-      <Paragraph className={className} key={index} style={style} />
+      <Paragraph className={className} key={String(index)} style={style} />
```
vedrani, line 38: What about `{String(index)-String{rows}}`?

package.json
havenchyk: I understand the reason of `return 1`, I'm interested in why do you need to explicitly return some value and why don't you rely on syncpack if it's a responsibility of the tool to return some exit code

.eslintrc.js
havenchyk: So do you think we should have this rule disabled in rules?

packages/picasso/src/TypographyLoader/TypographyLoader.tsx
TomasSlama: will it actually change anything? `rows` will have the same value for every item.

the output would be: 
1. `0-1`
2. `1-1`
3.  ...

.eslintrc.js
TomasSlama: As I understand it, overrides should be used for some glob pattern. But we want to disable that rule for all files, but `*.example.tsx`.

So if we would have it in overrides, we would have to select all files like `['*.tsx', '*.jsx', '*.ts', '*.js']`, and then disable rule. Having it in rules does the same and we don't have to worry, that we will forget to select some files.

package.json
TomasSlama: I wanted to add the hint to users, what can they do when the error appears.

So if syncpack returns that some of the packages need to be sync, at the bottom I echo them hint how to solve it, so they don't have to look it up.

packages/picasso-provider/package.json
TomasSlama: Guys you were right, resolution is not needed here at all.

packages/picasso/src/FileList/FileList.tsx
TomasSlama: I have added `size` and `lastModified` and even `String(index)` if someone would actually upload the same file twice.

packages/picasso/src/TypographyLoader/TypographyLoader.tsx
vedrani: If you change number of rows, then all items would remount, tbh not sure do we need it :)

vedrani (approved):
Thx for you effort 🙇

havenchyk (approved):

TomasSlama:
@toptal-bot run all

TomasSlama:
@toptal-bot run all

PR #2156, opened 2021-09-06
feat: expose vertical orientation property in tabs

pkondzior:
Component: https://picasso.toptal.net/?path=/story/layout-tabs--tabs

### Description

This change exposes orientation property from MaterialUi Tabs component. 

### Screenshots

### Review

<details>
<br />

List of available commands:

</details>

### DIFF
--- a/packages/picasso/src/Tab/styles.ts
+++ b/packages/picasso/src/Tab/styles.ts
@@ -19,9 +19,7 @@ PicassoProvider.override(({ breakpoints, palette, typography }: Theme) => ({
 
       color: palette.grey.dark,
 
-      '&:not(:last-child)': {
-        marginRight: '2em'
-      },
+      marginRight: '2em',
 
       [breakpoints.up('md')]: {
         minWidth: 'auto',
--- a/packages/picasso/src/Tabs/Tabs.tsx
+++ b/packages/picasso/src/Tabs/Tabs.tsx
@@ -24,23 +24,28 @@ export interface Props
 
   /** The value of the currently selected Tab. If you don't want any selected Tab, you can set this property to false. */
   value: TabsProps['value']
+
+  /** The tabs orientation (layout flow direction)
+   *
+   * @default horizontal */
+  orientation?: TabsProps['orientation']
 }
 
 export interface StaticProps {
   Tab: typeof Tab
 }
 
-const useStyles = makeStyles<Theme>(styles, {
+const useStyles = makeStyles<Theme, Props>(styles, {
   name: 'Tabs'
 })
 
 // eslint-disable-next-line react/display-name
-export const Tabs = forwardRef<HTMLButtonElement, Props>(function Tabs(
+export const Tabs = forwardRef<HTMLButtonElement, Props>(function Tabs (
   props,
   ref
 ) {
-  const { children, onChange, value, ...rest } = props
-  const classes = useStyles()
+  const { children, onChange, value, orientation, ...rest } = props
+  const classes = useStyles(props)
   const action = useTabAction()
 
   return (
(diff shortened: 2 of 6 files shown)


packages/picasso/src/Tabs/styles.ts
```diff
@@ -1,19 +1,24 @@
...
     root: {
       position: 'relative',
       minHeight: 0,
-
       '&::after': {
         position: 'absolute',
         content: '""',
         bottom: 0,
-        left: 0,
+        left: (props: Props) =>
```
havenchyk, line 15: we use different way of doing the same: create a class with desired css and conditionally include it in tsx file please

packages/picasso/src/Tabs/Tabs.tsx
```diff
@@ -24,23 +24,28 @@ export interface Props
...
 }
 
-const useStyles = makeStyles<Theme>(styles, {
+const useStyles = makeStyles<Theme, Props>(styles, {
   name: 'Tabs'
 })
 
 // eslint-disable-next-line react/display-name
-export const Tabs = forwardRef<HTMLButtonElement, Props>(function Tabs(
+export const Tabs = forwardRef<HTMLButtonElement, Props>(function Tabs (
```
havenchyk, line 43: please do not forget to update to run `yarn install` on top of master

packages/picasso/src/Tab/styles.ts
```diff
@@ -19,9 +19,7 @@ PicassoProvider.override(({ breakpoints, palette, typography }: Theme) => ({
 
       color: palette.grey.dark,
 
-      '&:not(:last-child)': {
-        marginRight: '2em'
-      },
+      marginRight: '2em',
```
havenchyk, line 22: does it mean that last child will have marginRight as well?

pkondzior: Yes, but I've didn't noticed that this change makes them look bad anyhow in horizontal representation (which is default here). Also not having that margin in vertical orientation is going to make last element always shifted slightly. So to make this works and looks good in both cases I've made the margin appear always. I don't think we can have here a conditional class approach because tab don't know if they are layout i horizontal or vertical orientation (that property exists only in the Tabs component and AFAIK is not being passed to Tab component)

pkondzior: This is what I'm trying to solve here
Screen Shot 2021-09-06 at 12 58 12 PM

pkondzior: This is how the horizontal will look by default (no changes in margin CSS):
Screen Shot 2021-09-06 at 1 47 43 PM

This is how it will look with my change
Screen Shot 2021-09-06 at 1 48 57 PM

So I don't know why we have this margin there at all, but as first screenshot shows it will affect vertical layout of tabs. Making it to appear on the final element will fix my problem but it wont affect horizontal layout:

Screen Shot 2021-09-06 at 1 51 59 PM

packages/picasso/src/Tabs/Tabs.tsx
```diff
@@ -24,13 +24,18 @@ export interface Props
 
   /** The value of the currently selected Tab. If you don't want any selected Tab, you can set this property to false. */
   value: TabsProps['value']
+
+  /** The tabs orientation (layout flow direction)
+   *
+   * @default horizontal */
+  orientation?: TabsProps['orientation']
```
TomasSlama, line 31: we agreed on TreeView and Stepper to use `direction` instead of orientation, so we should change it here too, so we are consistent.

packages/picasso/src/Tab/styles.ts
TomasSlama: It might be a signal to move this margin to parent, so Tabs will decide what margin to use on children. What do you think guys?

TomasSlama (changes requested):
We should also create a visual test in happo using cypress. Check `/cypress/component` for inspiration. 

I have noticed that you reached out to base team with questions about design, we should wait for the design before merging.

Otherwise LGTM

PR #2166, opened 2021-09-15
style(ButtonGroup): change color of active button

TomasSlama:
[FX-2055]
[Abstract](https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/eb7fd569-ad53-4037-b113-547d576d5795/commits/7a4ef40e212ba549265a23a2c39ad160fe951447/files/f58fa9b1-6361-4d06-a6c7-d57f8c9d715d/layers/E1D99393-578A-4FA8-B902-636204846884?collectionId=1f889287-ba7a-44c1-acb8-e6d02acfb133&collectionLayerId=32357314-53af-4e9b-bc58-a80f8266bfd7&mode=build&selected=3456208762-2114F9B4-4852-43F2-8795-46546D2F208D)

### Description

Uladzimir discussed with the base team, that Pagination and ButtonGroup are special kinds of Button. So for both, we are using a proxy component of Button.

- added new style for group buttons
- unified styles of PaginationButton and GroupButtons (communicated with the BASE team)
  - new style can't be a new variant of Button, because usage of these styles is specific only for pagination and group of buttons.
  - we can't force every button of ButtonGroup to have these styles, because ButtonSplit is also based on ButtonGroup and have different styling
  - created PaginationButton and ButtonGroupItem are proxy components of secondary button enhanced with styling for the active state of the button.

### How to test

- compare storybook with design

### Screenshots

| image | image |

### Review

<details>
<br />

List of available commands:

</details>

[FX-2055]: https://toptal-core.atlassian.net/browse/FX-2055


teimurjan (approved):
LGTM. The only thing I'd ask is to ask the BASE team is about the secondary button. Is it OK that it's changed too?

TomasSlama:
@teimurjan you were right, the secondary button can't be changed

packages/picasso/src/Button/Button.tsx
```diff
@@ -37,6 +37,7 @@ export type VariantType =
   | 'positive'
   | 'secondary'
   | 'transparent'
+  | 'group'
```
teimurjan, line 40: I'd not create a new type inside `Button`. `group` type should only be used inside the group, right? In this case, I'd use the existing `createVariant` function inside the `ButtonGroup`:  https://github.com/toptal/picasso/blob/695c08de591201d5ada8782a142776b728b9fcc1/packages/picasso/src/Button/styles.ts#L6

We use it for `PaginationButton`: https://github.com/toptal/picasso/blob/695c08de591201d5ada8782a142776b728b9fcc1/packages/picasso/src/PaginationButton/styles.ts#L11

TomasSlama: We agreed with Teimur and Milos to unify styles of PaginationButton and Button inside group. We are using new button variant `group`

havenchyk (approved):

havenchyk:
The only thing I don't like is having `Button variant='group'` and `Button.Group` @TomasSlama @teimurjan 
it looks confusing for me and I doubt it should be one of the publicly available variants of the component

teimurjan (dismissed):

packages/picasso/src/ButtonSplit/__snapshots__/test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`ButtonSplit closes menu on click to the item 1`] = `
 <svg
-  class="PicassoSvgArrowDownMinor24-root-95"
+  class="PicassoSvgArrowDownMinor24-root-92"
```
OleksandrNechai, line 5: What does this change mean?

packages/picasso/src/Button/styles.ts
```diff
@@ -3,6 +3,27 @@ import { alpha, outline, mix } from '@toptal/picasso-shared'
 
 const ICON_SPACING = '0.5em'
 
+export const createOutlineCommons = ({ palette }: Theme) => ({
```
OleksandrNechai, line 6: What would we loose if we call this function `outline`?

packages/picasso/src/Button/styles.ts
```diff
@@ -3,6 +3,27 @@ import { alpha, outline, mix } from '@toptal/picasso-shared'
...
+  },
+
+  '&$disabled': {
+    color: palette.grey.main,
+    borderColor: palette.grey.main,
+    backgroundColor: palette.common.white
+  }
+})
+
+export const activeGroupStyles = ({ palette }: Theme) => ({
```
OleksandrNechai, line 21: Why is this function defined here? What do we lose if we call it `activeGroup`?

packages/picasso/src/ButtonGroup/story/ButtonGroup.example.tsx
```diff
@@ -7,34 +7,18 @@ const Example = () => {
   return (
     <div>
       <Button.Group>
-        <Button
-          variant='secondary'
-          active={active === 0}
-          onClick={() => setActive(0)}
-        >
+        <Button.Group.Item active={active === 0} onClick={() => setActive(0)}>
```
OleksandrNechai, line 10: I remember we allways wanted to avoid this 2 dots API. Can we put Item to Button instead of Button.Group? So we have
```
<Button.Group>
   <Button.Item active={active === 0} onClick={() => setActive(0)}>First</Button.Item>
</Button.Group>
```

packages/picasso/src/ButtonGroup/styles.ts
```diff
@@ -1,14 +1,6 @@
...
-const baseButtonProps = {
-  transitionProperty: 'color, background',
-
-  '&:active, &$active, &:hover, &$hovered, &:focus, &$focused': {
-    // border overlap to keep proper border width, but on state change
-    // we need to move up overlapped border
-    zIndex: 1
-  }
-}
+const baseButtonProps = {}
```
OleksandrNechai, line 3: what is this for?

packages/picasso/src/ButtonGroupItem/ButtonGroupItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export interface Props extends ButtonProps {}
+
+const PaginationButton = (props: Props) => {
+  const classes = useStyles()
+
+  return (
+    <Button
+      {...props}
+      className={cx(props.className, classes.root, '__group', {
```
OleksandrNechai, line 20: not sure what is this `__group` is?

packages/picasso/src/ButtonGroupItem/styles.ts
```diff
@@ -0,0 +1,23 @@
+import { createStyles, Theme } from '@material-ui/core/styles'
+
+import { createOutlineCommons, activeGroupStyles } from '../Button/styles'
+
+export default (theme: Theme) =>
+  createStyles({
+    root: {
+      '&.__group': {
```
OleksandrNechai, line 8: I am not sure why this needs to be __group? Why can't we use just normal class?

OleksandrNechai (commented):
Looks like you sweated it a lot :sweat_smile:. I think we are not yet there.
I believe we need a better PR description as I could not understand what is proxy of Button. 
I saw few weird things. Have a look at comments :point_up: :smile:

packages/picasso/src/Button/styles.ts
TomasSlama: I think `outline` is not self-explanatory.

TomasSlama: and we are using outline from picasso-shared

packages/picasso/src/Button/styles.ts
TomasSlama: We need to use this function in Pagination Button and Group Button, so Button was the right place for me, or do you suggest a better place to store these sharable functions? 

I can change the name to `activeGroup`

packages/picasso/src/ButtonGroupItem/ButtonGroupItem.tsx
TomasSlama: I will change it to `classes.group`, we need one more class so we have more specific selector to override secondary button.

packages/picasso/src/ButtonGroup/styles.ts
TomasSlama: we don't need it anymore.

packages/picasso/src/ButtonSplit/__snapshots__/test.tsx.snap
TomasSlama: ButtonSplit is based on ButtonGroup, I don't know in-depth how this numbering is working, but I would not give it much value.

packages/picasso/src/ButtonGroup/story/ButtonGroup.example.tsx
TomasSlama: With Uladzimir we agreed on `Button.Group.Item`. Since you can use this item only in ButtonGroup, it would not be clear where to use Button.Item. 

The ideal would be, if we could name it `Button.Group`, but we use it already for container. And we didn't come up with another good name.

packages/picasso/src/ButtonSplit/__snapshots__/test.tsx.snap
OleksandrNechai: Well, me neither. But it would be nice if at some point you figure that out. It is important to understand how this technology works.

packages/picasso/src/Button/styles.ts
OleksandrNechai: well, what I am trying to say is that to me the word `create` does not add any value, because every function creates the result. Next, when we say `commons` it again does not tell me much. If the function is used in 2+ places, it already contains something common to these 2+ places. So in some sense, every function can be named this way: i.e. Math.Max => Math.createMaxValueUsingCommonAlgorithm

packages/picasso/src/Button/styles.ts
OleksandrNechai: This problem actually makes me think that we made a mistake by sharing a piece of style. This is a component-based UI, so we need to share components rather than pieces of styles. You see, we don't even have a place where to place common styles, because there is no such a thing in component based UI

packages/picasso/src/ButtonGroup/story/ButtonGroup.example.tsx
OleksandrNechai: I will leave it for you to decide, but think carefully if this will not open a can of worms? This way we might end up with 3 and more dots :dizzy_face:

OleksandrNechai (approved):

packages/picasso/src/ButtonGroupItem/styles.ts
```diff
@@ -0,0 +1,17 @@
...
+    root: {
+      '&$group': {
+        ...createOutlineCommons(theme),
+        '&:active, &$active': activeGroup(theme)
+      }
+    },
+    active: {},
+    hover: {},
+    focused: {},
+    group: {}
```
teimurjan, line 16: Why do you need `group` class if it's always applied to the component? Can you just add the styles to `root`?


```suggestion
  createStyles({
    root: {
      ...createOutlineCommons(theme),
      '&:active, &$active': activeGroup(theme)
    },
    active: {},
    hover: {},
    focused: {}
```

teimurjan (approved):

packages/picasso/src/ButtonGroupItem/styles.ts
TomasSlama: that was my first choice too, but it didn't work, the selector wasn't specific enough and was overwritten by the default behavior of the secondary button.

PR #2174, opened 2021-09-23
feat(Dropzone): create new component

TomasSlama:
[FX-1870]
[Abstract](https://app.abstract.com/share/2556a122-6443-4546-96db-3d40c37b2a0d?mode=design&sha=58ee1df911f7d6ca2d9214171088a69bdff23401) 
[Storybook](https://picasso.toptal.net/fx-1870-dropzone/?path=/story/forms-dropzone--dropzone)

### Description

- added new component Dropzone based on [react-dropzone](https://react-dropzone.js.org/)
- functionality is left for users to develop, I was trying to keep usage similar with FileInput, so users are not confused

### How to test

- check [Storybook](https://picasso.toptal.net/fx-1870-dropzone/?path=/story/forms-dropzone--dropzone) and compare with [Abstract](https://app.abstract.com/share/2556a122-6443-4546-96db-3d40c37b2a0d?mode=design&sha=58ee1df911f7d6ca2d9214171088a69bdff23401)

### Screenshots

image

### Review

<details>
<br />

List of available commands:

</details>

[FX-1870]: https://toptal-core.atlassian.net/browse/FX-1870


packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const useFiles = () => {
+  const [files, setFiles] = useState<FileUpload[]>([])
+  const [errorMessages, setError] = useState<string[]>([])
+
+  const addFiles: DropzoneOptions['onDrop'] = (
+    acceptedFiles,
+    rejectedFiles
+  ) => {
+    if (acceptedFiles.length) {
```
OleksandrNechai, line 26:
```suggestion
    if (acceptedFiles.length > 0) {
```

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,95 @@
...
+      const previousFiles = files
+      const newFiles = Array.from(acceptedFiles).map(file => ({
+        file,
+        uploading: true
+      }))
+
+      setFiles([...previousFiles, ...newFiles])
```
OleksandrNechai, line 33: Hmm, why is `previousFiles` introduced? What would happen if we did `[...files, ...newFiles]`

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,95 @@
...
+          clearInterval(interval)
+        } else {
+          setFiles([
+            ...previousFiles,
+            ...newFiles.map(file => ({ ...file, progress }))
+          ])
+          progress += 10
+        }
+      }, 250)
+    }
```
OleksandrNechai, line 50: I don't get what is this code doing. Why is this gradual increase of progress needed? To me, it looks like a simple setTimeout after 2500ms.

packages/picasso-lab/src/Dropzone/story/Hover.example.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  return {
+    files,
+    addFiles,
+    removeFile
+  }
+}
+
+const Example = () => {
```
OleksandrNechai, line 26: I think this example should be the only one that demonstrates how things work. All other examples just demonstrate different possible states of the `dropzone` and probably should be static ones with static data and no interactivity.

packages/picasso-lab/src/Dropzone/story/index.jsx
```diff
@@ -0,0 +1,178 @@
...
+    'https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/master/commits/58ee1df911f7d6ca2d9214171088a69bdff23401/files/13531207-e094-44ec-ae1f-f27628c1aea5/layers/27C9F1FA-E469-4AA4-936C-DD752F77ADE3?mode=build&selected=root-8BD3C500-A858-4CC4-8CE8-02E0453C0C4D&sha=58ee1df911f7d6ca2d9214171088a69bdff23401'
+  )}
+  `
+)
+
+page.createTabChapter('Props').addComponentDocs({
+  component: Dropzone,
+  name: 'Dropzone',
+  additionalDocs: {
+    accept: {
```
OleksandrNechai, line 17: I think exposing react-dropzone libraries props directly is a mistake. We need to sit down and think about which props do we really need and what their names should be. End-user should not know which library we use.

packages/picasso-lab/src/Dropzone/styles.ts
```diff
@@ -0,0 +1,39 @@
+import { rem } from '@toptal/picasso-shared'
+import { createStyles, Theme } from '@material-ui/core/styles'
+
+export default ({ palette, sizes }: Theme) =>
+  createStyles({
+    root: {
+      padding: '20px',
+      borderWidth: 1,
```
OleksandrNechai, line 8: `sizes.borderWidth`

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,107 @@
...
+    ...dropzoneOptions
+  } = props
+
+  const { getRootProps, isDragActive, getInputProps } = useDropzone(
+    dropzoneOptions
+  )
+
+  const isCompleted =
+    Array.isArray(value) &&
+    value.length &&
```
OleksandrNechai, line 49:
```suggestion
    value.length > 0 &&
```

packages/picasso-lab/src/Dropzone/styles.ts
```diff
@@ -0,0 +1,39 @@
+import { rem } from '@toptal/picasso-shared'
+import { createStyles, Theme } from '@material-ui/core/styles'
+
+export default ({ palette, sizes }: Theme) =>
+  createStyles({
+    root: {
+      padding: '20px',
```
OleksandrNechai, line 7: why can't we use standard Container padding?

packages/picasso-lab/src/Dropzone/styles.ts
```diff
@@ -0,0 +1,39 @@
...
+      borderWidth: 1,
+      borderRadius: sizes.borderRadius.medium,
+      borderColor: palette.grey.light2,
+      borderStyle: 'dashed',
+      backgroundColor: palette.common.white,
+      color: palette.grey.dark,
+      outline: 'none',
+      transition: 'border .24s ease-in-out',
```
OleksandrNechai, line 15: How did designers specify this to you? :smiley:

packages/picasso-lab/src/Dropzone/test.tsx
```diff
@@ -0,0 +1,63 @@
...
+        }
+      ],
+      'data-testid': 'foobar'
+    })
+
+    expect(
+      Array.from(getByTestId('foobar').classList).map(
+        classname => classname.split('-')[1]
+      )
+    ).toContain('completed')
```
OleksandrNechai, line 43: Can we check it without resorting to checking class? It looks like an implementation check. If we are after visual appearance, then maybe worth checking with Happo only?

packages/picasso-lab/src/Dropzone/test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const { getByTestId } = renderDropzone({
+      'data-testid': 'foobar'
+    })
+
+    expect(
+      Array.from(getByTestId('foobar').classList).map(
+        classname => classname.split('-')[1]
+      )
+    ).not.toContain('completed')
+  })
```
OleksandrNechai, line 56: Can we check it without resorting to checking class? It looks like an implementation check. If we are after visual appearance, then maybe worth checking with Happo only?

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,107 @@
...
+    value.filter(filterByUploading).length === 0
```
OleksandrNechai, line 50:
```suggestion
    value.some(file=>file.uploading)
```

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const classes = useStyles()
+
+  return (
+    <Container style={style}>
+      <Container
+        flex
+        direction='column'
+        alignItems='center'
+        data-testid={dataTestId}
+        {...getRootProps({
```
OleksandrNechai, line 61: now you see why I asked you to take that course? :)

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,107 @@
...
+          weight='semibold'
+        >
+          Click or drag file to upload
+        </Typography>
+        {hint && errorMessages.length === 0 && (
+          <FormHint className={cx(classes.hint)}>{hint}</FormHint>
+        )}
+        {errorMessages.length > 0 &&
+          errorMessages.map((error, index) => (
+            <FormError key={`${error}-${String(index)}`}>{error}</FormError>
```
OleksandrNechai, line 83: `String` probably is redundant

cypress/component/Dropzone.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+      </TestingPicasso>
+    )
+    cy.contains('Max file size').should('be.visible')
+    cy.get('body').happoScreenshot()
+  })
+
+  it('renders with hovered', () => {
+    mount(
+      <TestingPicasso>
+        {renderDropzone({ className: '__hover', hint: 'Max file size: 25MB' })}
```
OleksandrNechai, line 34: is `__hover` something standard?

cypress/component/Dropzone.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+    mount(
+      <TestingPicasso>
+        <DropzoneWithState />
+      </TestingPicasso>
+    )
+    cy.get('body').happoScreenshot()
+
+    cy.get('[class*=completed]').should('exist')
+    cy.get('button').as('removeButton').realClick()
+    cy.get('[class*=completed]').should('not.exist')
```
OleksandrNechai, line 107: I am not sure if querying classes is reliable. I have a belief that we should only check how things look and how they behave. And try to avoid looking inside. As soon as you look inside, the test becomes dependent on implementation and hence fragile.

OleksandrNechai (commented):
WOW :+1: 

But not there yet :point_down:

cypress/component/Dropzone.spec.tsx
TomasSlama: not really, I didn't want to create a new prop `hover` only for visual testing purposes, so I created `__hover` class which simulates the hover effect. What do you think about it?

cypress/component/Dropzone.spec.tsx
TomasSlama: you are probably right, I wanted to be sure, that states are changing when value changes. But it's probably overkill, I will remove this part.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
TomasSlama: we have been discussing this topic with others. This is how MUI is approaching keys in arrays when you can't find another unique key.

`key={`${error}-${index}`}` would generate an error, since we can not use the index in keys. But this hack allows us to use the index in some cases.

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
TomasSlama: it simulates upload, I took this example from FileInput. If you try to add some files in the default example in a storybook, it will show you this nice progress bar.

packages/picasso/src/Icon/svg/upload16.svg
```diff
@@ -0,0 +1,20 @@
...
+            </mask>
+            <use id="Combined-Shape" fill="#000000" fill-rule="nonzero" xlink:href="#path-1"></use>
+            <g id="Group" mask="url(#mask-2)" fill="#455065">
+                <g id="Colors-/-307-Gray-Dark">
+                    <rect id="Rectangle" x="0" y="0" width="16" height="16"></rect>
+                </g>
+            </g>
+        </g>
+    </g>
+</svg>
```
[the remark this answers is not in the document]
TomasSlama: will do.

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
TomasSlama: we need `previousFiles` to keep reference of previous files in setInterval

cypress/component/Dropzone.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+const renderDropzone = (props: DropzoneProps) => (
+  <Container padded='small'>
+    <Dropzone style={{ width: 300 }} {...props} />
+  </Container>
+)
+
+describe('Dropzone', () => {
+  it('renders without props', () => {
+    mount(<TestingPicasso>{renderDropzone({})}</TestingPicasso>)
```
denieler, line 15:
```suggestion
    mount(<TestingPicasso>{renderDropzone()}</TestingPicasso>)
```

cypress/component/Dropzone.spec.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useState } from 'react'
+import { Container } from '@toptal/picasso'
+import { Dropzone, DropzoneProps } from '@toptal/picasso-lab'
+import { TestingPicasso } from '@toptal/picasso/test-utils'
+import { mount } from '@cypress/react'
+
+const renderDropzone = (props: DropzoneProps) => (
+  <Container padded='small'>
+    <Dropzone style={{ width: 300 }} {...props} />
```
denieler, line 9:
```suggestion
    <Dropzone style={{ width: '300px' }} {...props} />
```
?

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,107 @@
...
+  hint?: string
+  /** Callback invoked when a file item is removed */
+  onRemove?: (fileName: string, index: number) => void
+  /** Value uses the File interface. */
+  value?: FileUpload[]
+  /** Reasons why files couldn't be droped into dropzone */
+  errorMessages?: string[]
+}
+
+const useStyles = makeStyles<Theme>(styles, { name: 'FileInputContent' })
```
denieler, line 24:
```suggestion
const useStyles = makeStyles<Theme>(styles, { name: 'Dropzone' })
```

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const isCompleted =
+    Array.isArray(value) &&
+    value.length &&
+    value.filter(filterByUploading).length === 0
+
+  const classes = useStyles()
+
+  return (
+    <Container style={style} ref={ref}>
```
denieler, line 51: why we don't have `className` applied at the same level as `style` prop?

packages/picasso-lab/src/Dropzone/Dropzone.tsx
denieler: eslint even has a rule https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md

packages/picasso-lab/src/Dropzone/story/index.jsx
```diff
@@ -0,0 +1,178 @@
+import { Dropzone } from '../Dropzone'
+import PicassoBook from '~/.storybook/components/PicassoBook'
+
+const page = PicassoBook.section('Forms').createPage(
```
denieler, line 4: why in `Forms`? it should be probably in lab?

packages/picasso-lab/src/Dropzone/story/index.jsx
denieler: I agree, I would not list them all. Maybe even none from the beginning

packages/picasso-lab/src/Dropzone/styles.ts
```diff
@@ -0,0 +1,40 @@
...
+      '&:hover, &$dragActive, &.__hover': {
+        borderColor: palette.blue.main,
+        cursor: 'pointer'
+      },
+      '&$completed': {
+        backgroundColor: palette.grey.lighter
+      }
+    },
+    text: {
+      lineHeight: rem('22px')
```
denieler, line 26: why do we need to change line height for the Typography? Does it mean that this text is not compatible with BASE in the dropzone component?

packages/picasso-lab/src/Dropzone/styles.ts
```diff
@@ -0,0 +1,40 @@
...
+      borderWidth: 1,
+      borderRadius: sizes.borderRadius.medium,
+      borderColor: palette.grey.light2,
+      borderStyle: 'dashed',
+      backgroundColor: palette.common.white,
+      color: palette.grey.dark,
+      outline: 'none',
+      transition: 'border .24s ease-in-out',
+      gap: '0.5rem',
+      '&:hover, &$dragActive, &.__hover': {
```
denieler, line 17: uh, this looks dirty with `__hover` only for visual tests 😃

cypress/component/Dropzone.spec.tsx
TomasSlama: It's the same, but if you think it's better, why not.

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: exactly, there is even a ticket in todo, to deal with this in general. We are using line-height 1.5em everywhere, but Milos was saying that line-height can't be an odd number. So for 14px it's supposed to be 22px not 21px.

packages/picasso-lab/src/Dropzone/story/index.jsx
TomasSlama: I think we need to keep the most important ones since they are needed for basic functionality. At least these which we are using in examples

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: design says 20px, container allowes as 8, 16, 24, ....

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: yeah, it was my suggestion how to solve problem with states in happo Uladzimir shared with us. And we are doing something similar with button where we have props like hover, focused which are used also only in tests, or at least I don't see any other real usage scenario.

packages/picasso-lab/src/Dropzone/test.tsx
TomasSlama: className is only difference, I agree we can check only with happo.

packages/picasso-lab/src/Dropzone/styles.ts
denieler: let's fix it in Typography then instead of adding new class, which we eventually will forget to remove 😃

cypress/component/Dropzone.spec.tsx
denieler: it's just more clear which unit is used

packages/picasso-lab/src/Dropzone/styles.ts
OleksandrNechai: Let's maybe check if this is not a designer's mistake?

packages/picasso-lab/src/Dropzone/styles.ts
OleksandrNechai: Hmm, interesting hack :) I think either we do it the same way as Button does or we don't do it. Other teams might also need it for tests. I don't see the need of introducing a new pattern.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { makeStyles, Theme } from '@material-ui/core/styles'
+import { BaseProps } from '@toptal/picasso-shared'
+import { Upload24 } from '@toptal/picasso/Icon'
+import { FormError, Container, FileList, Typography } from '@toptal/picasso'
+import FormHint from '@toptal/picasso/FormHint'
+
+import { FileUpload } from './types'
+import styles from './styles'
+
+export interface Props extends BaseProps, Omit<DropzoneOptions, 'children'> {
```
OleksandrNechai, line 13: I think we should not expose `DropzoneOptions`.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
OleksandrNechai: I mean if you concat string and number you get string, no need to apply String :smile:

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,103 @@
...
+    </Container>
+  )
+})
+
+Dropzone.displayName = 'PicassoDropzone'
+
+Dropzone.defaultProps = {
+  maxSize: Infinity,
+  minSize: 0,
+  multiple: true
```
OleksandrNechai, line 100: It is weird not seeing these props in `Props`

packages/picasso-lab/src/Dropzone/story/Error.example.tsx
```diff
@@ -0,0 +1,62 @@
...
+    <Dropzone
+      value={files}
+      onDrop={addFiles}
+      onRemove={removeFile}
+      hint='Max file size: 25MB'
+      accept='image/*'
+      maxFiles={1}
+      maxSize={1024}
+      errorMessages={errorMessages}
+    />
```
OleksandrNechai, line 58: These properties need to be in the `Props` of Dropzone explicitly defined.

packages/picasso/src/ModalTitle/ModalTitle.tsx
```diff
@@ -13,26 +13,27 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
...
+      ref={ref}
+      className={cx(classes.root, className)}
+      style={style}
+    >
+      <Typography variant='heading' size='large'>
+        {children}
+      </Typography>
+    </div>
+  )
+})
```
OleksandrNechai, line 36: reformatting?

OleksandrNechai (approved):

OleksandrNechai (changes requested):
Looks like Dropzone's interface is not defined properly yet

cypress/component/Dropzone.spec.tsx
teimurjan: Usually, we do not rely on the classes. It isn't good to have a prop just for VT, but we do the same for `Button`. That's why I'd do it with the prop just for consistency.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { forwardRef } from 'react'
+import { useDropzone, DropzoneOptions } from 'react-dropzone'
+import cx from 'classnames'
+import { makeStyles, Theme } from '@material-ui/core/styles'
+import { BaseProps } from '@toptal/picasso-shared'
+import { Upload24 } from '@toptal/picasso/Icon'
+import { FormError, Container, FileList, Typography } from '@toptal/picasso'
+import FormHint from '@toptal/picasso/FormHint'
```
teimurjan, line 8: I'd add an export of `FormHint` to the index file and import from `@toptal/picasso`.

teimurjan (approved):

teimurjan:
Is it OK that `Dropzone` in the uploading state allows uploading another file? Gray background makes me think that it should be disabled.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
TomasSlama: I will expose only some specific props, that are most important.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
TomasSlama: that is true, but without String will eslint throw error

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: I don't think its a mistake, they use quite commonly lot of different paddings

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: they didn't I will check with other components how hover transition is set

packages/picasso/src/ModalTitle/ModalTitle.tsx
TomasSlama: yep, it was done by `yarn lint`

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: I will ask Milos when I will be showing him

cypress/component/Dropzone.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+          ]
+        })}
+      </TestingPicasso>
+    )
+    cy.get('body').happoScreenshot()
+  })
+
+  it('renders completed', () => {
+    const DropzoneWithState = () => {
+      const [value, setValue] = useState<DropzoneProps['value']>([
```
OleksandrNechai, line 86: Why do we need useState? It looks like there is no interaction with Dropzone in this test.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,119 @@
...
+} from '@toptal/picasso'
+
+import { FileUpload } from './types'
+import styles from './styles'
+
+export interface Props extends BaseProps {
+  /**
+   * Set accepted file types. See https://github.com/okonet/attr-accept for more information.
+   */
+  accept?: DropzoneOptions['accept']
```
OleksandrNechai, line 22: Although simpler, I think we should avoid using DropzoneOptions in type definitions. If we do that, we make our clients depend on `DropzoneOptions` type definition. If we update our deps, our code might be fine but the client's code may fail to compile and there is no way we can discover that.

On the other hand, if we define the interface on our own, we will get type mismatches after the update while compiling Picasso.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,119 @@
...
+  /** Allow drag 'n' drop (or selection from the file dialog) of multiple files */
+  multiple?: boolean
+  /** The text of the hint */
+  hint?: string
+  /** Callback invoked when a file item is removed */
+  onRemove?: (fileName: string, index: number) => void
+  onDropAccepted?: DropzoneOptions['onDropAccepted']
+  onDropRejected?: DropzoneOptions['onDropRejected']
+  onDrop?: DropzoneOptions['onDrop']
+  validator?: DropzoneOptions['validator']
```
OleksandrNechai, line 38: Do we really need it atm? If not I would avoid adding props just in case. We don't yet know what our users will need. When we discover the needs, we will add more props.

Please, make sure each prop is documented.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,119 @@
...
+    hint,
+    onRemove,
+    value,
+    className,
+    errorMessages = [],
+    style,
+    'data-testid': dataTestId,
+    focused,
+    hovered,
+    ...dropzoneOptions
```
OleksandrNechai, line 63: let's list the options explicitly. It would be easier to read the code.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,119 @@
...
+        flex
+        direction='column'
+        alignItems='center'
+        data-testid={dataTestId}
+        {...getRootProps({
+          className: cx(classes.root, {
+            [classes.dragActive]: isDragActive,
+            [classes.hovered]: hovered,
+            [classes.disabled]: dropzoneOptions.disabled,
+            [classes.focused]: focused
```
OleksandrNechai, line 84: Can't it be that the Dropzone is hovered and is focused at the same time? In the code above, it looks like states are mutually exclusive.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
```diff
@@ -0,0 +1,119 @@
...
+      )}
+    </Container>
+  )
+})
+
+Dropzone.displayName = 'PicassoDropzone'
+
+Dropzone.defaultProps = {
+  disabled: false,
+  maxSize: Infinity,
```
OleksandrNechai, line 114: IDK, maybe we can set some reasonable limit here?

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,105 @@
...
+  })
+
+  return (
+    <Dropzone
+      value={files}
+      onDrop={addFiles}
+      onRemove={removeFile}
+      hint={`Files allowed 2. Max file size: ${MAX_SIZE / 1000}KB`}
+      accept='image/*'
+      validator={customSizeValidator}
```
OleksandrNechai, line 98: Why such validator is needed if we have maxSize prop?

packages/picasso/src/index.ts
```diff
@@ -31,6 +31,8 @@ export { default as FileList } from './FileList'
 export type { FileListProps } from './FileList'
 export { default as Form } from './Form'
 export type { FormProps } from './Form'
+export { default as FormHint } from './FormHint'
+export type { FormHintProps } from './FormHint'
```
OleksandrNechai, line 35: Why and how is this related to Dropzone?

OleksandrNechai (approved):
I still have a bunch of questions. I hope you will figure them out. So I am approving to avoid blocking you.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
TomasSlama: I'm not sure what you mean. Dropzone can be focused and hovered at the same time, these props are just to showcase these states in the story and cypress. This same approach we use in Button.

packages/picasso-lab/src/Dropzone/Dropzone.tsx
TomasSlama: I don't think it is up to us, users should define these restrictions.

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
TomasSlama: For custom error message

packages/picasso/src/index.ts
TomasSlama: This was a suggestion of @teimurjan. Since we use FormHint inside Dropzone, we can import FormHint as any other Component `import { FormHint } from '@toptal/picasso'` and not `import FormHint from '@toptal/picasso/FormHint`

cypress/component/Dropzone.spec.tsx
TomasSlama: fixed, it was leftover

packages/picasso-lab/src/Dropzone/Dropzone.tsx
TomasSlama: I think we need it, default validation error messages aren't the same as in design and I don't think we can predict the use cases of other developers. I have added documentation into the storybook, so I will add them here also.

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,105 @@
+import React, { useEffect, useState } from 'react'
+import { Dropzone, DropzoneProps } from '@toptal/picasso-lab'
+import { FileUpload } from '@toptal/picasso-lab/Dropzone'
+
+const MAX_SIZE = 600 * 1000
+const customSizeValidator: DropzoneProps['validator'] = file => {
```
denieler, line 6: I would not enforce people to use proxying of types, like `DropzoneProps['validator']`. For projects who use such props if any breaking changes happen would be pretty difficult to upgrade Picasso and find problems. I would create types in examples and use them

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,105 @@
...
+  if (file.size > MAX_SIZE) {
+    return {
+      code: 'size-too-large',
```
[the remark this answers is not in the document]
TomasSlama: good point, I will use the enum.

TomasSlama:
@toptal-bot run build

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,121 @@
...
+) => DropzoneFileError | DropzoneFileError[] | null
+type OnDropType = (
+  acceptedFiles: File[],
+  fileRejections: DropzoneFileRejection[],
+  event: DropzoneDropEvent
+) => void
+
+const MAX_SIZE = 600 * 1000
+
+const customSizeValidator: ValidatorType = file => {
```
denieler, line 22: maybe make this to simplify? so we would always return array of `DropzoneFileError` even if single error has happened
```suggestion
const customSizeValidator: DropzoneFileError[] | null = (file: File) => {
```

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  useEffect(() => {
+    if (files.length >= maxFiles) {
+      setDisabled(true)
+    } else {
+      setDisabled(false)
+    }
+  }, [maxFiles, files.length])
+
+  const addFiles: OnDropType = (acceptedFiles, rejectedFiles) => {
```
denieler, line 46: I would also inline types here
```suggestion
  const addFiles = (
    acceptedFiles: File[],
    fileRejections: DropzoneFileRejection[],
    event: DropzoneDropEvent
  ) => {
```

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,121 @@
...
+    if (files.length + acceptedFiles.length + rejectedFiles.length > maxFiles) {
+      return setError(['Too many files'])
+    }
+    if (acceptedFiles.length > 0) {
+      const previousFiles = files
+      const newFiles = Array.from(acceptedFiles).map(file => ({
```
denieler, line 52: Looks like it is already an array `File[]`
```suggestion
      const newFiles = acceptedFiles.map(file => ({
```

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
```diff
@@ -0,0 +1,121 @@
...
+          ])
+          clearInterval(interval)
+        } else {
+          setFiles([
+            ...previousFiles,
+            ...newFiles.map(file => ({ ...file, progress }))
+          ])
+          progress += 10
+        }
+      }, 250)
```
denieler, line 73: this doesn't look like a real example of file download 😃 can we get the real progress of the file upload?

packages/picasso-lab/src/Dropzone/story/Default.example.tsx
denieler: can we use it with some default error message from Dropzone?
If so - I would show Default example with super minimum configuration and add another example with the custom error message capabilities

packages/picasso-lab/src/Dropzone/styles.ts
denieler: @TomasSlama what was the conclusion from Milos? 😃

packages/picasso-lab/src/Dropzone/styles.ts
```diff
@@ -0,0 +1,44 @@
...
+        '&:hover': {
+          cursor: 'no-drop',
+          borderColor: palette.grey.light2
+        }
+      }
+    },
+    hint: {
+      margin: 0,
+      '& > *': {
+        lineHeight: rem('16px')
```
denieler, line 33: isn't it just `1em`?

packages/picasso-lab/src/Dropzone/test.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { render } from '@toptal/picasso/test-utils'
+import { OmitInternalProps } from '@toptal/picasso-shared'
+
+import Dropzone, { Props } from './Dropzone'
+
+const renderDropzone = (props: OmitInternalProps<Props>) =>
+  render(<Dropzone {...props} />)
+
+describe('Dropzone', () => {
+  it('renders', () => {
```
denieler, line 11: we need also one snapshot test of html generated for Dropzone render

packages/picasso-lab/src/Note/story/index.jsx
```diff
@@ -4,7 +4,7 @@ import noteSubtitleStory from '../../NoteSubtitle/story'
 import noteContentStory from '../../NoteContent/story'
 import { Note } from '../Note'
 
-const page = PicassoBook.section('Components').createPage(
+const page = PicassoBook.section('Picasso Lab').createPage(
```
denieler, line 7: 👍 🙇‍♂️

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: you are right, '1rem'

denieler: but what about `1em`? will it work, or is the font-size not default?

packages/picasso-lab/src/Dropzone/story/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+        'Set accepted file types. See https://github.com/okonet/attr-accept for more information. Keep in mind that mime type determination is not reliable across platforms. CSV files, for example, are reported as text/plain under macOS but as application/vnd.ms-excel under Windows. In some cases there might not be a mime type set at all. See: https://github.com/react-dropzone/react-dropzone/issues/276'
+    },
+    onDropAccepted: {
+      name: 'onDropAccepted',
+      type: {
+        name: 'function',
+        description: '(files: File[], event) => void'
+      },
+      description:
+        'Cb for when the drop event occurs. Note that if no files are accepted, this callback is not invoked.'
```
denieler, line 30:
```suggestion
        'Callback for when the drop event occurs. Note that if no files are accepted, this callback is not invoked.'
```

packages/picasso-lab/src/Dropzone/story/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    onDropRejected: {
+      name: 'onDropRejected',
+      type: {
+        name: 'function',
+        description: '(fileRejections: FileRejection[], event) => void'
+      },
+      description:
+        'Cb for when the drop event occurs. Note that if no files are rejected, this callback is not invoked.'
```
denieler, line 39:
```suggestion
        'Callback for when the drop event occurs. Note that if no files are rejected, this callback is not invoked.'
```

packages/picasso-lab/src/Dropzone/story/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+    },
+    onDrop: {
+      name: 'onDrop',
+      type: {
+        name: 'function',
+        description:
+          '(acceptedFiles: File[], fileRejections: FileRejection[], event:DragEvent | Event) => void'
+      },
+      description: `
+Cb for when the drop event occurs. Note that this callback is invoked after the getFilesFromEvent callback is done.
```
denieler, line 58:
```suggestion
Callback for when the drop event occurs. Note that this callback is invoked after the getFilesFromEvent callback is done.
```

packages/picasso-lab/src/Dropzone/test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const { queryByText, container } = renderDropzone({})
+
+    expect(queryByText('Click or drag file to upload')).toBeVisible()
```
denieler, line 14: if we have snapshot we probably don't need even this expectation 😃

denieler (approved):

packages/picasso-lab/src/Dropzone/styles.ts
TomasSlama: it wouldnt work, because there is 11px font-size

PR #2201, opened 2021-11-01
refactor(Container): make gap to fuction as other size props

TomasSlama:
### Description

Make `gap` property accept the same values as other size properties. For example `'small' | 'medium' | 'large'`.

It is a breaking change, but I believe that most users will not have to change anything because this property was introduced in the last major release.

### How to test

- no tests should be broken

### Review

<details>
<br />

List of available commands:

</details>

### DIFF
--- a/packages/picasso/src/Container/styles.ts
+++ b/packages/picasso/src/Container/styles.ts
@@ -55,6 +55,14 @@ const paddings = spacingVariants.reduce((acc, variant) => {
   return acc
 }, Object.create(null))
 
+const gaps = spacingVariants.reduce((acc, variant) => {
+  acc[`${variant}Gap`] = {
+    gap: spacingToRem(variant as SpacingType)
+  }
+
+  return acc
+}, Object.create(null))
+
 const colorVariant = (colorOptions?: SimplePaletteColorOptions | Color) => {
   if (!colorOptions) {
     return {}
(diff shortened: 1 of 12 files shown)


denieler (approved):

denieler (commented):
if it's a breaking change - I would release it as a breaking change

packages/picasso/src/Container/styles.ts
```diff
@@ -55,6 +55,14 @@ const paddings = spacingVariants.reduce((acc, variant) => {
   return acc
 }, Object.create(null))
 
+const gaps = spacingVariants.reduce((acc, variant) => {
+  acc[`${variant}Gap`] = {
+    gap: spacingToRem(variant as SpacingType)
+  }
+
+  return acc
+}, Object.create(null))
```
dmaklygin1, line 64: wdyt guys if we join this loop with the previous one:
```suggestion
const paddings = spacingVariants.reduce((acc, variant) => {
  acc[`${variant}Padding`] = acc[`${variant}Gap`] = {
    padding: spacingToRem(variant as SpacingType)
  }
  return acc
}, Object.create(null))
```

dmaklygin1 (approved):

packages/picasso/src/Container/styles.ts
TomasSlama: I don't have a strong opinion against it, but I don't like it much, suddenly `paddings` would have to be renamed to `paddingsAndGaps` and we would mix responsibilities.

PR #2194, opened 2021-10-19
feat(AvatarGroup): create new component Avatar.Group

TomasSlama:
[FX-2143]
[Abstract](https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/master/commits/c5c9dad3787a1f500fcbcc38f8ec74489b5a833b/files/6e64349b-3b5b-4af6-87a3-7f7d15b299ca/layers/1ACF9D6E-E04B-4D8E-AF99-0E655E250847?mode=build&selected=root-B0DC6F76-BCA2-4171-B246-A811CFAA25FC)

### Description

Creating a new component Avatar.Group to display multiple Avatars. The user can set limit on how many Avatars can be displayed. By default limit is set to 5, if 6 items provided, last item will be transformed to Avatar placeholder with +2 indicator

### How to test

- Compare design with temploy

### Screenshots

image

### Review

- [x] Read [CONTRIBUTING.md](https://github.com/toptal/picasso/blob/master/CONTRIBUTING.md) and [Component API principles](https://github.com/toptal/picasso/blob/master/docs/api-principles.md)
- [x] Annotate all `props` in component with documentation
- [x] Create `examples` for component
- [x] Ensure that deployed demo has expected results and good examples
- [x] Ensure that tests pass by running `yarn test`
- [x] Ensure that visuals tests pass by running `yarn test:visual`. If not - check the documentation [how to fix visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md#fixing-broken-visual-tests-inside-a-pr)
- [x] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).

<details>
<summary>PR commands</summary>
<br />

List of available commands:

- `@toptal-bot run all` - Run whole pipeline
- `@toptal-bot run build` - Check build
- `@toptal-bot run visual` - Run visual tests
- `@toptal-bot run deploy:documentation` - Deploy documentation
- `@toptal-bot run package:alpha-release` - Release alpha version

</details>

[FX-2143]: https://toptal-core.atlassian.net/browse/FX-2143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2202, opened 2021-11-01
refactor(Avatar): make avatar more reusable

TomasSlama:
### Description

- Refactor Avatar component so it is possible to be used inside AvatarGroup in [another PR](https://github.com/toptal/picasso/pull/2194). 
- Also fixed the size of clipped corner and Toptal logo so it complies with BASE 2.0

### How to test

- compare temploy with [design](https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/master/commits/c5c9dad3787a1f500fcbcc38f8ec74489b5a833b/files/6e64349b-3b5b-4af6-87a3-7f7d15b299ca/layers/1ACF9D6E-E04B-4D8E-AF99-0E655E250847?mode=build&selected=root-B0DC6F76-BCA2-4171-B246-A811CFAA25FC)
- change should have no impact for Avatar users, only internal changes

### Screenshots

| image | image |

### Review

<details>
<br />

List of available commands:

</details>


packages/picasso/src/Avatar/Avatar.tsx
```diff
@@ -1,12 +1,13 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
denieler, line 1: I would use `!` instead, it's very dangerous to have such rule per-file, better per-line

packages/picasso/src/Avatar/IconAvatar/styles.ts
```diff
@@ -0,0 +1,17 @@
...
+      position: 'absolute',
+      top: '50%',
+      left: '50%',
+      transform: 'translate(-50%, -50%)'
+    },
+    xxsmalLIcon: { fontSize: '1rem' },
+    xsmallIcon: { fontSize: '1rem' },
+    smallIcon: { fontSize: rem('24px') },
+    mediumIcon: { fontSize: rem('30px') },
+    largeIcon: { fontSize: rem('48px') }
```
denieler, line 16: can we use `em` for icon sizes?

packages/picasso/src/Avatar/TextAvatar/TextAvatar.tsx
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
denieler, line 1: same here

packages/picasso/src/Avatar/utils/getClipPathCornerMask.ts
```diff
@@ -0,0 +1,14 @@
+import { AvatarSizeType } from './../Avatar'
+
+const CORNER_SIZE = {
+  xxsmall: '0.5rem',
+  xsmall: '0.5rem',
+  small: '1rem',
+  medium: '1.5rem',
+  large: '1.5rem'
```
denieler, line 8: can we use `em` instead?

packages/picasso/src/Avatar/utils/getDimensions.ts
```diff
@@ -0,0 +1,20 @@
+import { VariantType } from '../Avatar'
+
+const DIMENSIONS_BY_VARIANT = {
+  square: {
+    width: '2.5em',
+    height: '2.5em'
+  },
+  portrait: {
+    width: '1.6667em',
```
denieler, line 9: is it supposed to be `25px`? I would use utility function in this case, so it will make more precise calculations and it will be more clear for other engineers

packages/picasso/src/Avatar/utils/getDimensions.ts
```diff
@@ -0,0 +1,20 @@
...
+    height: '2.5em'
+  },
+  landscape: {
+    width: '2.5em',
+    height: '1.6667em'
```
denieler, line 14: same here

denieler (commented):
looks like `__image_snapshots__/notifications-stream-options-effect-2-snap.png` shouldn't be changed

packages/picasso/src/Avatar/IconAvatar/styles.ts
TomasSlama: I think it would not be clear, because we already change the font size on the parent div, so on the first view it would be random numbers.

packages/picasso/src/Avatar/utils/getClipPathCornerMask.ts
TomasSlama: for same reason as above, I think we should keep rem

packages/picasso/src/Avatar/utils/getDimensions.ts
TomasSlama: it is not supposed to be 25. When we change the size property, it changes font size on wrapper div, so for every size, it is a different value, but with the same proportions.

packages/picasso/src/Avatar/Avatar.tsx
TomasSlama: I still don't follow what you mean by `!`. This warning shows because I'm using `!`. I will add this rule to every line.

packages/picasso/src/Avatar/Avatar.tsx
```diff
@@ -18,78 +18,68 @@ export interface Props extends StandardProps, HTMLAttributes<HTMLDivElement> {
...
-          classes={classes}
+        <TextAvatar
           className={className}
-          name={name}
-          size={size}
-          variant={variant}
-        />
+          fontSize={isLongText ? 'small' : 'large'}
+          /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+          size={size!}
```
dmaklygin1, line 65: @denieler, could you please explain what's the benefit of usage `!` instead of setting a default value for this prop? Thanks :)

packages/picasso/src/Avatar/Avatar.tsx
```diff
@@ -18,78 +18,68 @@ export interface Props extends StandardProps, HTMLAttributes<HTMLDivElement> {
...
   const renderAvatar = () => {
     if (src) {
       return (
         <ImageAvatar
           alt={alt}
-          classes={classes}
           className={className}
           name={name}
-          size={size}
+          /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
```
denieler, line 46: doesn't it work without this disable rule?

packages/picasso/src/Avatar/Avatar.tsx
denieler: it's how in most Picasso components it's done.

default values in props have some benefits though, we can discuss that during our daily

packages/picasso/src/Avatar/utils/getClipPathCornerMask.ts
denieler: sorry, I didn't understand why :) `em` also keep proportions

packages/picasso/src/Avatar/utils/getDimensions.ts
denieler: ok, got it. `.6667` ratio just looks non accurate, because it supposed to be `.666666(6)`, right? so after the calculation it can be slightly shifted and it might be visible on retina displays or high resolution

packages/picasso/src/Avatar/IconAvatar/styles.ts
denieler: for this component to me parent class is `root` defined above. It doesn't have font-size set. So for this component this font-size is a "basic" font-size kinda. Why it will be not clear?

packages/picasso/src/Avatar/ImageAvatar/styles.ts
```diff
@@ -0,0 +1,30 @@
...
+      // be resolved - https://github.com/cssinjs/css-vendor/issues/74
+      '-webkit-clip-path': (props: Props) => getClipPathCornerMask(props.size),
+      width: (props: Props) => getDimensions(props.variant).width,
+      height: (props: Props) => getDimensions(props.variant).height
+    },
+    logoContainer: {
+      display: 'flex',
+      position: 'absolute',
+      bottom: `1rem`,
+      left: `1rem`
```
denieler, line 24: why can't we use `em` here?

packages/picasso/src/Avatar/AvatarWrapper/AvatarWrapper.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const useStyles = makeStyles<Theme>(styles, {
+  name: 'PicassoAvatarWrapper'
+})
+
+const AvatarWrapper: FunctionComponent<Props> = ({
+  children,
+  size,
+  ...rest
+}: Props) => {
```
dmaklygin1, line 22: this is interesting, because it has two prop definitions:
1. FunctionComponent<Props>
2. `Props`

I'm not sure, but seems, we decided to use only the second one: define function's arguments

packages/picasso/src/Avatar/IconAvatar/IconAvatar.tsx
```diff
@@ -0,0 +1,34 @@
...
+import cx from 'classnames'
+import { makeStyles, Theme } from '@material-ui/core/styles'
+
+import type { AvatarSizeType, VariantType } from '../Avatar'
+import ImagePlaceholder from '../ImagePlaceholder/ImagePlaceholder'
+import { Profile16 } from '../..'
+import styles from './styles'
+
+type Props = {
+  className?: string
```
dmaklygin1, line 11: not sure, but we usually extend Props from `BaseProps` to define `className, style and data-testid`

packages/picasso/src/Avatar/ImageAvatar/ImageAvatar.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+type LogoProps = {
+  size: AvatarSizeType
+} & JssProps
+
+const useStyles = makeStyles<Theme, Props>(styles, {
+  name: 'PicassoImageAvatar'
+})
+
+const renderLogo = ({ size, classes }: LogoProps) => {
```
dmaklygin1, line 29: for me, it looks like a common React component, but its name starts with a lower letter. Wdyt @TomasSlama about the naming?

packages/picasso/src/Avatar/ImagePlaceholder/ImagePlaceholder.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import cx from 'classnames'
+import { makeStyles, Theme } from '@material-ui/core/styles'
+
+import styles from './styles'
+import type { AvatarSizeType, VariantType } from '../Avatar'
+
+export type Props = {
+  className?: string
```
dmaklygin1, line 9: The same question about extending `BaseProps` to define a `className`: do we have a consistent decision about defining such props?

packages/picasso/src/Avatar/ImagePlaceholder/styles.ts
```diff
@@ -0,0 +1,18 @@
...
+  createStyles({
+    root: {
+      clipPath: (props: Props) => getClipPathCornerMask(props.size),
+      // we can remove this prefix as soon as this issue will
+      // be resolved - https://github.com/cssinjs/css-vendor/issues/74
+      '-webkit-clip-path': (props: Props) => getClipPathCornerMask(props.size),
+      backgroundColor: palette.grey.main,
+      width: (props: Props) => getDimensions(props.variant).width,
+      height: (props: Props) => getDimensions(props.variant).height
+    }
```
dmaklygin1, line 17: will it work?

```suggestion
    root: (props: Props) => {
      const clipPath = getClipPathCornerMask(props.size)
      const dimensions = getDimensions(props.variant)
      
      return {
        clipPath,
        // we can remove this prefix as soon as this issue will
        // be resolved - https://github.com/cssinjs/css-vendor/issues/74
        '-webkit-clip-path': clipPath,
        backgroundColor: palette.grey.main,
        width: dimensions.width,
        height: dimensions.height
      }
    }
```

packages/picasso/src/PageTopBarMenu/__snapshots__/test.tsx.snap
```diff
@@ -15,11 +15,11 @@ exports[`PageTopBarMenu renders 1`] = `
           class="PicassoContainer-centerAlignItems PicassoContainer-flex PicassoUserBadge-root"
         >
           <div
-            class="PicassoAvatar-root PicassoAvatar-xxsmall"
+            class="PicassoAvatarWrapper-root PicassoAvatarWrapper-xxsmall"
           >
             <img
               alt="Jacqueline Roque"
-              class="PicassoImage-root PicassoAvatar-image PicassoAvatar-square PicassoAvatar-xxsmall PicassoAvatar-clippedCorner PicassoUserBadge-avatar"
+              class="PicassoImage-root PicassoImageAvatar-image PicassoImageAvatar-image PicassoUserBadge-avatar"
```
dmaklygin1, line 22: hm, seems the classname is duplicated

packages/picasso/src/UserBadge/__snapshots__/test.tsx.snap
```diff
@@ -9,13 +9,13 @@ exports[`UserBadge auto center alignment 1`] = `
       class="PicassoContainer-centerAlignItems PicassoContainer-flex PicassoUserBadge-root"
     >
       <div
-        class="PicassoAvatar-root PicassoAvatar-xsmall"
+        class="PicassoAvatarWrapper-root PicassoAvatarWrapper-xsmall"
       >
         <div
-          class="PicassoAvatar-textContainer PicassoAvatar-square PicassoAvatar-xsmall PicassoAvatar-clippedCorner PicassoUserBadge-avatar"
+          class="PicassoImagePlaceholder-root PicassoImagePlaceholder-root PicassoUserBadge-avatar"
```
dmaklygin1, line 15: here, we also have a duplicated className

packages/picasso/src/Avatar/Avatar.tsx
```diff
@@ -18,78 +18,68 @@ export interface Props extends StandardProps, HTMLAttributes<HTMLDivElement> {
...
+    return <IconAvatar className={className} size={size!} variant={variant!} />
   }
 
   return (
-    <div {...rest} className={cx(classes.root, classes[size])}>
+    /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
+    <AvatarWrapper size={size!} {...rest}>
       {renderAvatar()}
-    </div>
+    </AvatarWrapper>
```
dmaklygin1, line 82: @denieler, @TomasSlama, as this Component was deeply reworked, wdyt should we carefully rewrite Avatar's unit tests using new rules or not?

dmaklygin1 (commented):
Good work, left some comments :)

packages/picasso/src/Avatar/ImageAvatar/ImageAvatar.tsx
TomasSlama: you are right, we can use it as normal component. 👍

packages/picasso/src/Avatar/ImagePlaceholder/ImagePlaceholder.tsx
TomasSlama: For me this subcomponents inside /Avatar are only helper components to make code cleaner and they are not supposed to be reused anywhere else, but in Avatar. So I know what specific props I need. But I can extend Props from BaseProps so we are consistent with rest of codebase. 👍

packages/picasso/src/Avatar/AvatarWrapper/AvatarWrapper.tsx
TomasSlama: yep, it seems we are not consistent with this in our codebase, I'm not sure what is better solution of these two, definitely not both :)

packages/picasso/src/AccountSelect/__snapshots__/test.tsx.snap
```diff
@@ -32,16 +32,17 @@ exports[`AccountSelect renders 1`] = `
                   class="PicassoContainer-flexStartAlignItems PicassoContainer-flex PicassoUserBadge-root"
                 >
                   <div
-                    class="PicassoAvatar-root PicassoAvatar-xsmall"
+                    class="PicassoAvatarWrapper-root PicassoAvatarWrapper-size PicassoAvatarWrapper-size PicassoAvatarWrapper-corner PicassoAvatarWrapper-corner"
```
dmaklygin1, line 35: that's interesting, it still duplicates two classNames

packages/picasso/src/PageTopBarMenu/__snapshots__/test.tsx.snap
TomasSlama: the material ui duplicates classnames when we are using dynamic values in class. In reality it adds different numbers as suffix to this class.

packages/picasso/src/Avatar/Avatar.tsx
dmaklygin1: @TomasSlama wdyt if we memoize the result of `renderAvatar` function

packages/picasso/src/AccountSelect/__snapshots__/test.tsx.snap
TomasSlama: yeah, MUI does it internally when we use dynamic values in styles, it has different number as suffix, but snapshots are not showing these

packages/picasso/src/Avatar/IconAvatar/IconAvatar.tsx
```diff
@@ -0,0 +1,36 @@
...
+  size: AvatarSizeType
+}
+
+const useStyles = makeStyles<Theme>(styles, {
+  name: 'PicassoIconAvatar'
+})
+
+const IconAvatar = ({
+  size,
+  'data-testid': dataTestId,
```
dmaklygin1, line 20: will it work, btw?
```
'data-testid': dataTestId = 'photo-placeholder'
```

packages/picasso/src/Avatar/test.tsx
```diff
@@ -5,39 +5,51 @@ import { OmitInternalProps } from '@toptal/picasso-shared'
...
+  return render(<Avatar alt={alt} name={name} src={src} size={size} />)
 }
 
 describe('Avatar', () => {
   it('renders with initials', () => {
-    const { container } = renderAvatar({ name: 'Jacqueline Roque' })
+    const { getByText } = renderAvatar({ name: 'Jacqueline Roque' })
 
-    expect(container).toMatchSnapshot()
+    expect(getByText('JR')).toBeVisible()
```
dmaklygin1, line 17: I would use `screen.getByTestId`, this is the preferable way to select elements.
@denieler wdyt?

packages/picasso/src/PageTopBarMenu/__snapshots__/test.tsx.snap
```diff
@@ -15,11 +15,11 @@ exports[`PageTopBarMenu renders 1`] = `
           class="PicassoContainer-centerAlignItems PicassoContainer-flex PicassoUserBadge-root"
         >
           <div
-            class="PicassoAvatar-root PicassoAvatar-xxsmall"
+            class="PicassoAvatarWrapper-root PicassoAvatarWrapper-size PicassoAvatarWrapper-size PicassoAvatarWrapper-corner PicassoAvatarWrapper-corner"
```
dmaklygin1, line 18: duplication is also still here

packages/picasso/src/Avatar/IconAvatar/IconAvatar.tsx
```diff
@@ -0,0 +1,36 @@
...
+  className,
+  style
+}: Props) => {
+  const classes = useStyles()
+
+  return (
+    <Profile16
+      className={cx(className, classes.root, classes[`${size}Icon`])}
+      color='white'
+      data-testid={dataTestId || 'photo-placeholder'}
```
teimurjan, line 30: Why do we set a default value here? I'd avoid constant values (even if they are default)

dmaklygin1 (approved):

teimurjan (approved):

packages/picasso/src/Avatar/TextAvatar/styles.ts
```diff
@@ -0,0 +1,25 @@
...
+    },
+
+    xxsmall: { fontSize: '1em' },
+    xsmall: { fontSize: '1em' },
+    small: { fontSize: '2em' },
+    medium: { fontSize: '3em' },
+    large: { fontSize: '5em' },
+
+    smallFont: {
+      fontSize: '0.666666667em'
```
denieler, line 20: in such cases maybe it's better to use `rem('*px')` except, of course, if we need here dynamic font-size based on parent font-size

PR #2212, opened 2021-11-08
refactor: [FX-2180] Make ShowMore BASE 2 compliant

LashaJini:
[FX-2180](https://toptal-core.atlassian.net/browse/FX-2180?atlOrigin=eyJpIjoiM2M1ODhmYzM4MDNhNDkzNjlkNDNkYmY5ZDQ0YTg3Y2YiLCJwIjoiaiJ9)

### Description

- `ShowMore` component should have action button  with an `iconPosition` “right” instead of `Link` .
- Button should be same as this [button](https://app.abstract.com/projects/1b06c884-06af-482a-bf12-a82f521a19a1/branches/master/commits/01be317bc4bbf19461c9eb43a5304f732717305d/files/f58fa9b1-6361-4d06-a6c7-d57f8c9d715d/layers/BF5B326D-31FD-4D85-8D6A-3D81FB59944C?mode=design&sha=01be317bc4bbf19461c9eb43a5304f732717305d)

### How to test

- [ShowMore story](https://picasso.toptal.net/fx-2180-showmore-use-button/?path=/story/components-showmore--showmore)

### Screenshots

### Review

<details>
<br />

List of available commands:

</details>


OleksandrNechai (approved):
We might have a problem approving with Milos. Looks like they have changed in the approach with underline decoration for the action button. There should be no such decoration. 
If he rises this - just remove decoration in this same PR. It does not seem to be very complicated atm.

TomasSlama (approved):

packages/picasso/src/Button/styles.ts
```diff
@@ -52,7 +52,8 @@ export default (theme: Theme) => {
       borderRadius: sizes.borderRadius.small,
       border: `solid ${sizes.borderWidth} ${palette.grey.light2}`,
       fontSize: '1rem',
-      transition: `all ${transitions.duration.short}ms ${transitions.easing.easeOut}`,
```
LashaJini, line 55: There is explicit `transitionProperty` after this line, so using `all` in this case, is not necessary. Also, transition didn't look good when hovering on button which shows underline of text.

simplescreenrecorder-2021-11-08_15 04 31

packages/picasso/src/ShowMore/ShowMore.tsx
```diff
@@ -60,23 +60,25 @@ export const ShowMore = forwardRef<HTMLSpanElement, Props>(function ShowMore(
...
           onClick={() => {
             setShownMore(!shownMore)
             onToggle()
           }}
           className={classes.toggleText}
-          variant='action'
+          iconPosition='right'
         >
-          {shownMore ? lessText : moreText}
+          <div className={classes.textWrapper}>
```
LashaJini, line 71: Currently icon had `margin-left` and `rotation` applied together, which cause something like this:

margin-left-transform

Yes, we could instead applied different margin, but I think it is not a good developer experience and instead I've wrapper text in `div` element and applied right margin there.

packages/picasso/src/Button/styles.ts
TomasSlama: Did you check, if it didn't affect other variants of Button?

packages/picasso/src/ShowMore/ShowMore.tsx
TomasSlama: wouldn't `span` make more sense? `div` is block element, it seems weird to me to have block element inside button

packages/picasso/src/ShowMore/styles.ts
```diff
@@ -1,19 +1,24 @@
...
   createStyles({
     toggleText: {
       display: 'flex',
       alignItems: 'center',
-      marginTop: '0.5rem'
+      marginTop: '0.5rem',
+      transition: 'none'
+    },
+    textWrapper: {
+      marginRight: rem('8px'),
```
TomasSlama, line 13: on line 9 we already use `0.5rem`, I don't think we need to use `rem()` here. It should be used only when weird spacings need to be converted to rem, for example, 11px, 7px, ...

packages/picasso/src/ShowMore/ShowMore.tsx
TomasSlama: also when I check examples of `Button.Action`

```
<Button.Action icon={<Link16 />} iconPosition='right'>
  Default
</Button.Action>
```

Is it necessary to use it differently?

packages/picasso/src/Button/styles.ts
LashaJini: It didn't affect. If I didn't miss anything, other buttons animations only include border, color and background transitions.

packages/picasso/src/ShowMore/styles.ts
OleksandrNechai: still not fixed looks like :smile:

OleksandrNechai (approved):
LGTM :+1:

packages/picasso/src/ShowMore/styles.ts
```diff
@@ -5,15 +5,17 @@ export default ({ palette }: Theme) =>
...
-      marginTop: '0.5rem'
+      marginTop: '0.5rem',
+      transition: 'none',
+      fontWeight: 'normal'
     },
     iconWrapper: {
-      transform: 'rotate(90deg)',
-      fontSize: '0.6em',
-      marginLeft: '1em'
+      fontSize: '1rem !important',
```
LashaJini, line 13: This is necessary to override child's `font-size`, which already has `!important` applied to it. Without this, other components would require huge refactoring.

LashaJini:
@toptal-bot run all

PR #2223, opened 2021-11-11
fix: subpixel calculations for table separators

13ANM:
[FX-2260]
[SPT-2114](https://toptal-core.atlassian.net/browse/SPT-2114)

### This is a **POC**, open discussion to possible alternatives (current PR implementation is not definitive)

Chromium based browsers have a different rendering engine that handles the subpixel calculations different than other browsers, meaning they don't really support subpixel rendering and this leads to integer approximations on random elements that, in our case for the table rows borders, will become inconsistent.

For some users, depending on their monitor this will also happen without needing to zoom into the page, reason for which we should try and find a workaround for this until chromium will accept subpixel rendering.

With this approach we can add the borders to the table cells that will adjust based on the adjacent borders, so at least now, when they adjust while zooming, they should adjust all at once

### How to test

1. Go to [deploy](https://picasso.toptal.net/SPT-2114-inconsistent-table-separators/?path=/story/components-table--table) on a chromium based browser (Chrome, Brave, Edge, Vivaldi etc.)
2. Use all zoom variants provided by the browser and check if there are any discrepancies between the table separators
3. Compare to [current](https://picasso.toptal.net/?path=/story/components-table--table) as in the video below

https://user-images.githubusercontent.com/79158829/141304105-3fd779fe-01fd-417e-8064-0613db4aa8e4.mov

### Review

<details>
<br />

List of available commands:

</details>

[FX-2260]: https://toptal-core.atlassian.net/browse/FX-2260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/TableCell/styles.ts
+++ b/packages/picasso/src/TableCell/styles.ts
@@ -33,11 +33,14 @@ const narrowCellStyles = {
   paddingRight: '0.5rem'
 }
 
-export default ({ palette, typography }: Theme) =>
+export default ({ sizes, palette, typography }: Theme) =>
   createStyles({
     root: regularCellStyles,
     compact: compactCellStyles,
     narrow: narrowCellStyles,
+    bordered: {
+      borderBottom: `${sizes.borderWidth} solid ${palette.grey.lighter2}`
+    },
     header: {
       fontWeight: typography.fontWeights.semibold,
       color: palette.text.primary,
--- a/packages/picasso/src/Table/__snapshots__/test.tsx.snap
+++ b/packages/picasso/src/Table/__snapshots__/test.tsx.snap
@@ -6,7 +6,7 @@ exports[`Table renders 1`] = `
     class="Picasso-root"
   >
     <table
-      class="MuiTable-root"
+      class="MuiTable-root Table-root"
     >
       <thead
         class="MuiTableHead-root PicassoTableHead-root"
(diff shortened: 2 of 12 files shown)


TomasSlama:
For me, it works differently than for you. The separator is the same for every row, with your fix it just holds the height of the separator, so it doesn't seem so bugy as for you :)
image
Left is yours fix, right is master.

I don't mind solution

13ANM:
@TomasSlama Yes, this issue also depends on the screen and OS that you're using I guess, we can push this forward only if it fixes it for everyone and there are no downsides, as it was only reported by a few users

sergiubutnarasu:
I have the same behavior as you in the staff portal, but in the Picasso storybook, I have the same behavior as @TomasSlama mentioned.

For me, your solution is working.

OndrejTuma:
I'm just wondering if the border on table row should be removed then 🤔 
Also some snapshots needs updating, maybe use something like `yarn test:unit --updateSnapshot`?

packages/picasso/src/TableCell/styles.ts
```diff
@@ -33,11 +33,14 @@ const narrowCellStyles = {
...
 }
 
-export default ({ palette, typography }: Theme) =>
+export default ({ sizes, palette, typography }: Theme) =>
   createStyles({
     root: regularCellStyles,
     compact: compactCellStyles,
     narrow: narrowCellStyles,
+    bordered: {
+      borderBottom: `${sizes.borderWidth} solid ${palette.grey.lighter2}`
```
denieler, line 42: can you please explain why do we need the bottom border and on TableCell and on TableRow?

13ANM: @denieler I've removed it from the `TableRow`, the idea is to move the border to the cells, so in these cases when we zoom they will calculate differently, based on the adjacent borders

13ANM:
As discussed, we can abandon this for now since it affects only a few users and wait for a `chromium` or `material-ui` update

PR #2233, opened 2021-11-12
fix(TagSelector): closing when being scrolled

OleksandrNechai:
[FX-2068]

### Description

TagSelector and Autocomplete were closing when you clicked on scroll in its overlay. This bug is fixed.

### How to test

- Open https://picasso.toptal.net/fx-fix-closing-TagSelector-and-Autocomplete-when-scrolling/?path=/story/forms-tagselector--tagselector, Default example. 
- Scroll with the mouse by clicking on scroll now using the wheel.
- `TagSelector` should not close

### Review

<details>
<br />

List of available commands:

</details>

[FX-2068]: https://toptal-core.atlassian.net/browse/FX-2068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/ScrollMenu/ScrollMenu.tsx
+++ b/packages/picasso/src/ScrollMenu/ScrollMenu.tsx
@@ -75,12 +75,12 @@ const ScrollMenu: FunctionComponent<Props> = props => {
       className={classes.menu}
       style={style}
       role={role}
+      // eslint-disable-next-line react/jsx-props-no-spreading
+      {...rest}
       onMouseDown={e => {
         // ScrollMenu is used in dropdowns. Prevents blur --> dropdown close when scrolled.
         e.preventDefault()
       }}
-      // eslint-disable-next-line react/jsx-props-no-spreading
-      {...rest}
     >
       {fixedHeader}
       <div ref={menuRef} className={classes.scrollView} onBlur={onBlur}>
--- a/packages/picasso/src/ScrollMenu/ScrollMenu.tsx
+++ b/packages/picasso/src/ScrollMenu/ScrollMenu.tsx
@@ -75,6 +75,10 @@ const ScrollMenu: FunctionComponent<Props> = props => {
       className={classes.menu}
       style={style}
       role={role}
+      onMouseDown={e => {
+        // ScrollMenu is used in dropdowns. Prevents blur --> dropdown close when scrolled.
+        e.preventDefault()
+      }}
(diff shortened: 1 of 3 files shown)


packages/picasso/src/ScrollMenu/ScrollMenu.tsx
```diff
@@ -75,6 +75,10 @@ const ScrollMenu: FunctionComponent<Props> = props => {
       className={classes.menu}
       style={style}
       role={role}
+      onMouseDown={e => {
+        // ScrollMenu is used in dropdowns. Prevents blur --> dropdown close when scrolled.
+        e.preventDefault()
+      }}
```
OndrejTuma, line 81: this is unlikely, but if someone passes his own `onMouseDown` implementation, the bug will return

OndrejTuma (approved):

packages/picasso/src/ScrollMenu/ScrollMenu.tsx
OleksandrNechai: good catch! :) I might want to pass it right after the `rest`

TomasSlama: so, does it mean, that we are not allowing users to set `onMouseDown`, is it intentional?
```suggestion
     onMouseDown={e => {
        // ScrollMenu is used in dropdowns. Prevents blur --> dropdown close when scrolled.
        e.preventDefault()
        
        if (typeof rest.onMouseDown === 'function) {
          rest.onMouseDown(e)
        }
      }}
```

How about this?

TomasSlama (approved):
I left comment on `onMouseDown`, otherwise LGTM 👍

packages/picasso/src/ScrollMenu/ScrollMenu.tsx
OleksandrNechai: This is also an option. I thought we don't need it as the component is internal. But I like this variant.

OleksandrNechai: Checked it again. Looks like the current interface of a component does not really have `onMouseDown`. So TS does not allow to pass it and it is kinda not expected to have it. I don't see the reason to add it currently.

PR #2242, opened 2021-11-15
feat: [FX-2257] Remove unique class name prefix for ssr support

deniskaber:
[FX-2257]

### Description

Thid adds a parameter for PicassoProvider to disable adding unique class name prefixes. This is needed to support SSR usage of Picasso for Site Acqusition projects.

Unique class name causes a mismatch between server side rendered content and client side during rehydration process.

### How to test

- Check [this example](https://picasso.toptal.net/sat-820-htmlshell-ssr/?path=/story/components-picasso--picasso#classname-prefix-disabled) . Produced css classes should not have unique prefixes.
### Screenshots

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2257]: https://toptal-core.atlassian.net/browse/FX-2257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/picasso-provider/src/Picasso/Picasso.tsx
```diff
@@ -162,6 +163,8 @@ interface PicassoProps extends TextLabelProps {
   theme?: ThemeOptions
   /** Disables transitions for components like Loader, to make testing easier */
   disableTransitions?: boolean
+  /** Disables unique class name prefix, to prevent hydration issues after SSR */
+  disableClassNamePrefix?: boolean
```
OleksandrNechai, line 167: @denieler wdyt about having a centralized option to picasso `isServer` which would be used for all kind of alternation of logic depending whether server or not?

packages/picasso-provider/src/Picasso/Picasso.tsx
```diff
@@ -162,6 +163,8 @@ interface PicassoProps extends TextLabelProps {
   theme?: ThemeOptions
   /** Disables transitions for components like Loader, to make testing easier */
   disableTransitions?: boolean
+  /** Applies adjustments needed to work properly during SSR and hydration */
+  isSsrMode?: boolean
```
denieler, line 167: I would not use `ssrMode` name, but to be more explicit `disableDynamicPicassoClassNames` - something like this

packages/picasso-provider/src/Picasso/Picasso.tsx
```diff
@@ -190,7 +194,7 @@ const Picasso: FunctionComponent<PicassoProps> = ({
   const generateClassName = createGenerateClassName({
     // if there are multiples instances of Picasso
     // on the page we want each set of styles to be unique
-    seed: generateRandomStringOrGetEmptyInTest()
+    seed: isSsrMode ? '' : generateRandomStringOrGetEmptyInTest()
```
denieler, line 197: what will happen if we would have multiple Picasso providers on the page in SSR? I assume then we might have class name collisions, right?

packages/picasso-provider/src/Picasso/Picasso.tsx
deniskaber: Hey @denieler! Originally I had a more explicit name for this - `disableClassNamePrefix` , but there was a suggestion to fix it earlier in comments to this PR. Could you please take a look and give feedback there?

packages/picasso-provider/src/Picasso/Picasso.tsx
deniskaber: You are pobably right, and we would have to tackle this problem if it would ever occur.

At this point, non-deterministic classNames with random prefix won't work in SSR in any case, so we have to fix that first.

packages/picasso-provider/src/Picasso/Picasso.tsx
denieler: I don't expect much logic for Picasso provider at least. And additionally now we don't have any candidates except this particular case, so I would call it `disableClassNamePrefix` and start using `isSSRMode` or `isServer` later if 3 or more occurrences will be found

packages/picasso-provider/src/Picasso/Picasso.tsx
denieler: this might become a huge problem if eventually multiple projects will start using SSR Picasso independently and at some point will be displayed on the same page. This issue is very difficult to identify and in such a case it will become like a hidden bomb. So I would try to fix it right now and not postpone the solution.

@deniskaber can you please explain why this prefix doesn't work for you on SSR? because it can generate different results on the server side and on client side, when in both places Picasso Provider is used?

deniskaber: @denieler Yes, correct. This is exactly the issue

denieler (approved):

TomasSlama (approved):

deniskaber:
I added a follow-up ticket to support a case of having multiple instances of Picasso on the page - [SAT-832](https://toptal-core.atlassian.net/browse/SAT-802)

packages/picasso-provider/src/Picasso/Picasso.tsx
denieler: @deniskaber look - https://github.com/reactwg/react-18/discussions/111, it looks like in React 18 our problem can be solved easily

PR #2285, opened 2021-11-25
refactor: [FX-2146] Make List base 2 compliant

LashaJini:
[FX-2146](https://toptal-core.atlassian.net/browse/FX-2146)

[Abstract](https://share.goabstract.com/5db01096-264a-4989-b5aa-5662957282d8?mode=build&sha=4c5feb75fc43678e79189e577039f5c08950c62c)

> NOTE: Design for ordered list item's bullet Nr.1 is incorrect. Width and margin shouldn't be as is in the design. Numbers must satisfy `00.`

### Description

- Unordered list has correct bullet color. Should be `#455065`
- Line height of a list item should be `22px`
- Spacing between list items should be `8px`

### How to test

- [Picasso List component](https://picasso.toptal.net/fx-2146-make-list-base-2-compliant/?path=/story/components-list--list)

### Screenshots

| Screenshot from 2021-11-26 11-23-01  | Screenshot from 2021-11-26 14-24-04 |
| Screenshot from 2021-11-26 12-12-35 | Screenshot from 2021-11-26 12-13-33 |
| Screenshot from 2021-11-26 12-12-43 | Screenshot from 2021-11-26 12-13-48 |

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


packages/picasso/src/List/styles.ts
```diff
@@ -1,10 +1,23 @@
...
-      margin: 0
+      margin: 0,
+      height: '1.375em',
+      lineHeight: '1.375em',
+
+      '& > li': {
+        marginBottom: '0.5em'
+      },
+
+      '& > :last-child': {
```
LashaJini, line 16: We don't want margin on last child list item.

packages/picasso/src/List/List.tsx
```diff
@@ -30,10 +33,22 @@ export const List = (props: Props) => {
   )
 
   if (variant === 'unordered') {
-    return <ul {...listProps}>{listItems}</ul>
+    return (
+      <ul
+        className={cx(classes.root, classes.ul)}
```
OndrejTuma, line 38: Maybe add also `props.className` here? In case someone used it like this before

dmaklygin1: wdyt, @LashaJini if we use the same names for classnames as for variants. Let's see an example: https://github.com/toptal/picasso/blob/b0a00a0a8d3d92fcd7930a4a0bf0fdc09103ac6c/packages/picasso/src/Button/Button.tsx#L161

packages/picasso/src/List/test.tsx
```diff
@@ -0,0 +1,28 @@
...
+  render(<List {...props}>{props.children}</List>)
+
+describe('List', () => {
+  it('renders unordered list', () => {
+    const { container } = renderList({
+      children: [],
+      variant: 'unordered'
+    })
+
+    expect(container).toMatchSnapshot()
```
TomasSlama, line 17: If I'm not mistaken, we should use snapshots only when we reuse MUI component, for our own components we should avoid snapshots

packages/picasso/src/ListItem/ListItem.tsx
```diff
@@ -53,7 +52,9 @@ export const ListItem = (props: Props) => {
           inline
           right='small'
           justifyContent='flex-end'
-          className={classes.iconContainer}
+          className={cx(classes.iconContainer, {
```
TomasSlama, line 55: why not just `classes[variant]`? it will end up having one classname. In your approach you would end up with two classnames when `variant === 'ordered'`

packages/picasso/src/List/styles.ts
```diff
@@ -1,10 +1,23 @@
...
 
-export default () =>
+export default ({ palette }: Theme) =>
   createStyles({
     root: {
       listStyle: 'none',
       padding: 0,
-      margin: 0
+      margin: 0,
+      height: '1.375em',
```
TomasSlama, line 9: this is not correct, why having static height on whole list?, it breaks `list-custom-snap.png`

TomasSlama (changes requested):

packages/picasso/src/List/List.tsx
```diff
@@ -19,21 +20,37 @@ const useStyles = makeStyles<Theme>(styles, { name: 'PicassoList' })
...
+
+  return React.createElement(
+    ListTag,
+    {
+      className: cx(classes.root, classes[variant]),
+      'data-testid': dataTestId,
+      ...rest
+    },
+    listItems
+  )
```
OndrejTuma, line 53: why not use jsx here? 🤔

packages/picasso/src/List/styles.ts
```diff
@@ -1,10 +1,17 @@
...
       padding: 0,
-      margin: 0
+      margin: 0,
+
+      '& > li:not(:last-child)': {
+        marginBottom: '0.5em'
+      }
+    },
+    ul: {
+      color: palette.grey.dark
```
TomasSlama, line 15:
```suggestion
      color: palette.text
```

It is the same color, but more specific

packages/picasso/src/ListItem/ListItem.tsx
TomasSlama: ? why resolve?

TomasSlama (commented):
changeset is missing, added some more questions

LashaJini:
> What is the benefit of using `createElement`? :thinking:

The only way to use jsx was with self-closed tag: `<ListTag {...}/>`, and but `List` should have child elements, when I used `<ListTag {...}>{listItems}</ListTag>` it threw [ts(2322)](https://stackoverflow.com/questions/63252161/ts2322-property-children-does-not-exist-on-type-intrinsic-attributes-and-prop) error about `intrinsic` attributes. (Same thing Ondra was wondering and I've answered him, but couldn't reference that comment, so I just copy pasted here)

TomasSlama (approved):
LGTM 👍, only please remove the old image snapshots from `List/__image_snapshots__/`

.changeset/tidy-sloths-invite.md
```diff
@@ -0,0 +1,12 @@
+---
+'@toptal/picasso': patch
+'@toptal/picasso-lab': patch
+---
+
+- Unordered list item's bullet has primary text color, which is dark gray
```
denieler, line 6: maybe we should make a slightly different format?

```
# List

- Unordered list item's bullet has primary text color, which is dark gray
...
```

.changeset/tidy-sloths-invite.md
```diff
@@ -0,0 +1,12 @@
+---
+'@toptal/picasso': patch
```
denieler, line 2: it's not a bug fix, it's an intentional change of the design, I think it should be treated as a feature, so we should bump minor version here

packages/picasso/src/List/List.tsx
```diff
@@ -17,23 +18,36 @@ export type Props = BaseProps & {
...
+
 export const List = (props: Props) => {
   const classes = useStyles()
-  const { variant, children, start = 1, ...rest } = props
+  const { variant, children, start = 1, className, ...rest } = props
+
+  const listItems = React.Children.map(children, (child, index) => {
+    if (!React.isValidElement(child)) {
+      return child
+    }
```
denieler, line 33: why do we need this check?

packages/picasso/src/List/List.tsx
```diff
@@ -17,23 +18,36 @@ export type Props = BaseProps & {
...
+  })
 
-  if (variant === 'unordered') {
-    return <ul {...listProps}>{listItems}</ul>
-  }
+  const ListTag = Tags[variant]
 
-  return <ol {...listProps}>{listItems}</ol>
+  return (
+    <ListTag
```
denieler, line 44: nice solution with `ListTag` 👍

packages/picasso/src/List/styles.ts
```diff
@@ -1,10 +1,17 @@
...
   createStyles({
     root: {
       listStyle: 'none',
       padding: 0,
-      margin: 0
+      margin: 0,
+
+      '& > li:not(:last-child)': {
+        marginBottom: '0.5em'
+      }
```
denieler, line 12: to be this style looks wrong because each our component should be self-sufficient, its state and look should be defined only by its props. But in this case, the look of ListItem in fact defined by the parent component.

packages/picasso/src/List/test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  it('renders unordered list', () => {
+    const { getByTestId } = renderList({
+      children: [],
+      variant: 'unordered',
+      'data-testid': testIds.ul
+    })
+
+    const list = getByTestId(testIds.ul) as HTMLUListElement
+
+    expect(list.children).toHaveLength(0)
```
denieler, line 25: what exactly we are trying to test here? at least I see that the description of the test `enders unordered list` does not correspond to the test expectations

packages/picasso/src/List/List.tsx
```diff
@@ -17,23 +18,36 @@ export type Props = BaseProps & {
...
       variant,
       index: index + start - 1
     })
-  )
+  })
 
-  if (variant === 'unordered') {
-    return <ul {...listProps}>{listItems}</ul>
-  }
+  const ListTag = Tags[variant]
```
denieler, line 41: this variable probably can be moved out from the iteration (`map`), because all list items would have the same tag

packages/picasso/src/List/List.tsx
LashaJini: It was Tomas idea.

packages/picasso/src/List/styles.ts
LashaJini: Yes, kinda, but only parent `List` knows which element is last, thus we can't have this styling on child `ListItem`.

denieler: but `List` can pass a "flag" to the last list item to show that it is last, for example (it's only one of the solutions)

packages/picasso/src/List/List.tsx
LashaJini: But `ListTag` is outside `map` already. Am I missing something?

packages/picasso/src/List/List.tsx
LashaJini: To avoid `tsserver 2769: no overload matches this call`.

denieler: stupid typescript 😃

packages/picasso/src/List/List.tsx
denieler: sorry, probably my bad 🙇‍♂️

packages/picasso/src/List/styles.ts
LashaJini: I will refactor it.

packages/picasso/src/List/utils/generateListItems.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+import List from '../List'
+
+/* eslint-disable react/no-array-index-key */
+const generateListItems = (total: number, listItemProps?: any) =>
```
LashaJini, line 6: I have to use this function in two places, 1 in unit test and 1 in cypress test.

packages/picasso/src/List/test.tsx
LashaJini: @denieler refactored!

packages/picasso/src/List/styles.ts
LashaJini: @denieler refactored!

.changeset/tidy-sloths-invite.md
LashaJini: @denieler done!

.changeset/tidy-sloths-invite.md
LashaJini: @denieler done!

cypress/component/List.spec.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import { List, Referrals16 } from '@toptal/picasso'
+import { mount } from '@cypress/react'
+import { TestingPicasso } from '@toptal/picasso/test-utils'
+
+/* eslint-disable react/no-array-index-key */
+const generateListItems = (total: number, listItemProps?: any) =>
```
LashaJini, line 7: I don't like re-defining same function in two places, but can't figure out another solution. Any suggestions? The other one is in `picasso/src/List/utils/generateListItems.tsx`

.changeset/tidy-sloths-invite.md
```diff
@@ -0,0 +1,11 @@
+---
+'@toptal/picasso': minor
+'@toptal/picasso-lab': patch
```
denieler, line 3: if we have 2 different bumps for 2 packages it means each of them need to have its own changelog. So I would suggest to have 1 changeset file for picasso with its changes and 1 changeset file for picass-lab

cypress/component/List.spec.tsx
denieler: it's ok to duplicate such a tiny function to keep modules independency

packages/picasso/src/List/styles.ts
```diff
@@ -1,10 +1,13 @@
...
 
-export default () =>
+export default ({ palette }: Theme) =>
   createStyles({
     root: {
       listStyle: 'none',
       padding: 0,
       margin: 0
+    },
+    ul: {
```
denieler, line 10: sorry, I'm not sure where this class is used?

denieler (approved):

.changeset/tidy-sloths-invite.md
denieler: @LashaJini sorry, I haven't seen the change in the changeset log

packages/picasso/src/List/styles.ts
LashaJini: It should be `unordered`, thanks.

.changeset/tidy-sloths-invite.md
LashaJini: @denieler I've separated them.

.changeset/short-onions-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@toptal/picasso-lab': patch
+---
+
+Update image snapshots
```
denieler, line 5: If this is the only change in this package - it looks like it doesn't affect our users. But in fact `picasso-lab` depends on `picasso` package, so the real reason why we need to bump a version is that `picasso` package version bump, so I would share this in a changelog

packages/picasso/src/ListItem/styles.ts
```diff
@@ -13,7 +13,18 @@ PicassoProvider.override(() => ({
...
+      marginBottom: '0.5em'
+    },
+    ordered: {
+      minWidth: '1.25em',
+      marginRight: '0.75em'
+    },
+    unordered: {
       minWidth: '1rem'
+    },
+    isLastElement: {
```
denieler, line 27:
```suggestion
    lastElement: {
```

packages/picasso/src/List/test.tsx
```diff
@@ -0,0 +1,42 @@
...
+  it('renders unordered list', () => {
+    const { getByTestId } = renderList({
+      children: generateListItems(5),
+      variant: 'unordered',
+      'data-testid': testIds.ul
+    })
+
+    const list = getByTestId(testIds.ul) as HTMLUListElement
+
+    expect(list.tagName).toBe('UL')
```
denieler, line 26: why not to use snapshots here?

.changeset/tidy-sloths-invite.md
denieler: @LashaJini this is still valid ☝️

denieler:
https://github.com/toptal/picasso/pull/2285#discussion_r761867578

OndrejTuma (approved):

.changeset/tidy-sloths-invite.md
LashaJini: @denieler Can you suggest something please? 

I thought this was good enough.

> Unordered list item's bullet has dark gray `#455065` color.

packages/picasso/src/ListItem/styles.ts
LashaJini: Refactored!

packages/picasso/src/List/test.tsx
LashaJini: Not sure, thb. I didn't want to add additional snapshots, because I've already did cypress tests. @denieler Should I instead replace them with snapshot matches?

.changeset/short-onions-cheer.md
LashaJini: Not sure about this. Can you clarify please?

packages/picasso/src/List/test.tsx
denieler: no, I think they are needed for different purposes. So I don't see a problem having both at the same time 👍

.changeset/short-onions-cheer.md
denieler: I would write in changeset only this:

```
Bump of the @toptal/package
```

Because all the rest for picasso-lab our users don't care

.changeset/tidy-sloths-invite.md
denieler: yeah, I would just add `### List` heading before the changelog. Because it might not be clear in changelog for which component are all those changes

LashaJini: @denieler done!

.changeset/neat-glasses-compete.md
```diff
@@ -0,0 +1,16 @@
+---
+'@toptal/picasso': minor
+---
+
+### List
+
+Components changed:
+
+`packages/picasso/src/List` and `packages/picasso/src/ListItem`
```
denieler, line 9: I would maybe leave a link to https://picasso.toptal.net/?path=/story/components-list--list instead and call `List.Item` (as in docs) instead of `ListItem` (internal name)

denieler (approved):

PR #2308, opened 2021-12-06
chore: [FX-2178] Remove colored container borders

LashaJini:
[FX-2178](https://toptal-core.atlassian.net/browse/FX-2178)

### Description

- Colored containers should not have borders
- Only white container can have a border

### How to test

- Visit [Picasso Container Component](https://picasso.toptal.net/fx-2178-container-remove-colored-borders/?path=/story/layout-container--container). **Bordered variant** should have only one example. Click on `edit code`, and add `variant='red'` (or any other available color) to the `Container` component, it should not add borders to the element and typescript should also complain: Screenshot from 2021-12-06 11-56-48 Screenshot from 2021-12-06 11-58-22

### Screenshots

| Screenshot from 2021-12-06 11-31-22 | Screenshot from 2021-12-06 11-30-46 |

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


cypress/component/Container.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+              data-testid={`${color}-container`}
+              key={color}
+            >
+              {color} variant
+            </TestContainer>
+          ))}
+        </TestingPicasso>
+      )
+
+      colors.forEach(color => {
```
LashaJini, line 69: I am not sure if this is a good practice.

packages/picasso/src/Container/Container.tsx
```diff
@@ -129,7 +131,8 @@ export const Container = forwardRef<HTMLDivElement, Props>(function Container(
             `${kebabToCamelCase(justifyContent || '')}JustifyContent`
           ]]: justifyContent,
 
-          [classes.bordered]: bordered,
+          [classes.bordered]:
+            bordered && !variantsWithoutBorder.includes(variant as string),
```
OndrejTuma, line 135:
```suggestion
            bordered && variantsWithBorder.includes(variant as string),
```
WDYT?

cypress/component/Container.spec.tsx
OndrejTuma: it looks good to me 👍

packages/picasso/src/Container/Container.tsx
LashaJini: But there is only one `white` container that can actually have border. So, I think it's better like it is now.

OndrejTuma (approved):

cypress/component/Container.spec.tsx
OleksandrNechai: 1) No, it's probably not a good practice. Looks like we are testing implementation and I always against doing this.
2) In case we need to test few similar cases, we use
```
it.each([2,4,6])('should be even', (n) => {
    expect(n % 2).toBe(0);
});
3) This problem you are trying to test, I believe we can ensure with type system ;)
```

packages/picasso/src/Container/Container.tsx
```diff
@@ -62,6 +62,8 @@ export interface Props
   align?: PropTypes.Alignment
 }
 
+const variantsWithoutBorder = ['red', 'green', 'yellow', 'blue', 'grey']
```
OleksandrNechai, line 65:
```suggestion
const borderlessVariants = ['red', 'green', 'yellow', 'blue', 'grey']
```

packages/picasso/src/Container/Container.tsx
OleksandrNechai, line 135: Can we ensure bordered can only accept `white` as a variant by type system?
Probably cleaner solution would be to add one new variant:
```
['white', 'bordered-white', 'red', 'green', 'yellow', 'blue', 'grey']
```

packages/picasso/src/Container/test.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+describe('Container', () => {
+  it('renders default container', () => {
+    const { container } = renderContainer({ children: 'Some text' })
+
+    expect(container).toMatchSnapshot()
+  })
+
+  it.each([...colors] as VariantType[])(
```
OleksandrNechai, line 26: IMO the test is redundant. 1) Already tested with Cypress 2) If such tests needed we probably need to think how to improve API or check things with type system.

OleksandrNechai (changes requested):
I think we could do better with API of the component. Now nothing prevents user to do
`<Container variant='red' bordered />` which probably does not make much sense :man_shrugging:

packages/picasso/src/Container/Container.tsx
LashaJini: Actually, when it is applied only for `white` container it breaks, because `variant` is optional and some of the components are not using any of colored `variants`. If I remember this also breaks a lot of tests.

OleksandrNechai: I don't understand :)

OndrejTuma: That was my point, its easier to see one variant that allows border than 5 that doesn't :-)
But its a nit, up to you

LashaJini: My bad. It is not important, but I think I will be able to do as suggested.

OleksandrNechai: Sure! I appreciate you being candid! :) Please, bring API discussion to the team chat, see what others think, mine is only suggestion, it is not request to do exactly that :wink:

packages/picasso/src/Container/Container.tsx
```diff
@@ -10,20 +10,24 @@ import {
...
 
-export type VariantType = 'red' | 'green' | 'white' | 'yellow' | 'blue' | 'grey'
+type BorderableType = 'transparent' | 'white'
 
 const useStyles = makeStyles<Theme, Props>(styles, {
   name: 'PicassoContainer'
 })
 
-export interface Props
+export interface Props<V extends VariantType = VariantType>
```
LashaJini, line 30: @teimurjan :bowing_man:

.changeset/chilled-gifts-reply.md
```diff
@@ -0,0 +1,30 @@
+---
+'@toptal/picasso': minor
```
OleksandrNechai, line 2: hm, I thought it was breaking change, why minor?

cypress/component/Container.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { VariantType } from '@toptal/picasso/src/Container/styles'
+
+const colors: VariantType[] = [
+  'red',
+  'green',
+  'white',
+  'yellow',
+  'blue',
+  'grey'
+]
```
OleksandrNechai, line 14: we also have transparent now, why not test it on darker background?
why not have also bordered tests for white and transparent?

packages/picasso/src/Container/Container.tsx
```diff
@@ -157,3 +161,4 @@ Container.defaultProps = {
 }
 
 export default Container
+export { VariantType, BorderableType }
```
OleksandrNechai, line 164: Sure we want to export BorderableType?

OleksandrNechai (approved):

packages/picasso/src/Container/Container.tsx
LashaJini: Not really tbh. I just thought maybe someone would want it in the future 🤔

.changeset/chilled-gifts-reply.md
LashaJini: I will change it to major.

packages/picasso/src/Container/Container.tsx
LashaJini: Yeah, probably no1 will use it anyway.

packages/picasso/src/utils/forward-ref.ts
```diff
@@ -34,5 +34,5 @@ export const forwardRef = <P, T>(
   component: (props: P, ref: React.Ref<T>) => React.ReactElement | null
 ) =>
   (React.forwardRef(component) as unknown) as (
-    props: P
+    props: P & { ref?: React.Ref<T> }
```
teimurjan, line 37: It worked incorrectly before, and the final component returned by `forwardRef` didn't have `ref` prop.

Fixed it :)

PR #2312, opened 2021-12-07
feat(TextEditor): add new component

TomasSlama:
[FX-2293]

### Description

### How to test

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


packages/picasso-lab/src/TextEditor/hooks/useHandleAutofocus.tsx
```diff
@@ -0,0 +1,23 @@
...
+) => {
+  const firstRender = useRef(true)
+
+  useEffect(() => {
+    if (firstRender.current && editorRef.current && autofocus) {
+      // without setTimeout is focus called on last instance🤷‍♂️
+      setTimeout(() => {
+        editorRef.current.focus()
+        firstRender.current = false
+      }, 0)
```
TomasSlama, line 18: For some unknown reason, when we have multiple instances of editor, even with different id, without setTimeout was the `focus()` function called on wrong instance.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
```diff
@@ -0,0 +1,31 @@
...
+      const {
+        clipboard,
+        root: { innerHTML: editorValue }
+      } = editorRef.current
+
+      // is value set from outside
+      if (editorValue !== controllerValue) {
+        // TODO discuss with others, if we should sanitize here
+        // usecase - editing current job description
+        clipboard.dangerouslyPasteHTML(controllerValue)
```
TomasSlama, line 25: since `value` needs to be always HTML string, I'm not sure if we should rely on users to sanitize the value if they decide to change from outside of the editor. Or if we should sanitize it here by ourselves. Better safe than sorry. WDYT?

packages/picasso-lab/package.json
```diff
@@ -32,10 +32,12 @@
   },
   "dependencies": {
     "@toptal/picasso-provider": "0.4.1",
+    "@types/quill": "^2.0.9",
```
teimurjan, line 35: It should be in `devDependencies`, right?

packages/picasso-lab/src/TextEditor/TextEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  disabled?: boolean
+  /** unique identificator */
+  id: string
+  /**
+   * The maximum number of characters that the user can enter.
+   * If this value isn't specified, the user can enter an unlimited
+   * number of characters.
+   */
+  // TODO implement
+  maxlength?: number
```
teimurjan, line 32: Not sure about this one. I've never seen a WYSIWYG editor with a length limit. It'd be better to ask BASE about it.

packages/picasso-lab/src/TextEditor/hooks/useHandleAutofocus.tsx
```diff
@@ -0,0 +1,25 @@
...
+  { autofocus }: { autofocus: Props['autofocus'] }
+) => {
+  const firstRender = useRef(true)
+
+  useEffect(() => {
+    const editorApi = editorRef?.current
+
+    if (firstRender?.current && editorApi && autofocus) {
+      // without setTimeout is focus called on last instance🤷‍♂️
+      setTimeout(() => {
```
teimurjan, line 17: Can you clarify this one a bit? Extraction of a code piece into the timer queue is not a good solution. I think we'll be able to find a more elegant one.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeEvent.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { EditorRefType } from '../types'
+import { Props } from '../TextEditor'
+
+const useHandleChangeEvent = (
+  editorRef: EditorRefType,
+  { onChange }: { onChange: Props['onChange'] }
+) => {
+  useEffect(() => {
+    const { current: editorApi } = editorRef
```
teimurjan, line 12: We have `editorRef` meaning it contains `editor`, not `editorApi`. I'd rename `editorApi` to `editor`.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
teimurjan: Isn't it sanitized by Quill?

packages/picasso-lab/src/TextEditor/styles.ts
```diff
@@ -0,0 +1,8 @@
+import { createStyles } from '@material-ui/core/styles'
+
+export default () =>
+  createStyles({
+    root: {
+      height: 'calc(100% - 43px)' // TODO proper style editor
```
teimurjan, line 6: Are you going to fix it in this PR?

TomasSlama: @teimurjan no, we have another ticket for styling the editor. This was just a quick fix so it looks good in story.

packages/picasso-lab/src/TextEditor/hooks/useHandleAutofocus.tsx
TomasSlama: yeah, it would be great, I was struggling a bit with this one. The problem is, when you have two stories and you add `autofocus` on the first one, it will be triggered on the second one 🤷. I didn't come up with any idea why is that happening. But I will try to dig deeper.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
TomasSlama: nope 👉 [docs](https://quilljs.com/docs/modules/clipboard/#dangerouslypastehtml)

packages/picasso-lab/src/TextEditor/TextEditor.tsx
TomasSlama: @teimurjan I asked Milos, TextEditor will be used instead of textarea for the job description, and they are restricted with max length right now and it would be great to have it on TextEditor too.

packages/picasso-lab/src/TextEditor/hooks/useHandleAutofocus.tsx
teimurjan: https://github.com/toptal/picasso/pull/2317

teimurjan: 🔝 it solves the unsafe HTML problem as well

packages/picasso-lab/src/TextEditor/TextEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+import styles from './styles'
+import useInitEditor from './hooks/useInitEditor'
+import useHandleChangeEvent from './hooks/useHandleChangeEvent'
+import useHandleChangeFromController from './hooks/useHandleChangeFromController'
+import useDisableEditor from './hooks/useDisableEditor'
+import useHandleAutofocus from './hooks/useHandleAutofocus'
+
+type HTMLString = string
+
+export interface Props extends BaseProps {
```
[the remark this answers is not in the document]
LashaJini: Tbh, this confuses me. Will this text editor have only `HTML` as input, or it can also have `markdown` (in the future).

LashaJini:
When I visit texteditor component in picasso it jumps to the second example, is this intended?

simplescreenrecorder-2021-12-09_13 44 17

jest.config.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
   moduleNameMapper: {
     '\\.(jpg|jpeg|png|gif|webp|svg|ttf|woff|woff2)$':
       '<rootDir>/__tests__/fileMock.js',
+    '\\.css$': '<rootDir>/__tests__/styleMock.js',
```
TomasSlama, line 25: unit test stopped working because of using `import quill.css`

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const useDisableEditor = (
+  editorRef: EditorRefType,
+  { disabled }: { disabled: Props['disabled'] }
+) => {
+  useEffect(() => {
+    if (editorRef.current) {
+      editorRef.current.enable(!disabled)
+    }
+  }, [disabled, editorRef])
```
dmaklygin1, line 14: if I'm not wrong, the `editorRef` won't be changed, it's the initial instance of the ref. But `editorRef.current` might be changed

LashaJini (approved):

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect } from 'react'
+
+import type { EditorRefType } from '../types'
+import { Props } from '../TextEditor'
+
+const useDisableEditor = (
+  editorRef: EditorRefType,
+  { disabled }: { disabled: Props['disabled'] }
```
dmaklygin1, line 8: I'm thinking...I don't know whether it's a good experience to borrow interface from the parent's level. 
Here, we definitely know that `disabled` is boolean, shouldn't we directly describe it here? Wdyt?

packages/picasso-lab/src/TextEditor/TextEditor.tsx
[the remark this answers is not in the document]
TomasSlama: hm, `HTML` will be our only supported output.

Input should be any copy pasted text or written text. Programaticly you can set value with `HTML` (when you paste the output back to editor)

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
TomasSlama: actually the ref will never trigger the rerender for us, so it is there only to remove eslint warning. It would work the same with or without `editorRef`. 
[Interesting article on this issue](https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/)

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
TomasSlama: Teimur found better solution to solve this issue. We convert HTML back to quill delta.

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
teimurjan: Whenever you change `.current` [the pointer to the ref is also updated](https://codesandbox.io/s/hardcore-chaum-8ozth?file=/src/App.js).

TomasSlama:
> When I visit texteditor component in picasso it jumps to the second example, is this intended?
> 
> simplescreenrecorder-2021-12-09_13 44 17

Good catch @LashaJini, it should work now properly

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeEvent.tsx
```diff
@@ -0,0 +1,28 @@
...
+const useHandleChangeEvent = (
+  editorRef: EditorRefType,
+  { onChange }: { onChange: Props['onChange'] }
+) => {
+  useEffect(() => {
+    const { current: editor } = editorRef
+
+    if (editor) {
+      const handler: TextChangeHandler = () => {
+        onChange(editor.root.innerHTML)
```
dmaklygin1, line 16: seems, `editor` might be null here... in a very rare case

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
teimurjan: My fault. When you set `.current`, it doesn't update the pointer. Thank you @dmaklygin1

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * TextEditor works as controlled input, when someone changes props.value
+ * we need to also change value in editor itself
+ */
+const useHandleChangeFromController = (
+  editorRef: EditorRefType,
+  { value: controllerValue }: { value: Props['value'] }
+) => {
+  useEffect(() => {
+    if (editorRef?.current && typeof controllerValue === 'string') {
```
dmaklygin1, line 15:
```
typeof controllerValue === 'string'
```
just thinking, how it's possible to pass different from `string` type here (`value` type is strictly defined as a string)

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
```diff
@@ -0,0 +1,34 @@
...
+        clipboard,
+        root: { innerHTML: editorValue }
+      } = editorRef.current
+
+      // is value set from outside
+      if (editorValue !== controllerValue) {
+        // @types/quill has incorrect convert typing
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
+        const delta = clipboard.convert(controllerValue)
```
dmaklygin1, line 26: What if controllerValue is longer than `maxlength` prop? 
Should we cut an extra text? 
I found a solution: https://github.com/quilljs/quill/issues/3009#issuecomment-618271044

packages/picasso-lab/src/TextEditor/TextEditor.tsx
```diff
@@ -0,0 +1,95 @@
...
+  maxlength?: number
+  /**
+   * The minimum number of characters required that the user should enter.
+   */
+  // TODO implement
+  minlength?: number
+  /**
+   * Callback on text change
+   */
+  onChange: (value: HTMLString) => void
```
dmaklygin1, line 41: I wanted to talk about the definition of `onChange`: I would move out the definition of this callback, so the end developer may reuse it outside by effortlessly defining a callback, for ex:
```
export type TextEditorChangeHandler = (value: HTMLString) => void
```
and the implementation
```
const handleChange: TextEditorChangeHandler = () => ...
```


wdyt?

packages/picasso-lab/src/TextEditor/story/Default.example.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from 'react'
+import { TextEditor } from '@toptal/picasso-lab'
+import { Container } from '@toptal/picasso'
+
+const Example = () => {
+  const [value, setValue] = useState<string | undefined>()
+
+  const handleChange = (newValue?: string) => setValue(newValue)
```
dmaklygin1, line 8: according to `onChange` type definition, newValue may not be `undefined`.
To avoid inconsistency, I would create a separate type for the onChange (like I highlighted above: TextEditorChangeHandler) and reuse it here.

packages/picasso-lab/src/TextEditor/story/Limit.example.tsx
```diff
@@ -0,0 +1,36 @@
...
+const Example = () => {
+  const [value, setValue] = useState<string | undefined>()
+
+  const handleChange = (newValue?: string) => setValue(newValue)
+
+  return (
+    <>
+      <Container style={{ height: '300px' }}>
+        <TextEditor
+          id='limit'
```
dmaklygin1, line 14: what if we have two Components with the same ID. Whether they will work in integration?

dmaklygin1 (commented):
Good work! Left some comments

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeEvent.tsx
TomasSlama: we have condition on line 14

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
TomasSlama: you are right, removed unnecessary condition.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
TomasSlama: yeah, we should, `maxlength` isn't implemented yet, I wasn't sure if we will need it and left it open for discussion, I think I will create another task for it.

packages/picasso-lab/src/TextEditor/TextEditor.tsx
TomasSlama: love it, thanks for hint ❤️

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
OndrejTuma: I'm wondering why second param is an object here 🤔 
I don't see any other purpose for this hook other than setting true/false to quill `enable` method
(similar to `useHandleAutofocus`)
Do you have some other plan to these hooks?

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
```diff
@@ -0,0 +1,34 @@
...
+ */
+const useHandleChangeFromController = (
+  editorRef: EditorRefType,
+  {
+    value: controllerValue,
+    oldValue
+  }: { value: Props['value']; oldValue?: string }
+) => {
+  useEffect(() => {
+    if (editorRef?.current) {
```
OndrejTuma, line 19: Like @dmaklygin1 suggested, editorRef is mandatory here too, so no need for optional chaining here

OndrejTuma: nit: here also may be more readable if you first deal with undefined quill like so:
`if (!editorRef.current) { return }`

packages/picasso-lab/src/TextEditor/hooks/useHandleAutofocus.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { Props } from '../TextEditor'
+
+const useHandleAutofocus = (
+  editorRef: EditorRefType,
+  { autofocus }: { autofocus: Props['autofocus'] }
+) => {
+  const firstRender = useRef(true)
+
+  useEffect(() => {
+    const editorApi = editorRef?.current
```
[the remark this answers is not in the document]
TomasSlama: you are right, removed.

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
TomasSlama: ✅

packages/picasso-lab/src/TextEditor/story/Default.example.tsx
TomasSlama: ✅

packages/picasso-lab/src/TextEditor/story/Limit.example.tsx
TomasSlama: The `id` must be unique, otherwise it would just reinit the previous one.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeEvent.tsx
```diff
@@ -0,0 +1,37 @@
...
+    setOldValue
+  }: {
+    onChange: Props['onChange']
+    setOldValue: React.Dispatch<React.SetStateAction<string | undefined>>
+  }
+) => {
+  useEffect(() => {
+    const { current: editor } = editorRef
+
+    if (editor) {
```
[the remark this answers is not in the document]
TomasSlama: I don't know, I try to avoid negative conditions

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
TomasSlama: ✅

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
```diff
@@ -0,0 +1,34 @@
...
+      const editor = editorRef.current
+
+      // is value set from outside
+      if (editor && oldValue !== controllerValue) {
+        const delta = (editor.clipboard.convert as (html: string) => Delta)(
```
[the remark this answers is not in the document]
TomasSlama: no need, because we have in top-level condition check if we have `controllerValue`

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeEvent.tsx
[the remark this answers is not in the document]
OndrejTuma: It's a nit, but I was promoting early return pattern which I like :-) feel free to resolve it

package.json
```diff
@@ -98,6 +98,7 @@
     "happo-plugin-storybook": "^2.7.0",
     "happo.io": "^6.8.0",
     "hygen": "^6.0.4",
+    "identity-obj-proxy": "^3.0.0",
```
teimurjan, line 101: Why did you add it?

teimurjan (commented):
[The problem](https://github.com/toptal/picasso/pull/2312/files#r765614496) @dmaklygin1 raised is not solved yet.

If we pass `editorRef.current` to all the hooks, we'll

1. Remove duplicated lines like `const editorApi = editorRef.current`
2. Solve the issue when the ref is changed but `useEffect` didn't run.

packages/picasso-lab/src/TextEditor/hooks/useDisableEditor.tsx
TomasSlama: image

guys, @teimurjan, @dmaklygin1, it doesn't matter if I pass `ref` or `ref.current` to `useEffect` in dependency list, changes in ref will never trigger rerender. I pass `ref` there, just to remove eslint warning. It is common issue, when we pass ref to custom hook. If it would not be in custom hook, eslint would not throw the warning. Detail explanation is in [this article](Interesting article on this issue).

And TBH I don't want to pass only mutable part of ref to hooks, I think it is much better experience to pass whole `ref` around as we do in normal components.

package.json
TomasSlama: good point, it is leftover from some test. 

✅ removed.

packages/picasso-lab/src/TextEditor/hooks/useHandleChangeFromController.tsx
```diff
@@ -0,0 +1,46 @@
...
+const useHandleChangeFromController = (
+  editorRef: EditorRefType,
+  { value: controllerValue }: { value: Props['value'] }
+) => {
+  useEffect(() => {
+    if (editorRef?.current && controllerValue) {
+      const editor = editorRef.current
+      const currentValue = editor.root.innerHTML
+
+      const isValueChanged = currentValue !== controllerValue
```
dmaklygin1, line 21: I guess, this is an odd empty line :)

dmaklygin1 (approved):

teimurjan (approved):

OndrejTuma (approved):

separatio (approved):
Manually tested. Looks good to go. :+1: 

Nit: http://localhost:9001/?path=/story/picasso-forms-form--form
The editor from forms does not have a placeholder text. I wonder if the validation works correctly while having a placeholder text as well.

Otherwise, as I said, :+1:

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+    const defaultValueInHtml = useMemo(
+      // this effects needs to happen only once on first render
+      () =>
+        defaultValue
+          ? hastUtilToHtml(hastSanitize(defaultValue))
+          : defaultValue,
+      /* eslint-disable-next-line */
+      []
+    )
```
augustobmoura, line 153: Instead of using eslint ignore comments we could refactor this to use an `useState` hook, it accepts a function as first argument that is only run when the hook is created. We can ignore the setter

```suggestion
    const [defaultValueInHtml] = useState(
      // this effects needs to happen only once on first render
      () =>
        defaultValue
          ? hastUtilToHtml(hastSanitize(defaultValue))
          : defaultValue
    )
```

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,237 @@
...
+        tabIndex={-1}
+        style={style}
+        ref={node => {
+          if (typeof ref === 'function') {
+            ref(node)
+          } else if (ref != null) {
+            ref.current = node
+          }
+          // @ts-ignore
+          wrapperRef.current = node
```
augustobmoura, line 182: We shouldn't use `@ts-ignore` **ever**, there are always better ways to do it. If the problem is that `wrapperRef.current` is readonly you could initialize `wrapperRef` as `useRef<HTMLDivElement | null>()` instead

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,237 @@
...
+            headerSelect: testIds?.headerSelect,
+            boldButton: testIds?.boldButton,
+            italicButton: testIds?.italicButton,
+            unorderedListButton: testIds?.unorderedListButton,
+            orderedListButton: testIds?.orderedListButton
+          }}
+        />
+        <QuillEditor
+          ref={editorRef}
+          disabled={disabled!} // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
augustobmoura, line 208: We could just cast this variable to boolean and remove the non-null-assertion

```suggestion
          disabled={!!disabled}
```

packages/picasso/src/RichTextEditor/utils/convertFormat.ts
```diff
@@ -0,0 +1,38 @@
...
+  HeaderValue as EditorHeaderValue
+} from '../../QuillEditor'
+import { FormatType as ToolbarFormatType, HeaderValue } from '../store/toolbar'
+
+export const convertBoldFromEditorValue = (bold: EditorBoldValue) =>
+  bold || false
+export const convertItalicFromEditorValue = (italic: EditorItalicValue) =>
+  italic || false
+export const convertListFromEditorValue = (list: EditorListValue) =>
+  list || false
```
augustobmoura, line 15: What's the point of these functions?

tsconfig.json
```diff
@@ -34,6 +34,6 @@
       "@testing-library/jest-dom"
     ]
   },
-  "include": ["packages/*/src"],
+  "include": ["packages/*/src", "packages/picasso/RichTextEditor"],
```
augustobmoura, line 37: I don't think this line is needed anymore

augustobmoura (changes requested):

cypress/component/RichTextEditor.spec.tsx
```diff
@@ -0,0 +1,365 @@
...
+  it('renders default editor', () => {
+    mount(renderEditor(defaultProps))
+
+    cy.get('body').happoScreenshot()
+  })
+
+  it('handles keybindings correctly', () => {
+    // render the editor
+    mount(renderEditor(defaultProps))
+    setAliases()
```
ozgurkececioglu, line 92: Couldn't we use beforeEach for these initialization parts?

cypress/component/RichTextEditor.spec.tsx
```diff
@@ -0,0 +1,365 @@
...
+      cy.get('@olButton').should('not.have.attr', 'disabled')
+
+      setSelectValue(cy.get('@headerSelect'), 'normal')
+      cy.get('@boldButton').should('not.have.attr', 'disabled')
+      cy.get('@italicButton').should('not.have.attr', 'disabled')
+      cy.get('@ulButton').should('not.have.attr', 'disabled')
+      cy.get('@olButton').should('not.have.attr', 'disabled')
+    })
+
+    it.only('does not open selectbox on click when toolbar is disabled', () => {
```
ozgurkececioglu, line 347: Is `only` needed still?

packages/picasso-forms/src/Form/story/index.jsx
```diff
@@ -218,6 +218,10 @@ however, you may need custom validators for more complex types of fields.
     title: 'No scrolling case',
     description: "Showcase Form's behavior on form submission error."
   }) // picasso-skip-visuals
+  .addExample('Form/story/RichTextEditor.example.tsx', {
+    title: 'Rich text editor',
+    description: 'Showcase how to display field requirements.'
```
ozgurkececioglu, line 223: Copy/paste error?

packages/picasso/src/QuillEditor/hooks/useQuillInstance/test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    default: QuillMock
+  }
+})
+
+describe('useQuillInstance', () => {
+  it('returns quill instance in ref', () => {
+    const id = 'editor'
+
+    // quill needs to have container for both toolbar and editor
+    // to successfuly initit
```
ozgurkececioglu, line 30:
```suggestion
    // to successfuly init
```

ozgurkececioglu:
> When I visit texteditor component in picasso it jumps to the second example, is this intended?
> 
> simplescreenrecorder-2021-12-09_13 44 17

As Lasha pointed out before `autoFocus` is active for both RichtextEditor pages and Form stories. These pages are directly jumping to RichtextEditor which is not a good idea for storybook.

__tests__/styleMock.js
```diff
@@ -0,0 +1 @@
+module.exports = {}
```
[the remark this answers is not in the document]
TomasSlama: we use it in `jest.config`. It was necessary, we were importing CSS file directly. We don't anymore, so we don't need it anymore, but it might still be useful in future 🤔, when someone will need to use CSS file. 🤷

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
OndrejTuma: Agree, only input value should not be function in this case 👍

packages/picasso/src/RichTextEditor/utils/convertFormat.ts
TomasSlama: @augustobmoura Quill internally shows only active format, so when we get format from Quill.getFormat() it is `{}` by default. When you have bold active `{ bold: true }`, when toggle, back to `{}`. So we transform here `undefined` to `false`

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: ✅

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: ✅

cypress/component/RichTextEditor.spec.tsx
TomasSlama: ✅

packages/picasso-forms/src/Form/story/index.jsx
TomasSlama: ✅

packages/picasso/src/QuillEditor/hooks/useQuillInstance/test.tsx
TomasSlama: ✅

packages/picasso/src/RichTextEditor/hooks/useCounter/test.ts
```diff
@@ -0,0 +1,97 @@
...
+    const { counterMessage } = result.current
+
+    expect(counterMessage).toBe('custom message')
+  })
+
+  it('returns custom maxLength message when maxLengthMessage is provided', () => {
+    const hookOptions = {
+      maxLength: 8,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
+      maxLengthMessage: (_maxLength: number, _currLength: number) =>
```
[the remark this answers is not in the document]
TomasSlama: ✅

tsconfig.json
TomasSlama: ✅

.changeset/forty-paws-deliver.md
```diff
@@ -0,0 +1,10 @@
+---
+'@toptal/picasso': minor
+'@toptal/picasso-forms': minor
```
OndrejTuma, line 3: Only minor? 🙂

TomasSlama:
@LashaJini @OndrejTuma @ozgurkececioglu I have fixed the unexpected autoFocus of the second example. The issue was in `useDefaultValue` triggering `quill.setContents` with `api` event type. It seems that `api` event automatically focuses on the Editor. So I changed it to `user` and everything seems to work as expected. Thanks for pointing this issue out! 🙇

OndrejTuma:
> @LashaJini @OndrejTuma @ozgurkececioglu I have fixed the unexpected autoFocus of the second example. The issue was in `useDefaultValue` triggering `quill.setContents` with `api` event type. It seems that `api` event automatically focuses on the Editor. So I changed it to `user` and everything seems to work as expected. Thanks for pointing this issue out! 🙇

Its still happening to me 🙁 on latest [https://picasso.toptal.net/FX-2293-text-editor/](https://picasso.toptal.net/FX-2293-text-editor/?a14c771552e45f385643bbc1f63d24c99236c5f6)

ozgurkececioglu:
> > @LashaJini @OndrejTuma @ozgurkececioglu I have fixed the unexpected autoFocus of the second example. The issue was in `useDefaultValue` triggering `quill.setContents` with `api` event type. It seems that `api` event automatically focuses on the Editor. So I changed it to `user` and everything seems to work as expected. Thanks for pointing this issue out! 🙇
> 
> Its still happening to me 🙁 on latest [https://picasso.toptal.net/FX-2293-text-editor/](https://picasso.toptal.net/FX-2293-text-editor/?a14c771552e45f385643bbc1f63d24c99236c5f6)

It's fixed for Form section but it is still happening for me too for this page: https://picasso.toptal.net/FX-2293-text-editor/?path=/story/components-richtexteditor--richtexteditor

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,230 @@
...
+      editorRef,
+      toolbarRef,
+      wrapperRef,
+      onFocus,
+      onBlur,
+      dispatch
+    })
+
+    const [defaultValueInHtml] = useState(
+      defaultValue ? hastUtilToHtml(hastSanitize(defaultValue)) : defaultValue
```
augustobmoura, line 146: Because we are not wrapping the default in a function `hastUtilToHtml` is being run every re-render. React already supports the first argument as a function and only run it on creation
```suggestion
      () => defaultValue ? hastUtilToHtml(hastSanitize(defaultValue)) : defaultValue
```

augustobmoura (changes requested):

__tests__/styleMock.js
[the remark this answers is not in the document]
augustobmoura: Ok, I think we can leave it for now

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: nice, didn't know it.

packages/picasso/src/RichTextEditor/utils/convertFormat.ts
augustobmoura: Yeah, but they seem kinda pointless, their whole implementation could be replaced by `!!`, doesn't look like something that needs to be a special function, you could just inline it where is being called. I think we can postpone their removal to future ticket, your call

augustobmoura (approved):

augustobmoura:
LGMT! :rocket:

OndrejTuma (approved):
Everything works 🎉

PR #2315, opened 2021-12-08
feat(Modal): [FX-2345] Extend Modal component with `transitionProps` prop support

ascrazy:
[FX-2345]

### Description

This adds correct `transitionProps` prop support for Modal and Accordion components.

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2345]: https://toptal-core.atlassian.net/browse/FX-2345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/Modal/Modal.tsx
+++ b/packages/picasso/src/Modal/Modal.tsx
@@ -8,6 +8,7 @@ import React, {
 import { makeStyles, Theme } from '@material-ui/core/styles'
 import Dialog from '@material-ui/core/Dialog'
 import { PaperProps } from '@material-ui/core/Paper'
+import { TransitionProps } from '@material-ui/core/transitions'
 import cx from 'classnames'
 import { StandardProps, SizeType } from '@toptal/picasso-shared'
 import { usePicassoRoot, useBreakpoint } from '@toptal/picasso-provider'
@@ -43,6 +44,8 @@ export interface Props extends StandardProps, HTMLAttributes<HTMLDivElement> {
   hideBackdrop?: boolean
   /** Position of the modal relative to the browser's viewport */
   align?: Alignment
+  /** Animation lifecycle callbacks. Backed by [react-transition-group/Transition](https://reactcommunity.org/react-transition-group/transition#Transition-props) */
+  transitionProps?: TransitionProps
   transitionDuration?: number
   paperProps?: PaperProps
   testIds?: {
@@ -120,6 +123,7 @@ export const Modal = forwardRef<HTMLElement, Props>(function Modal(props, ref) {
     paperProps,
     align = 'centered',
     testIds,
+    transitionProps,
     ...rest
   } = props
   const classes = useStyles(props)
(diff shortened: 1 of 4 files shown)


packages/picasso/src/Modal/Modal.tsx
```diff
@@ -43,6 +44,8 @@ export interface Props extends StandardProps, HTMLAttributes<HTMLDivElement> {
   hideBackdrop?: boolean
   /** Position of the modal relative to the browser's viewport */
   align?: Alignment
+  /** Animation lifecycle callbacks. Backed by [react-transition-group/Transition](https://reactcommunity.org/react-transition-group/transition#Transition-props) */
+  TransitionProps?: TransitionProps
```
dmaklygin1, line 48: the only one doubt, about the capital letter of the name.

OleksandrNechai: @ascrazy we have different design philosophies. Picasso is not a MUI wrapper. It is a full-blown UI kit that just happens to be implemented ATM with MUI. We might use a different approach in the future.

And to be able to do it, we set up our own interface. For these reasons we should `transitionProps` and also we need to define our own type, not re-export TransitionProps

OleksandrNechai (changes requested):
Please, see the comment below :point_down:

OleksandrNechai:
BTW, this PR still has no changeset generated

packages/picasso/src/Modal/Modal.tsx
ascrazy: Well,

Hi @OleksandrNechai,

Ok, I'll rename the prop, though as I already mentioned that means consumers will effectively be able to use both `transitionProps` and `TransitionProps`. IMO we might need to think of a generic approach that would block that possibility.

ascrazy: Regarding defining own interface for `transitionProps`, would like to discuss this a bit more.

As I already mentioned this is actually backed by `react-transition-group/Transition` and `material-ui` basically re-exports `TransitionProps` interface from `react-transition-group` package. I'd prefer to re-export it directly form `react-transition-group` rather then from `material-ui`, but we do not have react-transition-group in direct deps of the package, so it's probably not a good idea to directly reference a dependency that comes as a transitive dependency of another package. 

On the other hand I do not quite get the purpose of defining own `TransitionProps` interface. It would make sense If we are to have own abstraction for animation transitions, but then IMO we might need to develop this abstraction first, define clear principles of how they are handled, which callbacks are supported and so on. 

In case we are to rely on `react-transition-group` as a standard for transitions (which is a good option from my understanding as this one seem to be kind of a de-factor standard for transitions in react world), then introducing own interface in between will just add unnecessary friction.

ascrazy:
- [ ] Add changeset

packages/picasso/src/Modal/Modal.tsx
ascrazy: - [ ] Rename `TransitionProps -> transitionProps`
- [ ] Define own type for TransitionProps

OleksandrNechai (approved):

PR #2349, opened 2021-12-24
chore: greet contributor

OleksandrNechai:
Extending our contributor GHA. This time, let's greet contributor and let them know what to expect next. Examples of messages see below in PR :point_down:


.github/workflows/notify-jira-about-contribution.yml
```diff
@@ -1,21 +1,31 @@
...
   pull_request:
-    types: [opened]
+    # types: [opened]
     branches:
       - master
 
 jobs:
   notify-jira:
     # if external contribution
-    if: ${{!contains(fromJson('["denieler", "TomasSlama", "OleksandrNechai", "OndrejTuma", "dmaklygin1", "LashaJini", "toptal-build"]'), github.event.pull_request.user.login)}}
```
LashaJini, line 13: I guess we will be using this line of code a lot of times 😁. Is it possible to define it somewhere and have it as a single source of truth?

OleksandrNechai: I think it is possible. Do you want to figure out how? My building time on the calendar comes to the end :)

LashaJini: Sure, why not.

OleksandrNechai: Feel free to do it. But only after TextEditor styles are done :)

.github/workflows/handle-contribution.yml
```diff
@@ -0,0 +1,40 @@
+name: Handle contribution
+
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened]
+    branches:
+      - master
```
LashaJini, line 8: I was wondering, what will happen when someone opens PR on another branch, they might still need this greeting in that case also. wdyt? :thinking:

LashaJini: No, they won't!

LashaJini (approved):

dmaklygin1 (approved):

PR #2380, opened 2022-01-12
chore(picasso): add support for returning undefined value from `getKey` selector of Autocomplete

sanex3339:
[SPT-2294]

### Description

Please do not merge this PR until i test it in Staff-Portal

#### Motivation
In staff portal as part of this task https://toptal-core.atlassian.net/browse/SPT-2294 we have to update all FE code for updated `node` type that becomes nullable in related BE PR (https://github.com/toptal/platform/pull/54372).

We have many places such as:
```
const getTagId = (tagEdge: Item) =>
  (tagEdge as TaskTagEdgeFragment).node.id

<Autocomplete getKey={getTagId} />
```

Where `(tagEdge as TaskTagEdgeFragment).node` is nullable value, so now to correctly pass this selector to Autocomplete component we have to do the following:
```
const getTagId = (tagEdge: Item) =>
  (tagEdge as TaskTagEdgeFragment).node?.id ?? ''

<Autocomplete getKey={getTagId} />
```
This potentially may cause some unexpected behavior.

To prevent it, support for returning undefined value from `getKey` selector was added to Autocomplete
Now if we return `undefined` value from `getKey` selector, such option wont be rendered at all, so we can keep the following code on our side:
```
const getTagId = (tagEdge: Item) =>
  (tagEdge as TaskTagEdgeFragment).node?.id

<Autocomplete getKey={getTagId} />
```

### How to test

1. Open https://picasso.toptal.net/SPT-2294-autocomplete-allow-undefined-return-from-get-key/?path=/story/forms-autocomplete--autocomplete
2. Scroll down to the `Custom options rendering` example
3. Open autocomplete popup.
4. It should have only 4 option items

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[SPT-2294]: https://toptal-core.atlassian.net/browse/SPT-2294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/breezy-waves-fold.md
+++ b/.changeset/breezy-waves-fold.md
@@ -1,0 +1,7 @@
+---
+'@toptal/picasso': minor
+---
+
+### Autocomplete
+
+- add support for returning undefined value from `getKey` selector for Autocomplete
--- a/packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
+++ b/packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
@@ -9,24 +9,27 @@ const renderOtherOption = (value: string) => (
   </Typography>
 )
 
-interface Country extends Item {
-  country: string
-  capital: string
-  code: string
-}
+type Country =
+  | (Item & {
+      country: string
+      capital: string
+      code?: string
+    })
+  | null
 
 const EMPTY_INPUT_VALUE = ''
 
 const allOptions: Country[] = [
   { country: 'Belarus', capital: 'Minsk', code: 'BE' },
-  { country: 'Croatia', capital: 'Zagreb', code: 'HR' },
+  null,
+  { country: 'Croatia', capital: 'Zagreb' },
   { country: 'Lithuania', capital: 'Vilnius', code: 'LU' },
   { country: 'Slovakia', capital: 'Bratislava', code: 'SK' },
   { country: 'Ukraine', capital: 'Kyiv', code: 'UA' }
 ]
 
 const getDisplayValue = (item: Item | null) =>
-  (item && (item as Country).country) || ''
+  (item && (item as Country)?.country) ?? ''
 
 const filterOptions = (str: string) =>
   str !== ''
@@ -43,22 +46,24 @@ const CustomOptionRenderer = () => {
         showOtherOption
         value={value}
         placeholder='Start typing country...'
-        options={options}
-        getKey={(item: Item) => (item as Country).code}
-        renderOption={(option: Partial<Country>, index) => (
-          <Container>
-            <Typography size='medium' weight='semibold'>
-              {option.country}
-            </Typography>
-            <Typography size='inherit' style={{ fontSize: '12px' }}>
-              {option.capital} ({index})
-            </Typography>
-          </Container>
-        )}
+        options={options as Item[]}
+        getKey={item => (item as Country)?.code}
+        renderOption={(option: Partial<Country>, index) =>
+          option && (
+            <Container>
+              <Typography size='medium' weight='semibold'>
+                {option.country}
+              </Typography>
+              <Typography size='inherit' style={{ fontSize: '12px' }}>
+                {option.capital} ({index})
+              </Typography>
+            </Container>
+          )
+        }
         renderOtherOption={renderOtherOption}
         onSelect={(option: Partial<Country>) => {
           window.console.log('onSelect returns item object:', option)
-          window.console.log('selected capital:', option.capital)
+          window.console.log('selected capital:', option?.capital)
 
           const itemValue = getDisplayValue(option)
 
--- a/packages/picasso/src/Autocomplete/test.tsx
+++ b/packages/picasso/src/Autocomplete/test.tsx
@@ -2,6 +2,7 @@
 /* eslint-disable max-lines-per-function */
 import React from 'react'
 import { render, fireEvent, PicassoConfig } from '@toptal/picasso/test-utils'
+import { Item } from '@toptal/picasso/Autocomplete/types'
 import { OmitInternalProps } from '@toptal/picasso-shared'
 import { generateRandomString } from '@toptal/picasso-provider'
 import * as titleCaseModule from 'ap-style-title-case'
@@ -262,6 +263,40 @@ describe('Autocomplete', () => {
 
       expect(renderOption).toHaveBeenCalledTimes(10)
       expect(getDisplayValue).not.toHaveBeenCalled()
+    })
+
+    describe('when cannot found custom option key', () => {
+      const partiallyInvalidTestOptions = [
+        { text: 'Belarus', value: 'BY' },
+        null,
+        { text: 'Croatia' },
+        { text: 'Lithuania', value: 'LU' }
+      ]
+
+      it('does not render custom option', () => {
+        const getDisplayValue = jest.fn()
+        const renderOption = jest.fn(item => (
+          <div data-testid='custom-option'>{item.value}</div>
+        ))
+        const getKey = jest.fn(option => option?.value)
+
+        const { getByTestId, getAllByTestId } = renderAutocomplete({
+          options: partiallyInvalidTestOptions as Item[],
+          value: '',
+          renderOption,
+          getDisplayValue,
+          getKey
+        })
+
+        const input = getByTestId('autocomplete')
+
+        fireEvent.click(input)
+
+        expect(renderOption).toHaveBeenCalledTimes(4)
+        expect(getAllByTestId('custom-option')).toHaveLength(2)
+        expect(getDisplayValue).not.toHaveBeenCalled()
+        expect(getByTestId(testIds.scrollMenu)).toMatchSnapshot()
+      })
     })
 
     it('should not transform options text to title case when Picasso titleCase property is true', () => {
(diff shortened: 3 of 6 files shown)


packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
```diff
@@ -43,8 +44,8 @@ const CustomOptionRenderer = () => {
         showOtherOption
         value={value}
         placeholder='Start typing country...'
-        options={options}
-        getKey={(item: Item) => (item as Country).code}
+        options={options as Country[]}
+        getKey={item => (item as typeof allOptions[0])?.code}
```
pudek357, line 48: Can you explain the type `typeof allOptions[0]` instead of `Country`?

packages/picasso/src/Autocomplete/test.tsx
```diff
@@ -264,6 +265,41 @@ describe('Autocomplete', () => {
...
 
+    describe('when cannot found custom option key', () => {
+      const partiallyInvalidTestOptions = [
+        { text: 'Belarus', value: 'BY' },
+        null,
+        { text: 'Croatia' },
+        { text: 'Lithuania', value: 'LU' }
+      ]
+
+      it('does not renders custom option', () => {
```
pudek357, line 276:
```suggestion
      it('does not render custom option', () => {
```

packages/picasso/src/Autocomplete/test.tsx
```diff
@@ -264,6 +265,41 @@ describe('Autocomplete', () => {
...
+          renderOption,
+          getDisplayValue,
+          getKey
+        })
+
+        const input = getByTestId('autocomplete')
+
+        fireEvent.click(input)
+
+        // TODO: figure out why renderOption is called twice
```
pudek357, line 295: I think one TODO is enough in the file regarding `renderOption`

.changeset/breezy-waves-fold.md
```diff
@@ -0,0 +1,7 @@
+---
+'@toptal/picasso': minor
+---
+
+## Autocomplete
+
+Add support for returning undefined value from `getKey` selector of Autocomplete
```
denieler, line 7: please use https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md for changesets 🙇‍♂️

packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
sanex3339: Nice catch, i updated Country type
https://github.com/toptal/picasso/pull/2380/commits/e74eb9226d3a1f29961c44acbbde89dae300f65b

packages/picasso/src/Autocomplete/test.tsx
sanex3339: https://github.com/toptal/picasso/pull/2380/commits/e74eb9226d3a1f29961c44acbbde89dae300f65b

packages/picasso/src/Autocomplete/test.tsx
sanex3339: https://github.com/toptal/picasso/pull/2380/commits/e74eb9226d3a1f29961c44acbbde89dae300f65b

pudek357 (approved):

packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
```diff
@@ -9,24 +9,27 @@ const renderOtherOption = (value: string) => (
...
+    })
+  | null
 
 const EMPTY_INPUT_VALUE = ''
 
 const allOptions: Country[] = [
   { country: 'Belarus', capital: 'Minsk', code: 'BE' },
-  { country: 'Croatia', capital: 'Zagreb', code: 'HR' },
+  null,
+  { country: 'Croatia', capital: 'Zagreb' },
```
denieler, line 25: why do you think having such an API for the `Autocomplete` component is a good idea?

sanex3339: I described our case in the PR description.
We have nullable values in `getKey` selectors so we have to return an empty string right now that is not good because these values are used as component keys

.changeset/breezy-waves-fold.md
sanex3339: Done

sanex3339:

packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
denieler: @sanex3339 I understand, but if you would think from the component perspective, would you like for this component on isolation to support such options?

sanex3339: Ideally - no, of course. But such API simplifies the handling of such cases.
Even in terms of performance.

Of course if we have such nullable types and values we can filter out all nullable options before passing to the `Autocomplete` (and also create a new type that has only defined fields), so

#### With existing API:
Pros:
- better API

Cons:
- iteration over options just for filtering
- this requires adding many additional code

#### With changes in PR:
Pros:
- no additional iteration over options, it is done inside Autocomplete component with existing `map`
- much simpler to handle such cases, less code

Cons:
- API contains `| undefined` type

denieler: @sanex3339 yeah, I would treat Picasso component as an individual component, not coupled with Staff Portal only. And I would not move issues happening in Staff Portal to Picasso. Because other teams who use Picasso might not struggle with that, but will struggle with the complexity of the API of components

dmaklygin1: I completely agree with @denieler

sanex3339: Hmm. 
Can we revert types to the previous (without `| undefined`) but keep
```
const key = getKey(option)

if (!key) { // or if (key === '')
  return null
}
```
to not break anything when key is empty string?

TomasSlama (commented):
as @denieler  pointed out, it seems wrong to change the functionality of Select, because of BE API changed. Wouldn't be sufficient to filter options before passing to select? Maybe create a proxy component inside SP with this functionality, so you don't have to repeat yourself.

.changeset/breezy-waves-fold.md
denieler: @sanex3339 it looks like the new changeset still don't follow the example from https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md

packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
sanex3339: It won't be useless because type signature allows us to pass strings and in our case, if we want to return at least something in case of nullable value, the only way is to return an empty string. This is allowed by types but causes unexpected behavior, because these keys are used as option components keys.

About pre-filtered options.
It is possible(except that this is an additional iteration over options).
The problem is mostly not in options that we pass to `options` property, but `getKey` function, because we receive `Item` type in it and we have to cast it to some specific entity type, for example, `TagAutocompleteOption`, but this type has nullable values. Of course we can create additional type like `TagAutocompleteOptionsWithNonNullableFooField` but isn't it looks redundantly? 
I mean if we have the ability to pass custom objects to the Autocomplete component and `getKey` property that allows us to map the structure of custom object to internal `Item` object, why we can't support the basic case when such field may be nullable?

sanex3339:
About pre-filtered options.
It is possible(except that this is an additional iteration over options).
The problem is mostly not in options that we pass to options property, but getKey function, because we receive Item type in it and we have to cast it to some specific entity type, for example, TagAutocompleteOption, but this type has nullable values. Of course we can create additional type like TagAutocompleteOptionsWithNonNullableFooField but isn't it looks redundantly?
I mean if we have the ability to pass custom objects to the Autocomplete component and getKey property that allows us to map the structure of custom object to internal Item object, why we can't support the basic case when such field may be nullable?

packages/picasso/src/Autocomplete/story/CustomOptionRenderer.example.tsx
denieler: @sanex3339 you can use a type guard to receive a new type after `.filter`, I've prepared an example for you - https://www.typescriptlang.org/play?#code/C4TwDgpgBA8mwEsD2A7KBeKBvAUFfUAbgIYA2ArhAFxQrkC2ARhAE4A0eBA1hCDQM7AWCFAHMcAXxw4AxqkFQk8ZCn40AFHESooAH1rlSpAJQBtALoYopzvixEylGgEY2UHnygByZ14kcCAyMAgnsSCmooACY3DxovKL8cc2kAM3IUGW00BH4ANTIEABMtFXUlbJpSnX06I2MaCpUoXNhlHVxAlghgchY0Jp0AQnRMOtJJaTlVYFoIAHdq1StB1QA6VIRSYFZ1XILSYqXjIA

PR #2415, opened 2022-02-03
chore: refactor picasso lab codemod

LashaJini:
[FX-NNNN]

### Description

- Fix duplication of imports in picasso-lab removal codemod

### How to test

### Screenshots

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


denieler (commented):
What about using this eslint rule instead? https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md It has `--fix` option

LashaJini:
> What about using this eslint rule instead? https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md It has `--fix` option

@denieler I think it still won't solve the problem we are facing:

> If you like to split up imports across lines or may need to import a default and a namespace, you may not want to enable this rule.

Because we definitely have `default` imports (I haven't seen `namespace` imports but still an easy fix), and there are other edge cases, for example comment related. But it still good to have that rule in davinci.

denieler:
> > What about using this eslint rule instead? https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md It has `--fix` option
> 
> @denieler I think it still won't solve the problem we are facing:
> 
> > If you like to split up imports across lines or may need to import a default and a namespace, you may not want to enable this rule.
> 
> Because we definitely have `default` imports (I haven't seen `namespace` imports but still an easy fix), and there are other edge cases, for example comment related. But it still good to have that rule in davinci.

but adding this rule would solve like 80-90% of cases with almost no effort 🤷‍♂️  and no maintenance cost 🤷‍♂️ 
I see you already have this codemod, so effort already has been done, but still please consider this proposition

packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/basic.input.tsx
```diff
@@ -1,15 +1,22 @@
...
-import { Container } from '@toptal/picasso-lab'
-import { Calendar } from '@toptal/picasso-lab/Calendar'
+import { _Existing1, _Existing2 } from '@toptal/picasso'
+import _Component1 from '@toptal/picasso-lab/Component1'
+import { _Component2 } from '@toptal/picasso-lab'
+import _Button1, { _Button1Type } from '@toptal/picasso-lab/Button1'
+import _Component4, { _Component4Type } from '@toptal/picasso-lab/Component4'
+import { _Calendar } from '@toptal/picasso-lab/Calendar/even/deeper'
+import { default as _Component5Default } from '@toptal/picasso-lab/Component5'
+import { Component6 as _CustomName1 } from '@toptal/picasso-lab/Component6'
```
denieler, line 11: these are invalid imports for Picasso components, we should not support them. Later with new ESM they will be prohibited by design

packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/basic.input.tsx
```diff
@@ -1,15 +1,22 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+/* eslint-disable import/no-named-default */
 // @ts-nocheck
-import { Container } from '@toptal/picasso-lab'
-import { Calendar } from '@toptal/picasso-lab/Calendar'
+import { _Existing1, _Existing2 } from '@toptal/picasso'
+import _Component1 from '@toptal/picasso-lab/Component1'
```
denieler, line 5: same here

packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/basic.input.tsx
```diff
@@ -1,15 +1,22 @@
...
+import { _Component2 } from '@toptal/picasso-lab'
+import _Button1, { _Button1Type } from '@toptal/picasso-lab/Button1'
+import _Component4, { _Component4Type } from '@toptal/picasso-lab/Component4'
+import { _Calendar } from '@toptal/picasso-lab/Calendar/even/deeper'
+import { default as _Component5Default } from '@toptal/picasso-lab/Component5'
+import { Component6 as _CustomName1 } from '@toptal/picasso-lab/Component6'
+import * as _Picasso from '@toptal/picasso-lab'
+import { _NOPE1 } from '@toptal/nope'
+import { _NOPE2 } from '@toptal/picasso-shared'
+import { _NOPE3 } from '@toptal/picasso-shared/deeper'
```
denieler, line 15: neither this import is valid for Picasso component

LashaJini:
> > > What about using this eslint rule instead? https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md It has `--fix` option
> > 
> > 
> > @denieler I think it still won't solve the problem we are facing:
> > > If you like to split up imports across lines or may need to import a default and a namespace, you may not want to enable this rule.
> > 
> > 
> > Because we definitely have `default` imports (I haven't seen `namespace` imports but still an easy fix), and there are other edge cases, for example comment related. But it still good to have that rule in davinci.
> 
> but adding this rule would solve like 80-90% of cases with almost no effort man_shrugging and no maintenance cost man_shrugging I see you already have this codemod, so effort already has been done, but still please consider this proposition

No, no, I agree with you 100%. I think, this eslint rule is a must-have rule in davinci, it's just in this case (as the duplication part is almost done) it doesn't solve whatever edge cases are left. If you'd like after this PR, I will add that rule in davinci.

packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/basic.input.tsx
denieler: max one folder can be used - `@toptal/picasso/utils`

packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/basic.input.tsx
LashaJini: Sorry could you be specific, please?

denieler: we don't allow deep imports, except `@toptal/picasso/Icon` and `@toptal/picasso/utils`

packages/picasso-codemod/src/utils/find-components.ts
```diff
@@ -1,8 +1,8 @@
-import { Core, JSCodeshift } from 'jscodeshift'
+import { RootType, JSCodeshift } from './types'
 
 const findComponents = (
   componentName: string,
-  root: ReturnType<Core>,
+  root: RootType,
```
denieler, line 5: why this has to be changed?

LashaJini: Not to repeat it.

LashaJini: I think we are not making a good use of `utils` directory overall.

.changeset/kind-dryers-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+'@toptal/picasso-codemod': minor
+---
+
+- Refactor `picasso-lab.ts` codemod so it covers most of the edge cases when
+  replacing `picasso-lab` with `picasso`
```
denieler, line 6:
```suggestion
- Fix duplication of imports in picasso-lab removal codemod
```

LashaJini: Much better, thank you.

.eslintignore
```diff
@@ -1,3 +1,5 @@
 dist-package
 build
 node_modules
+
+packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/*.tsx
```
denieler, line 5: I don't think we need to remove fixture from eslint. We might want to disable single rule for such fixtures, but we can do that via inline comments or in `.eslintrc` config file

packages/picasso-codemod/src/utils/find-components.ts
denieler: in this case you create another level of abstraction which causes more problems than solves. So to understand which type root has I need to go into `./types` file. Also, I don't see directly that `root` somehow related to `jscodeshift` internal implementation, because I don't have import from `jscodeshift` in this file

.eslintignore
LashaJini: I will check what I can do.

LashaJini: I couldn't make it work with inline comments.

packages/picasso-codemod/src/utils/is-specified-import.ts
```diff
@@ -0,0 +1,23 @@
+import { ImportDeclaration } from 'jscodeshift'
+
+/**
+ * Usage:
+ *
+ *  j(file.source)
+ *    .find(j.ImportDeclaration, (node) => isSpecifiedImport(node, 'module/name'))
+ */
+const isSpecifiedImport = (
```
denieler, line 9: I didn't understand what is this `specified` import 🤔  maybe a better word would help explain better or maybe additional comment to how to use this function 🙇‍♂️

.eslintignore
denieler: please share the rule you want to disable, so I can try to help with fixing with inline comment

packages/picasso-codemod/src/v18.0.0/picasso-lab/__testfixtures__/basic.input.tsx
```diff
@@ -1,15 +1,35 @@
-// @ts-nocheck
-import { Container } from '@toptal/picasso-lab'
-import { Calendar } from '@toptal/picasso-lab/Calendar'
+import { _MUI } from 'mui'
+/* @ts-rule */
+import { _useState } from 'react'
+import { _ExistingComponent1, _ExistingComponent2 } from '@toptal/picasso'
+
+/* Only these cases can have deeper imports according to picasso guidelines */
```
denieler, line 6: it seems in this file we have 2 types of comments:
- to explain some imports
- to emulate comments in real code

I would keep only one type only

LashaJini: Good point. I guess I will keep the second option only.

.eslintignore
LashaJini: Screenshot from 2022-02-09 13-18-54

`@typescript-eslint/no-unused-vars`, `import/order`, `import/no-duplicates`, `import/no-named-default`

packages/picasso-codemod/src/utils/is-specified-import.ts
LashaJini: Better naming would be better, could you provide some, please? Basically, this function just filters out imports, it will only give us specified import name, passed by user as an `moduleSpecifier` argument.

LashaJini: I tried to follow `jscodeshift` API naming, so we won't have (too) different naming.

TomasSlama (approved):
I don't see any other problem than what Daniel suggested. GJ @LashaJini

OndrejTuma (approved):
Nice work 🙂

packages/picasso-codemod/src/utils/is-specified-import.ts
LashaJini: @TomasSlama suggested:
> what about something like isImportByPath?
Which look much better than current version, wdyt @denieler ?

packages/picasso-codemod/src/utils/find-components.ts
LashaJini: But if we won't abstract things then we will have duplications, which is even worse. And we have to always check what type actually is hiding behind abstractions whenever we want more details. For example, `import { ImportDeclaration } from jscodeshift'`, I still forget what these type actually has behind curtains, and I still have to check source code, however the naming gives me a good hint that it's related to import declaration and I can just use code completion to quickly understand what it offers. Same thing is in `root`'s case, `root` has `RootType` which is like a "crisp abstraction", it is concrete and abstracted at the same time. 

This is not the issue with `root` only, as the number of code increases in the future, we would want to abstract away types/utils.

TomasSlama: @LashaJini but you are not creating any new type, you are just abstracting an already created type

LashaJini: Yeah, I guess you are right guys. I will revert it.

packages/picasso-codemod/src/utils/has-top-level-comment.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  root
+    .find(j.Declaration)
+    .at(0)
+    .forEach(path => {
+      const comments = path.value.comments
+
+      if (
+        comments &&
+        comments.length > 0 &&
```
separatio, line 14: Would `comments?.length > 0` have worked here as well, instead of `comments &&`?

LashaJini: We need to do a lot of things to calm ts compiler down :smiley:
Screenshot from 2022-02-17 11-54-08
Screenshot from 2022-02-17 11-54-25

separatio: I understand. Sorry to hear that. 😅

denieler (approved):

PR #2476, opened 2022-02-28
feat(FieldRequirements): add new component

ozgurkececioglu:
[FX-2525]

### Description

- FieldRequirements: added a new component
- forms/PasswordInput: now supporting FieldRequirements as default but it can be hidden with `hideRequirements` (ex: confirm password field)
- I had to touch `FormField` and `FieldWrapper` so please share your considerations for the future

### How to test

- PasswordInput: https://picasso.toptal.net/fx-2525-add-strength-validations-section-to-passwordinput/?path=/story/forms-passwordinput--passwordinput
- FieldRequirements: https://picasso.toptal.net/fx-2525-add-strength-validations-section-to-passwordinput/?path=/story/components-fieldrequirements--fieldrequirements
- Form version(bottom of the page): https://picasso.toptal.net/fx-2525-add-strength-validations-section-to-passwordinput/?path=/story/picasso-forms-form--form

### Screenshots

Screen Shot 2022-03-01 at 10 26 32
Screen Shot 2022-03-01 at 10 26 46
Screen Shot 2022-03-01 at 10 27 02

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2525]: https://toptal-core.atlassian.net/browse/FX-2525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/dirty-points-applaud.md
+++ b/.changeset/dirty-points-applaud.md
@@ -5,7 +5,7 @@
 ---
 ### FieldWrapper
 
-- add new prop to render field requirements if exist
+- add new prop to render passed field requirements
 
 ### PasswordInput
 
--- a/.changeset/dirty-points-applaud.md
+++ b/.changeset/dirty-points-applaud.md
@@ -1,0 +1,13 @@
+---
+'@toptal/picasso-forms': minor
+---
+
+---
+### FieldWrapper
+
+- add new prop to render field requirements if exist
+
+### PasswordInput
+
+- update the component to render field requirements
+
--- a/packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
+++ b/packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
@@ -51,6 +51,10 @@ export type Props<
     hideLabelRequiredDecoration?: boolean
     fieldType?: string
     children: (props: any) => React.ReactNode
+    renderFieldRequirements?: (props: {
+      value?: ValueType
+      error?: boolean
+    }) => React.ReactNode
   }
 
 type FieldMeta<T> = FieldMetaState<T> & {
@@ -155,6 +159,7 @@ const FieldWrapper = <
     enableReset,
     onResetClick,
     'data-testid': dataTestId,
+    renderFieldRequirements,
     // FieldProps - https://final-form.org/docs/react-final-form/types/FieldProps
     afterSubmit,
     allowNull,
@@ -267,7 +272,15 @@ const FieldWrapper = <
   )
 
   return (
-    <PicassoForm.Field error={error} hint={hint} data-testid={dataTestId}>
+    <PicassoForm.Field
+      error={error}
+      hint={hint}
+      data-testid={dataTestId}
+      fieldRequirements={renderFieldRequirements?.({
+        value: input.value,
+        error: error
+      })}
+    >
       {!hideFieldLabel && label && (
         <PicassoForm.Label
           requiredDecoration={requiredDecoration}
--- a/packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
+++ b/packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
@@ -52,7 +52,7 @@ export type Props<
     fieldType?: string
     children: (props: any) => React.ReactNode
     renderFieldRequirements?: (props: {
-      value?: ValueType
+      value?: TInputValue
       error?: boolean
     }) => React.ReactNode
   }
--- a/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
+++ b/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
@@ -15,6 +15,8 @@ export type Props = PasswordInputProps &
 
 const { composeValidators } = validators
 
+const TIMEOUT = 500
+
 export const PasswordInput = ({
   validate,
   hideRequirements,
@@ -51,7 +53,7 @@ export const PasswordInput = ({
       open={showContent}
       error={error}
       description='Please make sure that your password contains:'
-      timeout={500}
+      timeout={TIMEOUT}
       requirements={[
         {
           message: 'At least 8 characters',
@@ -95,7 +97,10 @@ export const PasswordInput = ({
         const handleBlur = (event: FocusEvent<HTMLInputElement>) => {
           inputProps.onBlur?.(event)
 
-          setShowContent(false)
+          // Hide the requirements after a short delay
+          setTimeout(() => {
+            setShowContent(false)
+          }, TIMEOUT)
         }
 
         return (
--- a/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
+++ b/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
@@ -1,7 +1,8 @@
 import React, { useState, FocusEvent, ReactNode } from 'react'
 import {
   PasswordInput as PicassoPasswordInput,
-  PasswordInputProps
+  PasswordInputProps,
+  FieldRequirementValueType
 } from '@toptal/picasso'
 import { FieldValidator } from 'final-form'
 
@@ -15,7 +16,24 @@ export type Props = PasswordInputProps &
 
 const { composeValidators } = validators
 
-const TIMEOUT = 500
+const ANIMATION_TIMEOUT = 500 // same as in Picasso/FieldRequirements
+
+const validatePassword: FieldValidator<PasswordInputProps['value']> = (
+  value?: string
+) => {
+  if (!value) {
+    return undefined
+  }
+
+  const isValidPassword =
+    passwordValidators.atLeastEightCharacters(value) &&
+    passwordValidators.atLeastOneNumber(value) &&
+    passwordValidators.atLeastOneUpperCaseCharacter(value) &&
+    passwordValidators.atLeastOneLowerCaseCharacter(value) &&
+    passwordValidators.atLeastOneSpecialCharacter(value)
+
+  return isValidPassword ? undefined : 'Invalid password'
+}
 
 export const PasswordInput = ({
   validate,
@@ -24,36 +42,23 @@ export const PasswordInput = ({
 }: Props) => {
   const [showContent, setShowContent] = useState(false)
 
-  const validatePassword: FieldValidator<PasswordInputProps['value']> = (
-    value?: string
-  ) => {
-    if (!value || hideRequirements) {
-      return undefined
-    }
-
-    const validationResult =
-      passwordValidators.atLeastEightCharacters(value) &&
-      passwordValidators.atLeastOneNumber(value) &&
-      passwordValidators.atLeastOneUpperCaseCharacter(value) &&
-      passwordValidators.atLeastOneLowerCaseCharacter(value) &&
-      passwordValidators.atLeastOneSpecialCharacter(value)
-
-    return validationResult ? undefined : 'Invalid password'
-  }
+  const validationsObject = hideRequirements
+    ? validate
+    : composeValidators([validatePassword, validate])
 
   const renderFieldRequirements: ({
     value,
     error
   }: {
-    value?: any
+    value?: FieldRequirementValueType
     error?: boolean
   }) => ReactNode = ({ value, error }) => (
-    <Form.FieldRequirements
-      value={value}
+    <Form.FieldRequirements<string>
+      value={(value ?? '') as string}
       open={showContent}
       error={error}
       description='Please make sure that your password contains:'
-      timeout={TIMEOUT}
+      timeout={ANIMATION_TIMEOUT}
       requirements={[
         {
           message: 'At least 8 characters',
@@ -82,7 +87,7 @@ export const PasswordInput = ({
   return (
     <FieldWrapper<PasswordInputProps>
       {...rest}
-      validate={composeValidators([validatePassword, validate])}
+      validate={validationsObject}
       renderFieldRequirements={
         !hideRequirements ? renderFieldRequirements : undefined
       }
@@ -100,7 +105,7 @@ export const PasswordInput = ({
           // Hide the requirements after a short delay
           setTimeout(() => {
             setShowContent(false)
-          }, TIMEOUT)
+          }, ANIMATION_TIMEOUT)
         }
 
         return (
--- a/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
+++ b/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
@@ -1,18 +1,110 @@
-import React from 'react'
+import React, { useState, FocusEvent, ReactNode } from 'react'
 import {
   PasswordInput as PicassoPasswordInput,
   PasswordInputProps
 } from '@toptal/picasso'
+import { FieldValidator } from 'final-form'
 
+import { validators } from '../utils'
 import FieldWrapper, { FieldProps } from '../FieldWrapper'
+import passwordValidators from './validators'
+import Form from '../Form'
 
-export type Props = PasswordInputProps & FieldProps<PasswordInputProps['value']>
+export type Props = PasswordInputProps &
+  FieldProps<PasswordInputProps['value']> & { hideRequirements?: boolean }
 
-export const PasswordInput = (props: Props) => {
+const { composeValidators } = validators
+
+export const PasswordInput = ({
+  validate,
+  hideRequirements,
+  ...rest
+}: Props) => {
+  const [showContent, setShowContent] = useState(false)
+
+  const validatePassword: FieldValidator<PasswordInputProps['value']> = (
+    value?: string
+  ) => {
+    if (!value || hideRequirements) {
+      return undefined
+    }
+
+    const validationResult =
+      passwordValidators.atLeastEightCharacters(value) &&
+      passwordValidators.atLeastOneNumber(value) &&
+      passwordValidators.atLeastOneUpperCaseCharacter(value) &&
+      passwordValidators.atLeastOneLowerCaseCharacter(value) &&
+      passwordValidators.atLeastOneSpecialCharacter(value)
+
+    return validationResult ? undefined : 'Invalid password'
+  }
+
+  const renderFieldRequirements: ({
+    value,
+    error
+  }: {
+    value?: any
+    error?: boolean
+  }) => ReactNode = ({ value, error }) => (
+    <Form.FieldRequirements
+      value={value}
+      open={showContent}
+      error={error}
+      description='Please make sure that your password contains:'
+      timeout={500}
+      requirements={[
+        {
+          message: 'At least 8 characters',
+          validator: passwordValidators.atLeastEightCharacters
+        },
+        {
+          message: '1 number',
+          validator: passwordValidators.atLeastOneNumber
+        },
+        {
+          message: '1 uppercase character',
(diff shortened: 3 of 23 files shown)


packages/picasso/src/FormField/FormField.tsx
```diff
@@ -37,6 +47,7 @@ export const FormField = forwardRef<HTMLDivElement, Props>(function FormField(
       {children}
       {error && <FormError className={classes.error}>{error}</FormError>}
       {hint && <FormHint className={classes.hint}>{hint}</FormHint>}
+      {fieldRequirements}
```
ozgurkececioglu, line 50: To achieve placing the field-requirements below our generic field error, I had to do this.

packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
```diff
@@ -51,6 +51,10 @@ export type Props<
     hideLabelRequiredDecoration?: boolean
     fieldType?: string
     children: (props: any) => React.ReactNode
+    renderFieldRequirements?: (props: {
```
ozgurkececioglu, line 54: This has to be like that to get the latest value and error state from form-context.

packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
```diff
@@ -1,18 +1,110 @@
...
+          message: '1 special character',
+          validator: passwordValidators.atLeastOneSpecialCharacter
+        },
+        {
+          message: '1 lowercase character',
+          validator: passwordValidators.atLeastOneLowerCaseCharacter
+        }
+      ]}
+    />
+  )
```
dmaklygin, line 78: :+1:

packages/picasso/src/FieldRequirements/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface FieldRequirement {
+  message: string
+  validator: (value: any) => boolean
```
dmaklygin, line 3: I'm thinking about possible types of `value`. Maybe we could be more specific about possible types?

packages/picasso/src/FieldRequirements/story/Default.example.tsx
```diff
@@ -0,0 +1,40 @@
...
+import type { FieldRequirement } from '@toptal/picasso'
+
+const requirements: FieldRequirement[] = [
+  {
+    message: 'Min 5 characters',
+    validator: value => value.length >= 5
+  },
+  {
+    message: 'Max 10 characters',
+    validator: value => value.length < 10
```
dmaklygin, line 12: shouldn't it be: `value.length <= 10` ?

dmaklygin (approved):
Great job!

packages/picasso/src/FieldRequirements/story/Default.example.tsx
ozgurkececioglu: Copilot mistake 🙂 I'll fix it.

packages/picasso/src/FieldRequirements/types.ts
ozgurkececioglu: There is a perfect match [ValueType](https://github.com//toptal/picasso/blob/f3da7ae3a8bc3aad1ea67ff8ad81ce2de45aa989/packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx#L23) but it is not exported. And most other places also use `any`. I would prefer exporting ValueType and using it around the repo. WDYT?

vvmarulin:
@ozgurkececioglu 

Checked the demo. Everything works great except for one small thing. When I tried to click submit button after I entered the password (focus on password field), the button wasn't clicked because there is an animation. That is a UX problem because it's unclear from the user's perspective that the button hasn't been clicked.

https://user-images.githubusercontent.com/3861498/156132937-b1a2cf48-8fb1-4972-b148-51e83ff4df66.mp4

You can see on the video, that after the first click on the submit button nothing happened. And only after the second click a validation error for the second field was shown.

Maybe we need some delay for animation so the button doesn't move immediately? WDYT?

packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx
```diff
@@ -0,0 +1,59 @@
...
+          color='inherit'
+          data-testid={`${dataTestId}-valid-icon`}
+        />
+      ) : error ? (
+        <CloseMinor16 color='red' data-testid={`${dataTestId}-error-icon`} />
+      ) : (
+        <Bullet16
+          color='dark-grey'
+          data-testid={`${dataTestId}-default-icon`}
+        />
```
OndrejTuma, line 46: I'm wondering in which case this can happen. Because to me it seems that either valid or error state are possible. I can only think of some validation that is somewhat informative only (does not reach a valid state) but is not considered in the `validate` function of FieldWrapper. And that seems like a hell of an edgecase 🙂

vvmarulin: When you focus on the password field for the first time it will show a list with bullet icons. Also while you typing a new value for the first time the requirements will either have a bullet icon or check icon.

ozgurkececioglu: @OndrejTuma according to our designs here: https://www.figma.com/file/q2nvjiyO2CLqBv4DeJnU3i/Product-Library-Documentation?node-id=702%3A11760 there is a default state that we are showing bullets only as @vvmarulin explained perfectly.

packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
```diff
@@ -1,18 +1,115 @@
...
+          inputProps.onFocus?.(event)
+
+          setShowContent(true)
+        }
+
+        const handleBlur = (event: FocusEvent<HTMLInputElement>) => {
+          inputProps.onBlur?.(event)
+
+          // Hide the requirements after a short delay
+          setTimeout(() => {
```
ozgurkececioglu, line 101: According to our decision, there should be a delay to start hiding animation for the requirements after onBlur. `transitionDelay` didn't work so I had to provide this solution.

Another possibility is that I can implement my own transition animation with pure CSS, rather than using MUI's Collapse component.

packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx
OndrejTuma: Oh, I missed that 🙇  thanks for explanation

OndrejTuma (approved):
GJ 👍

vvmarulin (approved):

.changeset/dirty-points-applaud.md
```diff
@@ -0,0 +1,13 @@
+---
+'@toptal/picasso-forms': minor
+---
+
+---
+### FieldWrapper
+
+- add new prop to render field requirements if exist
```
separatio, line 8: I'd rephrase this to: `- add new prop to render passed field requirements`

packages/picasso/src/FieldRequirements/FieldRequirements.tsx
```diff
@@ -0,0 +1,98 @@
...
+          ))}
+        </Grid>
+      </Collapse>
+    )
+  }
+)
+
+FieldRequirements.defaultProps = {
+  open: false,
+  timeout: 500
```
TomasSlama, line 93: is this related to the `TIMEOUT` const in the PasswordInput file? maybe we can reuse it?

packages/picasso/src/FieldRequirements/styles.ts
```diff
@@ -0,0 +1,20 @@
...
+export default ({ palette, spacing, sizes: { input } }: Theme) =>
+  createStyles({
+    root: {
+      display: 'flex',
+      width: input.width
+    },
+    description: {
+      marginTop: '0.4rem'
+    },
+    fieldRequirementItem: {
```
TomasSlama, line 12: maybe I'm blind, but I didn't find usage of this class?

packages/picasso-forms/src/PasswordInput/validators.ts
```diff
@@ -0,0 +1,28 @@
...
+  return /[!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?]/.test(value)
+}
+
+export default {
+  atLeastEightCharacters,
+  atLeastOneUpperCaseCharacter,
+  atLeastOneLowerCaseCharacter,
+  atLeastOneNumber,
+  atLeastOneSpecialCharacter
+}
```
TomasSlama, line 28: Should we unit test those validators?

packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
```diff
@@ -1,18 +1,115 @@
...
+  const [showContent, setShowContent] = useState(false)
+
+  const validatePassword: FieldValidator<PasswordInputProps['value']> = (
+    value?: string
+  ) => {
+    if (!value || hideRequirements) {
+      return undefined
+    }
+
+    const validationResult =
```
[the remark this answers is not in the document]
ozgurkececioglu: I was planning something different but forgot to update this after changing my mind. Good catch!

augustobmoura (changes requested):

packages/picasso/src/FieldRequirements/styles.ts
ozgurkececioglu: It is being used inside `FieldRequirementItem`. see [here](https://github.com//toptal/picasso/blob/c157f45afe2907210b0ba8840c52f4e950c468c8/packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx#L39)

packages/picasso/src/FieldRequirements/types.ts
ozgurkececioglu: I created an identical clone type into FieldRequirements and started using it as generic base type 👍

packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+      ) : status === 'error' ? (
+        <CloseMinor16
+          color={colorMap[status]}
+          data-testid={`${dataTestId}-error-icon`}
+        />
+      ) : (
+        <Bullet16
+          color={colorMap[status]}
+          data-testid={`${dataTestId}-default-icon`}
+        />
```
TomasSlama, line 56:
```suggestion
      const Icons = {
        error: CloseMinor16,
        // ..
      }
      const IconComponent = Icons[status]
```

and also I think we should not hard code test id `data-testid={`${dataTestId}-valid-icon`}`. So I would suggest `testIds.validIcon`

packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { PropsWithChildren } from 'react'
+import { makeStyles, Theme } from '@material-ui/core/styles'
+
+import Bullet16 from '../Icon/Bullet16'
+import Typography from '../Typography'
+import Grid from '../Grid'
+import styles from './styles'
+import { CheckMinor16, CloseMinor16 } from '../Icon'
```
TomasSlama, line 8:
```suggestion
import Typography from '../Typography'
import Grid from '../Grid'
import styles from './styles'
import { CheckMinor16, CloseMinor16, Bullet16 } from '../Icon'
```

packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  return (
+    <Grid.Item
+      small={6}
+      className={classes.fieldRequirementItem}
+      data-testid={dataTestId}
+    >
+      <IconComponent
+        color={colorMap[status]}
+        data-testid={`${dataTestId}-${status}-icon`}
```
TomasSlama, line 52: we should not have any hardcoded data test id in Picasso. `testIds: { successIcon: string, ...}`

packages/picasso/src/FieldRequirements/FieldRequirementItem.tsx
ozgurkececioglu: I'm so use to clean these imports with some shortcuts but here, we are not using I guess 🙂 thanks 👍

packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
```diff
@@ -51,6 +51,10 @@ export type Props<
...
+      value?: ValueType
```
augustobmoura, line 55: Shouldn't this type be `TInputValue`? This value would be the same as the input right? So in a number input a number, in a password a string, etc

packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
```diff
@@ -51,6 +51,10 @@ export type Props<
...
+      error?: boolean
+    }) => React.ReactNode
```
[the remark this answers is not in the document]
augustobmoura: I don't know if I understood it correctly. I'm not seeing the difference between a function returning a ReactNode and a common component (in fact, FunctionComponents have essentially the same signature as this prop).

Why would calling a function that returns a new ReactNode not trigger a react re-render, but directly creating a ReactNode would do it?

packages/picasso-forms/src/PasswordInput/PasswordInput.tsx
```diff
@@ -1,18 +1,120 @@
...
+
+  const renderFieldRequirements: ({
+    value,
+    error
+  }: {
+    value?: FieldRequirementValueType
+    error?: boolean
+  }) => ReactNode = ({ value, error }) => (
+    <Form.FieldRequirements<string>
+      value={(value ?? '') as string}
```
augustobmoura, line 57: Why not type `value` as a `string` instead of type asserting it? Passwords will always be a `string` (at least for now)

If the problem is that `FieldWrapper` on accepts an non-generic `value`, refer to my comment on `FieldWrapper.tsx` about using `TInputValue` as the type of value

packages/picasso/src/FieldRequirements/FieldRequirements.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+import styles from './styles'
+import FieldRequirementItem, {
+  FieldRequirementItemStatus
+} from './FieldRequirementItem'
+import Typography from '../Typography'
+import { FieldRequirement, ValueType } from './types'
+import Grid from '../Grid'
+
+export interface Props<TInputValue extends ValueType = ValueType>
```
augustobmoura, line 14: I don't think that this should have a default type, it's better to require the developer to specify it. Being too generic of type can lead to type assertion hacks or untyped code.

FieldWrapper for example, requires it to be defined by the developer https://github.com/toptal/picasso/blob/510c31766197cce7df7191cb8c27df4b967920e0/packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx#L36

packages/picasso/src/FieldRequirements/FieldRequirements.tsx
```diff
@@ -0,0 +1,108 @@
...
+    gridContainer?: string
+  }
+}
+
+const ANIMATION_TIMEOUT = 500
+
+const useStyles = makeStyles<Theme>(styles)
+
+export const FieldRequirements = function <
+  TInputValue extends ValueType = ValueType
```
augustobmoura, line 40: I don't think that this should have a default type, same argument as in the first comment on `FieldRequirements.tsx`

packages/picasso/src/FieldRequirements/index.ts
```diff
@@ -0,0 +1,10 @@
+import { OmitInternalProps } from '@toptal/picasso-shared'
+
+import { Props } from './FieldRequirements'
+import { ValueType } from './types'
+
+export { default } from './FieldRequirements'
+export * from './types'
+export type FieldRequirementsProps<
+  TInputType extends ValueType = ValueType
```
augustobmoura, line 9: I don't think that this should have a default type, same argument as in the first comment on `FieldRequirements.tsx`

packages/picasso/src/FieldRequirements/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  | number
+  | boolean
+  | null
+  | undefined
+  | FileUpload[]
+  | DateOrDateRangeType
+  | Item
+  | Item[]
+
+export interface FieldRequirement<TInputType extends ValueType = ValueType> {
```
augustobmoura, line 18: I don't think that this should have a default type, same argument as in the first comment on `FieldRequirements.tsx`

augustobmoura (changes requested):

packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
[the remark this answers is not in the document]
ozgurkececioglu: Your logic is totally fine. I like it. I've just not been able to find a solution to pass the same component result to `FieldWrapper`, before passing it as `ReactNode` to `FormField`. I've created a variant branch to talk with more solid examples here: https://github.com/toptal/picasso/tree/fx-2525-add-strength-validations-section-to-passwordinput-variant

To achieve your suggestion, as you can see [here](https://github.com/toptal/picasso/blob/4f3d761e4296be3d817121333d6df2624a9d607c/packages/picasso-forms/src/PasswordInput/PasswordInput.tsx#L54), I had to add one more wrapper to pass `showContent` to my `FieldRequirementsComponent` to start and stop the animation. If you run storybook for this variant branch, you would see that `FieldRequirementsComponent` is being recreated after each focusing and bluring. Even if we accept that, because of the recreation, animations couldn't work.

With the render function solution, I've just postponed the return phase to `FieldWrapper's render` and it solved the issue.

If you assist on this variant branch with a working solution, I would definitely be happy and accept that 🙂

separatio:
Didn't finish manually testing, but please pay attention to min-maxing. I have exactly 10 characters written, and the field is red.
image

separatio:
Haven't checked designs yet, but it feels weird that the field is not red while it's clearly failing validation.
image

ozgurkececioglu:
> Haven't checked designs yet, but it feels weird that the field is not red while it's clearly failing validation. image

Again, it is about our input fields globally. I've just used a very simple one to accommodate the validations running. You can check others from: https://picasso.toptal.net/?path=/story/forms-input--input

I think in focus, none of our fields provide an error signal.

packages/picasso-forms/src/FieldWrapper/FieldWrapper.tsx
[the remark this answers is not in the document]
augustobmoura: Hmmm, I think I understand the problem now. `FieldRequirements` needs an intermediate value, so we create an anonymous component to wrap it up.

The difference between a simple function and a component, in this case, is that when using the JSX syntax the function identity is used on reconciliation. Because we are creating an anonymous component every time, the reconciliation fails every time.

Maybe there's a way to work around that, but I don't think the effort is worth it. We can keep it as a function then. Thanks for writing the example repo for illustration!

augustobmoura (approved):

PR #2499, opened 2022-03-09
feat(RichText): enable value in html string format

TomasSlama:
### Description

Users will need in some cases show preview before they send data to BE. In this case, we need to be able to take output directly from `RichTextEditor` (HTML) and show it.

I have explained in the documentation, that using HTML string should be used carefully since there is unnecessary parsing involved. AST from BE should be always prioritized if possible.

### How to test

- Checkout new example in storybook `With HTML value`

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -70,8 +69,12 @@ const picassoMapper = (child: ReactNode): ReactNode => {
   return createElement(type, { key: child.key }, mappedChildren)
 }
 
-const useRichText = (value: ASTType): ReactNodeArray | ReactNode => {
+const useRichText = (value: ASTType): ReactNode[] | ReactNode => {
```
TomasSlama, line 72: `ReactNodeArray` is now deprecated.

packages/picasso/src/utils/test.tsx
```diff
@@ -230,7 +231,7 @@ describe('useWidthOf', () => {
     }
     const element = {
       getBoundingClientRect: () => rect
-    }
+    } as ReferenceObject
```
TomasSlama, line 234: unrelated change, but it was throwing TS error.

packages/picasso/src/utils/html-to-hast.ts
```diff
@@ -0,0 +1,18 @@
+import hastSanitize from 'hast-util-sanitize'
+import hastFromDom from 'hast-util-from-dom'
+
+import { ElementType, ASTType } from '../RichText/types'
+
+const htmlToAst = (html: string) => {
```
ozgurkececioglu, line 6: Since the string will come from another source, do you think it is a good idea to provide error handling? Because if the parsing fails, the component may also fail to render. WDYT?

TomasSlama: @ozgurkececioglu I had the same idea and I actually had try catch block there. But it will never fail to parse if we pass a string, which we force with TS.

Check test case I have added for this function, where I pass invalid HTML string `foo`. It just returns `{ type: 'root', children: undefined }`

TomasSlama: image

dmaklygin: yeah, confirm:
```
new DOMParser().parseFromString(html, 'text/html')
```
doesn't throw an error. In some cases, like parsing XML, it may return a `parsererror`, but doesn throw an Error.
https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString#error_handling

packages/picasso/src/utils/html-to-hast.ts
```diff
@@ -0,0 +1,18 @@
...
+  const dom = new DOMParser().parseFromString(html, 'text/html')
+  const domHast = hastSanitize(hastFromDom(dom.body)) as ElementType
```
dmaklygin, line 8: concerning `hastSanitize`, I can see an option to specify a list of available tag names: https://github.com/syntax-tree/hast-util-sanitize#tagnames
wdyt to specify this white list?

@types/hast-util-from-dom/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+  type HastChild = Parent['children'][number]
+  type HastNode = HastChild | Root
+
+  type Options = {
+    afterTransform: Function
+  }
+  const fromDom: (node: Node, options?: Options) => HastNode
+
+  export default fromDom
+}
```
dmaklygin, line 13: 👍

dmaklygin: seems, there is no definitions for this package: https://github.com/DefinitelyTyped/DefinitelyTyped/find/master

packages/picasso/src/utils/html-to-hast.ts
TomasSlama: @dmaklygin great idea, I will also export this schema, to use it also in RichTextEditor

dmaklygin (approved):

ozgurkececioglu (approved):

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -70,8 +69,12 @@ const picassoMapper = (child: ReactNode): ReactNode => {
...
   const mappedTextNodes = useMemo(() => {
+    if (!value.children) {
```
OndrejTuma, line 74: From the `ASTType` definition, children is always array, not undefined. In that case this condition would never pass. But in `packages/picasso/src/utils/test.tsx` there is a case when children is undefined. So one of these cases should probably be updated?

OndrejTuma (approved):
Good job 💪

PR #2521, opened 2022-03-15
chore: add SSR support for react use layout effect hook

deniskaber:
[FX-2546]

### Description

This adds `useIsomorphicLayoutEffect` hook to support using Picasso components in SSR mode. 
Also adds an eslint rule to use `useIsomorphicLayoutEffect` over `useLayoutEffect`

### How to test

- By default everything should work as before.
- In case of usage Drawer component in SSR mode, it should not generate a warning.

### Review

- [-] Annotate all `props` in component with documentation
- [-] Create `examples` for component
- [-] Ensure that deployed demo has expected results and good examples

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2546]: https://toptal-core.atlassian.net/browse/FX-2546?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/shared/src/hooks/use-isomorphic-layout-effect.ts
```diff
@@ -0,0 +1,4 @@
+import { useLayoutEffect, useEffect } from 'react'
+
+export const useIsomorphicLayoutEffect =
+  typeof document !== 'undefined' ? useLayoutEffect : useEffect
```
TomasSlama, line 4: I'm wondering if it might be even useful for users of Picasso in their projects. Maybe we can reexport this hook from `@toptal/picasso/utils` what do you think?

Also we already have `isBrowser` util function in Picasso. Maybe we could move the implementation to shared so we are able to reuse the function here. And just reexport it from `picasso/utils`

What do you think?

packages/picasso-provider/src/Picasso/FontsLoader.tsx
```diff
@@ -1,4 +1,5 @@
-import { useLayoutEffect } from 'react'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import useIsomorphicLayoutEffect from '@toptal/picasso/utils/use-isomorphic-layout-effect'
```
TomasSlama, line 2: here in the Picasso provider package, we should use the implementation from picasso-shared, because this way we are introducing circular dependency. Picasso is depending on the provider and vice versa.

packages/picasso/src/utils/index.ts
```diff
@@ -46,6 +46,7 @@ export { default as unsafeErrorLog } from './unsafe-error-log'
 export { default as sum } from './sum'
 export { default as htmlToHast, hastSanitizeSchema } from './html-to-hast'
 export type { ReferenceObject } from './use-width-of'
+export { default as useIsomorphicLayoutEffect } from './use-isomorphic-layout-effect'
```
TomasSlama, line 49: I would move implementation back to shared and here we can just reexport it. We can do the same with `isBrowser` so we can use it even in shared.

packages/picasso-provider/src/Picasso/FontsLoader.tsx
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line no-restricted-imports
 import { useLayoutEffect } from 'react'
```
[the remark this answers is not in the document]
deniskaber: It was done because we restricted import of picasso-shared here - https://github.com/toptal/picasso/blob/master/.eslintrc.js#L5, but since we moved useIsomorphicLayoutEffect to picasso/utils - it's no longer the case, it's fixed

packages/shared/src/hooks/use-isomorphic-layout-effect.ts
deniskaber: Yep, that's actually a great idea! Thanks! I've moved the hook to utils and reused isBrowser helper.

packages/picasso-provider/src/Picasso/FontsLoader.tsx
denieler: interesting, our eslint rules or syncpack checks should fail on such imports, which are not mentioned in `package.json` file. We need to check them out.

denieler (changes requested):
we should not import in picasso-provider package anything from picasso package

packages/picasso/src/TreeView/useNodes.ts
```diff
@@ -123,7 +124,7 @@ export const useNodes = (
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [dynamicNodes, initialized, direction, verticalMargin, horizontalMargin])
 
-  useLayoutEffect(() => {
```
denieler, line 126: can this be converted to `useEffect` maybe instead?

packages/picasso/src/TreeView/PointNode/PointNode.tsx
```diff
@@ -26,7 +20,7 @@ export const PointNode = forwardRef<SVGGElement, Props>(
       return `translate(${node.x},${node.y})`
     }, [node.x, node.y])
 
-    useLayoutEffect(() => {
```
denieler, line 29: maybe convert it to `useEffect`?

packages/picasso/src/StaticTreeView/StaticTreeView.tsx
```diff
@@ -55,7 +56,7 @@ const StaticTreeView = (props: Props) => {
     leftMostNodeX: 0
   })
 
-  useLayoutEffect(() => {
```
denieler, line 58: same here, maybe to `useEffect`?

packages/picasso/src/DatePicker/DatePicker.tsx
```diff
@@ -185,18 +184,18 @@ export const DatePicker = (props: Props) => {
   // Keep the input value in sync with date value update
   // Updating on incoming date value or timezone change
   // Should not update when input is focused to prevent overriding it's value
-  useLayoutEffect(() => {
```
denieler, line 188: maybe also possible to convert to `useEffect`

packages/picasso-provider/src/Picasso/FontsLoader.tsx
deniskaber: The same works for picasso-shared - it also depends on picasso-provider and it can also cause circular dependency.

packages/picasso/src/TreeView/useNodes.ts
deniskaber: I suggest to do it in a separate ticket as it doesn't relates to the current one and it will require testing in the projects that use this component in order to be shure that applied changes does not break anything.

denieler: it does relate, because then you don't need to convert it to isomorphic effect

packages/picasso-provider/src/Picasso/FontsLoader.tsx
denieler: exactly... then probably if you need such hook in a provider - implement it in the picasso-provider package

deniskaber: I would suggest to keep using `useLayoutEffect` as this logic for font loading would require a bigger re-factoring to support SSR and we can move to isomorphic version when we'll refactor FontLoader to fully support SSR.

packages/picasso/src/TreeView/useNodes.ts
deniskaber: Do you know projects that use this component?

deniskaber: We now stick with the same `useLayoutEffect` hook when using component in browser. If we would change it to `useEffect` in browser we can experience undesirable glitches in projects using these components.

denieler: > If we would change it to useEffect in browser we can experience undesirable glitches in projects using these components.

I understand that, but usage of `useLayoutEffect` could be a mistake from the beginning. If it's possible - I would double-check that, because in this case we would have less tech debt if we can just replace it to `useEffect`

.eslintrc.js
```diff
@@ -14,7 +14,13 @@ const generateConfig = () =>
         rules: {
           'no-restricted-imports': [
             'error',
-            ...forbiddenPackageNames.map(name => `@toptal/${name}`)
+            ...forbiddenPackageNames.map(name => `@toptal/${name}`),
+            {
+              name: 'react',
+              importNames: ['useLayoutEffect'],
+              message:
+                '`useLayoutEffect` causes a warning in SSR. Use `useIsomorphicLayoutEffect`'
```
denieler, line 22: I would add also from where to use `useIsomorphicLayoutEffect`, like `Use useIsomorphicLayoutEffect from @toptal/picasso-shared`, or something like that

denieler (approved):

TomasSlama (approved):

OndrejTuma (approved):
Good job 👍

dmaklygin:
Hey :)
Actually, I didn't get one thing: why only part of `useLayoutEffect` usages was replaced by `useIsomorphicLayoutEffect`. I can see some of the components are not touched.

denieler:
@deniskaber @dmaklygin did you guys find those missing `useLayoutEffect`, which were not replaced?

PR #2522, opened 2022-03-16
feat(RichTextEditor): add picasso-forms wrapper

augustobmoura:
[FX-2511](https://toptal-core.atlassian.net/browse/FX-2511)

### Description

We are adding a new wrapper for the RichTextEditor in picasso-forms

### How to test

- Open the storybook for Form, there's an example for the RichTextEditor, check on submitting and the validations

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


augustobmoura:
@toptal-bot run all

augustobmoura:
@toptal-bot run all

augustobmoura:
@toptal-bot run deploy:documentation

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,19 @@
...
+  RichTextEditor as PicassoRichTextEditor,
+  RichTextEditorProps
+} from '@toptal/picasso'
+
+import FieldWrapper, { FieldProps } from '../FieldWrapper'
+
+export type Props = RichTextEditorProps & FieldProps<string>
+
+export const RichTextEditor = (props: Props) => (
+  <FieldWrapper<RichTextEditorProps & { value: string }> value='' {...props}>
```
ozgurkececioglu, line 12: In other field wrapper files, there are new type definitions to handle type operations. I would also follow that and do this intersection outside like `FormRichTextEditorProps` and pass it to `<FieldWrapper>`

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -91,13 +95,14 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
       defaultValue,
       disabled,
       id,
-      onChange,
+      onChange = noop,
```
ozgurkececioglu, line 98: After making `onChange` optional, I would prefer putting `onChange` into `defaultProps` and add `no-non-null-assertion` eslint rule to return as Tomas did for `disabled` prop. One day, while working on these `defaultProps`, we can find this one and refactor easily this way. WDYT?

ozgurkececioglu:
@augustobmoura @TomasSlama when I checked the implementation of RichTextEditor, I couldn't see any callback props about `onFocus` and `onBlur`. WDYT about a solution like this:
```
    const { handleFocus, handleBlur } = useOnFocus({
      editorRef,
      toolbarRef,
      wrapperRef,
      onFocus: useCallback(() => {
        setIsEditorFocused(true)
        onFocus?.() // notify final-form that this input has been focused
      }, [setIsEditorFocused]),
      onBlur: useCallback(() => {
        setIsEditorFocused(false)
        onBlur?.() // notify final-form that this input has been blurred
      }, [setIsEditorFocused]),
      dispatch
    })
```

ozgurkececioglu:
@TomasSlama when I type some values and delete them back, I can't see any `empty` state for the editor. I think it will bring doubts to the user. WDYT?

https://user-images.githubusercontent.com/7733167/158538527-e6d6a12d-0111-487d-a893-3e06dc51d7f4.mov

Suggestion: when I was working on another richtext editor wrapper, I was checking if any renderable text exists after each typing and calling `onChange` with an empty string if none of them left to notify form library that this editor has no value anymore. WDYT about my suggestion 🙂

TomasSlama:
@ozgurkececioglu good suggestion, I have created separated task for it https://toptal-core.atlassian.net/browse/FX-2549

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -158,7 +163,8 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
           classes.editorWrapper,
           {
             [classes.disabled]: disabled,
-            [classes.focused]: isEditorFocused
+            [classes.focused]: isEditorFocused,
+            [classes.invalid]: error
```
ozgurkececioglu, line 167: I would follow the same name as `classes.error` because `invalid` doesn't belong to MUI but `error` is commonly used.

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -32,6 +32,8 @@ export type CounterMessageSetter = (
   isError: boolean
 ) => string
 
+const noop = () => {}
```
TomasSlama, line 35:
```suggestion
import { noop } from '@toptal/picasso/utils'
```

TomasSlama (approved):
Oz left some cool suggestions, other than that, it looks cool.

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
ozgurkececioglu: Since we are still using `defaultProps` for rendering our props table on storybook, I think it will be beneficial to support it until we find out a new way to prepare our props table. 

Otherwise, I'm totally up for getting rid of it 👍

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: We could create a private type to handle `value`, but it shouldn't be exposed to the developer because we don't support it, and thinking more about it, we may need to create an intermediate state to keep track of `onChange`s, passing the value as an empty string always might have some weird interactions with final-form

augustobmoura:
> @augustobmoura @TomasSlama when I checked the implementation of RichTextEditor, I couldn't see any callback props about `onFocus` and `onBlur`.

That is true, I forgot about wiring `onFocus` and `onBlur`. I will do it right now

> @TomasSlama when I type some values and clear the input afterwards, I can't see any `empty` state for the editor. I think it will bring doubts to the user.

I noticed that problem and I would talk about it today with you guys. It looks like [an issue in quill](https://github.com/quilljs/quill/issues/1328), it should be easily to workaround on our side

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: Our Picasso Book shows default parameters on the props table as well:

image

We can have a talk with the team later about what should be the default

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: Makes sense, I will change it

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: Yeah, I agreed that we can separate into a different type def.

The problem about the intermediate state is a different thing though, I committed a code around it, the problem is that I was passing an constant empty string to final form as value, that could make some weird behavior between syncing state, the intermediate state should solve it

ozgurkececioglu (approved):
Thanks for the updates 👍

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: @augustobmoura it was discussed a few months back, and we agreed, that might be good to refactor. But it should probably be done in a separate ticket and do it across the whole codebase.

TomasSlama (approved):

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,39 @@
...
+} from '@toptal/picasso'
+
+import FieldWrapper, { FieldProps } from '../FieldWrapper'
+
+export type Props = RichTextEditorProps & FieldProps<string>
+
+type InternalProps = RichTextEditorProps & { value: string }
+
+export const RichTextEditor = ({ onChange, ...rest }: Props) => {
+  const [value, setValue] = useState('')
```
OndrejTuma, line 14: I tried but have not made `defaultValue` work. Did I miss something or should it be added? 🙏

separatio:
image

I can submit an empty field. That shouldn't be possible.
The way to reproduce this is:
1. Insert any text.
2. Delete all text.
3. Submit the form.

separatio (approved):
Alright. That empty string bug is fixed in https://github.com/toptal/picasso/pull/2527

Manual tests are ok.

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: It is bugged, I was trying to make it work since friday

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,42 @@
...
+    [onChange, setValue]
+  )
+
+  return (
+    <FieldWrapper<InternalProps>
+      value={value}
+      onChange={handleOnChange}
+      {...rest}
+    >
+      {(inputProps: RichTextEditorProps) => (
```
dmaklygin, line 35: I'm thinking, that... `inputProps` argument contains the `value`. doesn't it?

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -122,16 +138,23 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
       handleTextFormat,
       format: state.toolbar.format
     })
+
+    const wrappedOnFocus = useCallback(() => {
+      setIsEditorFocused(true)
+      onFocus?.()
```
dmaklygin, line 144: this question mark is not needed anymore

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -122,16 +138,23 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
...
+    }, [setIsEditorFocused, onFocus])
+
+    const wrappedOnBlur = useCallback(() => {
+      setIsEditorFocused(false)
+      onBlur?.()
```
dmaklygin, line 149: this question mark is not needed anymore

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
ozgurkececioglu: I think so. Final-form should keep all the values for us. There shouldn't be a need to store the value externally unless something weird happens about RichTextEditor 🤔

packages/picasso-forms/src/Form/story/RichTextEditor.example.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { Container } from '@toptal/picasso'
+import { Form } from '@toptal/picasso-forms'
+
+const RichTextEditorExample = () => (
+  <Container>
```
denieler, line 6: I think we can remove `Container` from here

packages/picasso-forms/src/Form/story/RichTextEditor.example.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <Form onSubmit={data => window.alert(JSON.stringify(data))}>
+      <Container bottom='small'>
+        <Form.RichTextEditor
+          required
+          defaultValue={{
+            type: 'root',
+            children: [{ type: 'text', value: 'Example of default text' }]
+          }}
```
denieler, line 14: this default value structure looks a bit complex, why can't we just make `defaultValue='Example of default text'`?

packages/picasso-forms/src/utils/types.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * This is an Omit implementation that works with indexed types (meaning it ignores index types altogether)
+ * For background see the issues:
+ * https://github.com/microsoft/TypeScript/issues/45367
+ * https://github.com/microsoft/TypeScript/issues/42436
+ * https://github.com/microsoft/TypeScript/issues/31153
+ */
+export type OmitKnown<T, K extends keyof NoIndex<T>> = Omit<NoIndex<T>, K> & OnlyIndex<T>
+
```
denieler, line 31: I'm really against such complex and unclear types ☝️ it's super hard to understand what they do as well as to read the code produced by those types

denieler: in my opinion even having not so much strong typing is better than having such types in the code

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -91,13 +104,16 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
       defaultValue,
       disabled,
       id,
-      onChange,
+      onChange = noop,
+      onFocus = noop,
+      onBlur = noop,
```
denieler, line 109: those default values should be in `defaultProps`

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
denieler: yep, agree with @TomasSlama let's stick to the code style and we can refactor all at once, but meanwhile, we will have code with the same style all over the codebase

packages/picasso/src/RichTextEditor/styles.ts
```diff
@@ -20,6 +20,12 @@ export default ({ palette, sizes }: Theme) =>
       border: `1px solid ${palette.grey.lighter2}`
     },
 
+    error: {
+      '&, &:hover:not($disabled)': {
```
denieler, line 24: what about achieving here
```
    error: {
      borderColor: palette.red.main
    },
```
?

packages/picasso-forms/src/utils/types.ts
dmaklygin: I was also talking with @augustobmoura today about it.
This is flexible solution to replace some keys of a interface. But, it's definitely difficult to realize how it works.

dmaklygin: There are possible solutions, like: https://stackoverflow.com/questions/41285211/overriding-interface-property-type-defined-in-typescript-d-ts-file/65561287#65561287
maybe it may help, at least in the naming of the Generic types.

augustobmoura: The team talked about it today, and the problem is not about being strongly typed but a specific interaction with `FieldWrapper` and `RichTextEditor` prop `defaultValue`. We agreed on making it more clear to understand why are we doing this way, but we still gonna need them for solving this problem the simplest way

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: True, I will remove it

packages/picasso-forms/src/utils/types.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Merge 2 object types, if the two have the a property with the same, prefer the type of the second one
+ */
+export type Merge<T1, T2> = {
+  [k in keyof NoIndex<T1>]: k extends keyof T2 ? T2[k] : NoIndex<T1>[k]
+} & OnlyIndex<T1>
```
denieler, line 40: sorry, but `Merge` still doesn't make more clear the typing 😞

packages/picasso-forms/src/Form/story/RichTextEditor.example.tsx
augustobmoura: Agree, I will remove it

packages/picasso-forms/src/Form/story/RichTextEditor.example.tsx
denieler: yeah, got it, it's ASTType 🙇 maybe we can move it to some variable, like `defaultValueAST`?

packages/picasso/src/RichTextEditor/styles.ts
augustobmoura: It still doesn't work. The other to do it is to plug a `:not($error)` into the `editorWrapper` style, it should be terser

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: Even though we are using `defaultProps` I would still use default in the variables, effectively using both approaches to prevent the usage of `eslint-ignore`, what you guys think?

packages/picasso-forms/src/utils/types.ts
dmaklygin: we merge to Record types with each other: it may be valuable when we have two types with the same attributes.

dmaklygin (approved):

packages/picasso-forms/src/Form/story/RichTextEditor.example.tsx
augustobmoura: Yep, I separated it into a variable, I think it's better now

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: then you would repeat yourself by setting the default value twice. I would be consistent and if you want you can create a ticket right away and refactor it codebase wide. We could have one constant `defaultValues` and we can use that in both places 🤔

TomasSlama (approved):

TomasSlama (approved):

ozgurkececioglu (approved):

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,49 @@
...
+} from '@toptal/picasso'
+import React, { useCallback, useState } from 'react'
+import { Object } from 'ts-toolbelt'
+
+import FieldWrapper, { FieldProps } from '../FieldWrapper'
+
+type OverriddenProps = {
+  defaultValue?: RichTextEditorProps['defaultValue']
+  value?: never
+  initialValue?: never
```
denieler, line 13: this typing looks not right 🤔

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -122,16 +138,23 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
       handleTextFormat,
       format: state.toolbar.format
     })
+
+    const wrappedOnFocus = useCallback(() => {
```
denieler, line 142:
```suggestion
    const handleFocus = useCallback(() => {
```

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -122,16 +138,23 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
...
+      setIsEditorFocused(true)
+      onFocus()
+    }, [setIsEditorFocused, onFocus])
+
+    const wrappedOnBlur = useCallback(() => {
```
denieler, line 147:
```suggestion
    const handleBlur = useCallback(() => {
```

packages/picasso-forms/package.json
```diff
@@ -34,7 +34,8 @@
     "final-form-arrays": "^3.0.2",
     "react-final-form": "^6.5.1",
     "react-final-form-arrays": "^3.1.3",
-    "react-final-form-listeners": "^1.0.3"
+    "react-final-form-listeners": "^1.0.3",
+    "ts-toolbelt": "^9.6.0"
```
denieler, line 38: I guess this should be dev dependency

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export type Props = Object.Merge<
+  OverriddenProps,
+  RichTextEditorProps & FieldProps<string>
+>
+
+type InternalProps = RichTextEditorProps & { value: string }
+
+export const RichTextEditor = ({ onChange, defaultValue, ...rest }: Props) => {
+  const [internalValue, setInternalValue] = useState('')
```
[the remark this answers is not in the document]
augustobmoura: I think `internalValue` makes it more explicit that is just the internal value that we keep for final-form, differentiating it from `props.value` that is another property, but I don't have a strong opinion on this

packages/picasso-forms/package.json
augustobmoura: You are right, will fix it

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
denieler: what if we move
```
const [isEditorFocused, setIsEditorFocused] = useState(autoFocus!) // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
into `useOnFocus` hook? then here we would be able to pass just
```
onBlur,
onFocus
```

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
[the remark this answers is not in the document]
denieler: idk, to me `props.value` and `value` are different 😃 also it's a single `useState` here, so as smaller variable name as better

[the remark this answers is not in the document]
augustobmoura: I prefer writing a name that makes its function obvious than a smaller name 😅
But again, no strong opinions, I will change it

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: We would need to provide `isEditorFocused` outside of `useOnFocus` hook. I prefer that the hook would only handle calling the callback, the logic is pretty complex.

In fact, I would even argue that we should move the line `toolbarActions.setDisabled(dispatch)(false)` to outside of the hook, wdyt?

yarn.lock
```diff
@@ -7618,6 +7618,7 @@ bottleneck@^2.18.1:
 boxen@^1.2.1, boxen@^1.3.0:
   version "1.3.0"
   resolved "https://registry.yarnpkg.com/boxen/-/boxen-1.3.0.tgz#55c6c39a8ba58d9c61ad22cd877532deb665a20b"
+  integrity sha512-TNPjfTr432qx7yOjQyaXm3dSR0MH9vXp7eT1BFSl/C51g+EFnOR9hTg1IreahGBmDNCehscshe45f+C1TBZbLw==
```
denieler, line 7621: it looks like locally after running `yarn` I see changed `yarn.lock` file

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
denieler: idk, on my taste in this file too many `onFocus` and `onBlur`. And they don't really bring any value, except just proxy-calling other methods 🤔 just making code look more complex. Maybe we can refactor this `useOnFocus` hook somehow 🤔

augustobmoura: I agree, there is a lot of room to improve, but I think these different refactoring would be better in a different PR. `useOnFocus` is a bit complex and I don't want to tackle in this task, I don't even know if I understand everything that is going on on it

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,49 @@
...
+  RichTextEditor as PicassoRichTextEditor,
+  RichTextEditorProps
+} from '@toptal/picasso'
+import React, { useCallback, useState } from 'react'
+import { Object } from 'ts-toolbelt'
+
+import FieldWrapper, { FieldProps } from '../FieldWrapper'
+
+type OverriddenProps = {
+  defaultValue?: RichTextEditorProps['defaultValue']
```
denieler, line 11: we can use here probably `ASTType` type 🤔

denieler (approved):

yarn.lock
augustobmoura: I could not reproduce that on my side, running `yarn` doesn't change `yarn.lock`, can you share what it is being changed?

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
augustobmoura: Done

PR #2547, opened 2022-03-26
feat(DatePicker): add popperProps that allow data-* attrs for the Popper

ascrazy:
[TOP-2072]
[FX-2565]

### Description

We at TopTeam use [data-private attribute to control visibility of some parts of UI for LogRocket](https://docs.logrocket.com/docs/privacy#exclude-data-in-videos). On of the things we need to hide this way is a DatePicker-powered input on member termination modal. Though currently if we render `<DatePicker data-private />` it's only going to be applied to the `<input` within date picker, while calendar popup still stays visible.

This PR adds proper support of `data-private` prop to DatePicker component, so that both input and calendar popup are hidden when you do render `<DatePicker data-private />`

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions. 

</details>

[TOP-2072]: https://toptal-core.atlassian.net/browse/TOP-2072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FX-2565]: https://toptal-core.atlassian.net/browse/FX-2565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/DatePicker/DatePicker.tsx
+++ b/packages/picasso/src/DatePicker/DatePicker.tsx
@@ -101,6 +101,10 @@ export interface Props
   timezone?: string
   /** Custom parser for `DatePicker`'s input value to process custom input value, like, human-readable dates */
   parseInputValue?: DatePickerInputCustomValueParser
+  /** Additional data-* attrs for the inner Popper */
+  popperProps?: {
+    [key: `data-${string}`]: unknown
+  }
   testIds?: InputProps['testIds'] & {
     calendar?: string
     input?: string
@@ -131,6 +135,7 @@ export const DatePicker = (props: Props) => {
     testIds,
     error,
     status,
+    popperProps,
     ...rest
   } = props
   const classes = useStyles()
@@ -388,6 +393,7 @@ export const DatePicker = (props: Props) => {
           container={popperContainer}
           popperOptions={DEFAULT_POPPER_OPTIONS}
           ref={popperRef}
+          {...popperProps}
         >
           <Calendar
             activeMonth={activeMonth}
--- a/.changeset/popular-beans-turn.md
+++ b/.changeset/popular-beans-turn.md
@@ -1,0 +1,8 @@
+---
+'@toptal/picasso': minor
+---
+
+---
+### DatePicker
+
+- add `popperProps` prop to let consumers control data-* attrs of the Popper


packages/picasso/src/DatePicker/DatePicker.tsx
```diff
@@ -344,7 +345,11 @@ export const DatePicker = (props: Props) => {
 
   return (
     <>
-      <Container inline={width !== 'full'} ref={inputWrapperRef}>
+      <Container
+        inline={width !== 'full'}
+        ref={inputWrapperRef}
+        {...getDataPrivate(props)}
```
ozgurkececioglu, line 351: I don't think you need this kind of spreading operation.
```suggestion
       data-private={props["data-private"]}
```
will do the same work.

ozgurkececioglu: Do you need this `data-private` in all of these components: `Container`, `Input`, and `Calendar`?

packages/picasso/src/DatePicker/DatePicker.tsx
```diff
@@ -97,6 +97,7 @@ export interface Props
     calendar?: string
     input?: string
   }
+  'data-private'?: string | boolean
```
ozgurkececioglu, line 100: I think this is too much information about a customization prop. We shouldn't be documenting these ones like a prop. I would prefer a solution like spreading input props into calendar as well 🤔

OndrejTuma: I'd add property like `wrapperProps` or `containerProps` that spreads on the container 🤔

packages/picasso/src/DatePicker/DatePicker.tsx
ascrazy: >I don't think you need this kind of spreading operation.

Will do. Somehow I missed react won't attach to the DOM those props that are explicitly set to `undefined`, so was trying to handle this smartly. Turned out I don't need it.

- [ ] pass data-private value explicitly instead of spreading

ascrazy: > Do you need this `data-private` in all of these components: `Container`, `Input`, and `Calendar` at the same time?

Well, what this PR is about. when DatePicker is rendered with data-private I need both input and popup excluded from the video. And as popup works trough react portals (rendered outside of the subtree) I can't just put that onto a container. 

Though I might not need it on both popup container and calendar. Will check

- [ ] Get rid of extra data-privates if possible

packages/picasso/src/DatePicker/DatePicker.tsx
ascrazy: I'm fine with `wrapperProps` approach (though I most probably need popperProps, not wrapper). But what would be the type definition for wrapperProps to allow `data-private`? If I still define `data-private` explicitly like `popperProps: { 'data-private': boolean | string }`, then I think it's not really any better then current approach when `data-private` is defined on top-level props and gets redirected to both input and popper. Or you have ideas how to define type to allow `data-private` implicitly?

ozgurkececioglu: I still don't think that we should provide a named prop for this purpose. We can define a `popperProps: object` or `popperProps: Record<string, any>` and just spread it to the popper component inside. We don't need to know which prop is being used. MUI has the same mechanism:

my examples with MUI DatePicker: https://codesandbox.io/s/basicdatepicker-material-demo-forked-pe352x?file=/demo.js
explanation: https://mui.com/guides/api/#spread

I know this is not exactly the same but we can follow it. WDYT?

augustobmoura: And, we shouldn't write code in Picasso relying on the existence of `data-` props, if it is something that needs to be put in an `if` or passed to someplace else is better to write a proper `private` property (without the `data-` prefix)

dmaklygin: great! let's use `wrapperProps` and `popperProps` then :)

ascrazy: >We can define a popperProps: object or popperProps: Record<string, any> and just spread it to the popper component inside.

I doubt this is a viable solution as it would let consumer provide any props popper knows, even those that are not supported by picasso but supported by MUI, which is kind of a leak in the abstraction that will let consumers access some internals we don't explicitly support. Will make it significantly harder to introduce changes in future. I was previously told that I should avoid introducing such leaks in picasso, e.g. here https://github.com/toptal/picasso/pull/2315#discussion_r766444248

>[Typescript will always accept data- props](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChTgA7GJKTdJOAMQggAUcwBnOAb1LhxgAEwD8ALjhcYUKgHNSAX3Kk0EStOas4AXjgAKMJy6SW7YwEpdAPjgAeYcABu1lF2GY7AekcuVXrzhhFBgUAFojZxDGYB4IAGs4AAtaRkxoOAAVAE8wJC40WTAYVXVNJAAPXXszIJDwyKdouC9rIA) in any element be it typed or no

that looks interesting, I do not see how it could work. I got you point that you can use data-private in JSX even when it's not typed explicitly, though in this case I need to provide it outside of JSX
```
<DatePicker popperProps={{ 'data-private': true }} />
```
and in this case popperProps are strictly controlled by the type definition. This definition 
```
wrapperProps: { [`data-${string}`]: unknown }
```` 
you suggested seem to be invalid, as you are trying to interpolate type into a value here

Comment (ascrazy) [reply packages/picasso/src/DatePicker/DatePicker.tsx:100]: > great! let's use `wrapperProps` and `popperProps` then :)

I'm totally fine with it, and would actually love if we could avoid adding any knowledge about data-private and its semantics to picasso. I just do not see any working solution apart from defining `popperProps: Record<string, unknown>` as @ozgurkececioglu suggested above, though I doubt this is really a good one

Comment (ascrazy) [reply packages/picasso/src/DatePicker/DatePicker.tsx:100]: >Maybe we should consider updating it, it depends on how hard that it would be of course

@augustobmoura do you think someone from FX team could do this in the near future? and how this is going to affect consumers? are they going to be required to upgrade to 4.4+ too?

Comment (augustobmoura) [reply packages/picasso/src/DatePicker/DatePicker.tsx:107]: After some discussion with the team, we agreed in upgrading the Typescript version. I will start working on it right now. 

Projects that depend on `picasso` will need to update their typescript version as well, but after some discussion between the team we think that this should be straightforward and often necessary (if not customary). Do you think that this would be a problem for topteam?

Comment (augustobmoura) [reply packages/picasso/src/DatePicker/DatePicker.tsx:107]: After some discussion with the team, we agreed in upgrading the Typescript version. I will start working on it right now. 

Projects that depend on `picasso` will need to update their typescript version as well, but after some discussion between the team we think that this should be straightforward and often necessary (if not customary). Do you think that this would be a problem for topteam?

Comment (ascrazy) [reply packages/picasso/src/DatePicker/DatePicker.tsx:107]: Topteam is actually already at 4.3.5 and I don't think it's going to be a problem for us.

So will be waiting on the upgrade to arrive in the picasso and then will update this PR accordingly.

Comment (TomasSlama) [reply packages/picasso/src/DatePicker/DatePicker.tsx:107]: I'm just wondering, isn't already all data attributes part of 'HTMLDivElement'?, can't we somehow reuse that?

Comment (ascrazy) [reply packages/picasso/src/DatePicker/DatePicker.tsx:107]: > I'm just wondering, isn't already all data attributes part of 'HTMLDivElement'?, can't we somehow reuse that? 🤔 🤷

Hi @TomasSlama,

From what I understand, it's not. It's allowed in the JSX because there's special exception in how TS handles JSX. Though it does not work outside of JSX

https://www.typescriptlang.org/docs/handbook/jsx.html
image

and as I mentioned above, in our case it's going to be defined outside if JSX

```
<DatePicker data-private popperProps={{ 'data-private': true }} />
```

ascrazy:
this is blocked by #2569 at the moment

denieler:
@ascrazy #2569 has been merged 👍 you are unblocked

packages/picasso/src/DatePicker/DatePicker.tsx
ascrazy: Hi guys,

Just updated PR accordingly to discussion above. Please have a look!

ozgurkececioglu (approved):

augustobmoura (approved):

OndrejTuma (approved):
looks good 👍

PR #2566, opened 2022-03-30
fix: handle empty children of ReactNode

TomasSlama:
[FX-NNNN]

### Description

Update RichText stories, so it is clear how to use the RichText component. While testing in the storybook, an unhandled error occurred, when ReactNode had no children, so I had to add a small fix.

### How to test

- check stories of RichText

### Screenshots

| image | image |

### Review

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

### DIFF
--- a/.changeset/late-vans-type.md
+++ b/.changeset/late-vans-type.md
@@ -1,0 +1,8 @@
+---
+'@toptal/picasso': patch
+---
+
+---
+### RichText
+
+handle empty children of ReactNode
--- a/.changeset/late-vans-type.md
+++ b/.changeset/late-vans-type.md
@@ -5,4 +5,4 @@
 ---
 ### RichText
 
-handle empty children of ReactNode
+Handle empty children of ReactNode.
--- a/packages/picasso/src/RichText/hooks/useRichText/test.tsx
+++ b/packages/picasso/src/RichText/hooks/useRichText/test.tsx
@@ -7,7 +7,7 @@ import useRichText from './useRichText'
 describe('useRichText', () => {
   describe('not allowed tags', () => {
     it('returns the unmapped tag', () => {
-      const tree = ({
+      const tree = {
         type: 'root',
         children: [
           {
@@ -17,7 +17,7 @@ describe('useRichText', () => {
             children: [{ type: 'text', value: 'foobar' }]
           }
         ]
-      } as unknown) as ASTType
+      } as unknown as ASTType
       const { result } = renderHook(() => useRichText(tree))
 
       const header = result.current as ReactElement
@@ -120,5 +120,56 @@ describe('useRichText', () => {
     expect(liElementFirst.props.children).toEqual(['foo'])
     expect((liElementSecond.type as Function).name).toEqual('Li')
     expect(liElementSecond.props.children).toEqual(['bar'])
+  })
+
+  it('handles empty children', () => {
+    const { result } = renderHook(() =>
+      useRichText({
+        type: 'root',
+        children: []
+      })
+    )
+
+    expect(result.current).toBeNull()
+  })
+
+  it('handles empty children deeper in the tree', () => {
+    const tree: ASTType = {
+      type: 'root',
+      children: [
+        {
+          type: 'element',
+          tagName: 'h3',
+          properties: {},
+          children: [{ type: 'text', value: 'foobar' }]
+        },
+        {
+          type: 'element',
+          tagName: 'br',
+          properties: {},
+          children: []
+        }
+      ]
+    }
+
+    const { result } = renderHook(() => useRichText(tree))
+
+    const [headingElement, brElement] = result.current as [
+      ReactElement,
+      ReactElement
+    ]
+
+    expect((headingElement.type as Function).name).toEqual('H3')
+    expect(brElement.type).toEqual('br')
+  })
+
+  it('handles undefined children', () => {
+    const { result } = renderHook(() =>
+      useRichText({
+        type: 'root'
+      })
+    )
+
+    expect(result.current).toBeNull()
   })
 })
--- a/packages/picasso/src/RichText/hooks/useRichText/test.tsx
+++ b/packages/picasso/src/RichText/hooks/useRichText/test.tsx
@@ -121,55 +121,60 @@ describe('useRichText', () => {
     expect((liElementSecond.type as Function).name).toEqual('Li')
     expect(liElementSecond.props.children).toEqual(['bar'])
   })
-
-  it('handles empty children', () => {
-    const { result } = renderHook(() =>
-      useRichText({
-        type: 'root',
-        children: []
-      })
-    )
-
-    expect(result.current).toBeNull()
-  })
-
-  it('handles empty children deeper in the tree', () => {
-    const tree: ASTType = {
-      type: 'root',
-      children: [
-        {
-          type: 'element',
-          tagName: 'h3',
-          properties: {},
-          children: [{ type: 'text', value: 'foobar' }]
-        },
-        {
-          type: 'element',
-          tagName: 'br',
-          properties: {},
+  describe('when children are empty', () => {
+    it('returns null', () => {
+      const { result } = renderHook(() =>
+        useRichText({
+          type: 'root',
           children: []
-        }
-      ]
-    }
+        })
+      )
 
-    const { result } = renderHook(() => useRichText(tree))
-
-    const [headingElement, brElement] = result.current as [
-      ReactElement,
-      ReactElement
-    ]
-
-    expect((headingElement.type as Function).name).toEqual('H3')
-    expect(brElement.type).toEqual('br')
+      expect(result.current).toBeNull()
+    })
   })
 
-  it('handles undefined children', () => {
-    const { result } = renderHook(() =>
-      useRichText({
-        type: 'root'
-      })
-    )
+  describe('when children of child are empty', () => {
+    it('returns correct node', () => {
+      const tree: ASTType = {
+        type: 'root',
+        children: [
+          {
+            type: 'element',
+            tagName: 'h3',
+            properties: {},
+            children: [{ type: 'text', value: 'foobar' }]
+          },
+          {
+            type: 'element',
+            tagName: 'br',
+            properties: {},
+            children: []
+          }
+        ]
+      }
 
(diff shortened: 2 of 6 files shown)


.changeset/late-vans-type.md
```diff
@@ -0,0 +1,8 @@
+---
+'@toptal/picasso': patch
+---
+
+---
+### RichText
+
+handle empty children of ReactNode
```
denieler, line 8:
```suggestion
Handle empty children of ReactNode.
```

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -64,18 +64,27 @@ const picassoMapper = (child: ReactNode): ReactNode => {
 
   const type =
     componentMap[child.type as keyof typeof componentMap] || child.type
-  const mappedChildren = child.props.children.map(picassoMapper)
+
+  const hasChildChildren = child.props.children
```
denieler, line 68:
```suggestion
  const hasChildChildren = Boolean(child.props.children)
```
it will work even without this, but we do this for a better readability almost everywhere in the code

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -64,18 +64,27 @@ const picassoMapper = (child: ReactNode): ReactNode => {
...
+    ? child.props.children.map(picassoMapper)
+    : null
 
   return createElement(type, { key: child.key }, mappedChildren)
 }
 
 const useRichText = (value: ASTType): ReactNode[] | ReactNode => {
   const mappedTextNodes = useMemo(() => {
-    if (!value.children) {
+    const isEmpty =
```
denieler, line 79:
```suggestion
    const isEmpty = React.Children.count(props.children) === 0
```
should work I guess

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -64,18 +64,27 @@ const picassoMapper = (child: ReactNode): ReactNode => {
...
+      !value.children ||
+      (Array.isArray(value.children) && value.children.length === 0)
+
+    if (isEmpty) {
       return null
     }
 
-    const isSingleChild = value.children.length === 1
+    const isSingleChild = value.children?.length === 1
```
denieler, line 87:
```suggestion
    const isSingleChild = React.Children.only(props.children)
```
can be used, but it throws an error if not only 1 children. You also can use `React.Children.count(props.children) === 1` as well

packages/picasso/src/RichText/story/HTML.example.tsx
```diff
@@ -1,14 +1,142 @@
...
+          onChange={onChange}
+          id='editor'
+        />
+      </Grid.Item>
+      <Grid.Item small={12} large={6}>
+        <RichText value={htmlToHast(html)} />
+      </Grid.Item>
+    </Grid>
+  )
+  // return
```
denieler, line 24: 🤔

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
TomasSlama: @denieler I tried that, and it does not work, we don't work here with React children, but with children from AST type. I have refactored it a little bit so it is cleaner code, and I have also added unit tests to cover those errors that happened to me.

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
TomasSlama: the same, we work with AST.children, not react.

packages/picasso/src/RichText/story/HTML.example.tsx
TomasSlama: ✅

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -64,18 +64,25 @@ const picassoMapper = (child: ReactNode): ReactNode => {
...
 
   return createElement(type, { key: child.key }, mappedChildren)
 }
 
 const useRichText = (value: ASTType): ReactNode[] | ReactNode => {
   const mappedTextNodes = useMemo(() => {
-    if (!value.children) {
+    const { children } = value
+
+    if (!children || children.length === 0) {
```
augustobmoura, line 81: Could be simplfied to:
```suggestion
    if (!children?.length) {
```

packages/picasso/src/RichText/story/HTML.example.tsx
```diff
@@ -1,14 +1,141 @@
...
-import { RichText } from '@toptal/picasso'
+import React, { useCallback, useState } from 'react'
+import { RichText, RichTextEditor, Grid, ASTType } from '@toptal/picasso'
 import { htmlToHast } from '@toptal/picasso/utils'
 
-const html =
-  '<h3>Position Description</h3><p>We’re looking for hardworking, self-starting Designers for our <strong>Product Design</strong> team to help us define how talent interacts with Toptal.</p><p>You’ll build beautiful and inspiring design experiences that help users discover and connect with resources they need in truly innovative ways.</p><h3>Requirements</h3><ol><li>Collaborate with PMs and other designers to ship your first product features.</li><li>Learn about our design system.</li></ol><h3>Requirements</h3><ul><li>Proficiency with various design and prototyping tools (such as Sketch, Abstract, Marvel, Principle, Figma), as well as knowledge of HTML and CSS.</li><li>An understanding that phenomenal experiences come from collaborative decision-making with front-end developers, engineers, researchers, content strategists, and other disciplines.</li></ul>'
+const Example = () => {
+  const [html, setHtml] = useState<string>('')
```
augustobmoura, line 6: This type is unnecessary, Typescript already infers it as a `string` because of the default parameter

```suggestion
  const [html, setHtml] = useState('')
```

packages/picasso/src/RichText/story/HTML.example.tsx
```diff
@@ -1,14 +1,141 @@
...
 
-const style = { maxWidth: '500px' }
+  const onChange = useCallback((value: string) => setHtml(value), [setHtml])
```
augustobmoura, line 8: `useCallback` unnecessary because setters from `useState` are already "wrapped" in `useCallback` (they are not really wrapped but they always have the same identity, see the note [in the docs](https://reactjs.org/docs/hooks-reference.html#usestate))
```suggestion
  const onChange = setHtml
```

In this case this variable is even necessary?

augustobmoura (changes requested):

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
denieler: @TomasSlama what? why then it named `children`?

packages/picasso/src/RichText/hooks/useRichText/test.tsx
```diff
@@ -121,4 +121,55 @@ describe('useRichText', () => {
     expect((liElementSecond.type as Function).name).toEqual('Li')
     expect(liElementSecond.props.children).toEqual(['bar'])
   })
+
+  it('handles empty children', () => {
```
dmaklygin, line 125: let's maybe rename the test case. 
I think we should test whether it handles children or not because then we have to cover as well other inputs.
Instead, I would suggest checking just two cases: when children are empty and are not empty.
The idea is to create these use cases via `describe` block and expect the result
In case when chilren are empty, the hook must return `null`

packages/picasso/src/RichText/hooks/useRichText/test.tsx
```diff
@@ -121,4 +121,55 @@ describe('useRichText', () => {
...
+    const [headingElement, brElement] = result.current as [
+      ReactElement,
+      ReactElement
+    ]
+
+    expect((headingElement.type as Function).name).toEqual('H3')
+    expect(brElement.type).toEqual('br')
+  })
+
+  it('handles undefined children', () => {
```
dmaklygin, line 166: I think, this is also a case we have to cover. 
`when children are not defined`
and the response:
`returns null`

dmaklygin (approved):

ozgurkececioglu (approved):

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
TomasSlama: @denieler [hast](https://github.com/syntax-tree/hast) also uses children, and `value` here is actually just HAST format.

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -64,18 +64,21 @@ const picassoMapper = (child: ReactNode): ReactNode => {
...
+
+  const mappedChildren = child.props.children?.map(picassoMapper) ?? null
 
   return createElement(type, { key: child.key }, mappedChildren)
 }
 
 const useRichText = (value: ASTType): ReactNode[] | ReactNode => {
   const mappedTextNodes = useMemo(() => {
-    if (!value.children) {
+    const { children } = value
```
denieler, line 75:
```suggestion
    const { children: astChildren } = value
```
what about changing to this?

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
```diff
@@ -64,18 +64,21 @@ const picassoMapper = (child: ReactNode): ReactNode => {
...
+
+    if (!children?.length) {
       return null
     }
 
-    const isSingleChild = value.children.length === 1
+    const isSingleChild = children.length === 1
```
denieler, line 81: also further I see this code:
```
const textNodes = reactElement.props.children
return textNodes.map(picassoMapper)
```

is this about react children or again ast children?

packages/picasso/src/RichText/hooks/useRichText/useRichText.tsx
TomasSlama: I can do that 👍

augustobmoura (approved):

PR #2701, opened 2022-04-28
chore: [FX-2607] update inline alert design

LashaJini:
[FX-2607](https://toptal-core.atlassian.net/browse/FX-2607?atlOrigin=eyJpIjoiY2ZhY2IzYmZlMzQ4NDMwMDg0OTIzY2RjNGE1NmM4NDIiLCJwIjoiaiJ9)

[Figma](https://www.figma.com/file/q2nvjiyO2CLqBv4DeJnU3i/Product-Library-Documentation?node-id=19%3A353)

### Description

- Change text color so it satisfies the design
- Change font size from `14px` to `13px`
- Change font weight from `400` to `600`
- Change font weight from `22px` to `20px`

> NOTE:
> Team decided to separate block `Alert` logic from `AlertInline`.

### How to test

- [Picasso Inline Alert example](https://picasso.toptal.net/fx-2607-update-inline-alert-design/?path=/story/components-alert--alert#default) should look like the one in the figma

### Screenshots

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

- [x] codemod is created and showcased in the changeset

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


packages/picasso/src/AlertInline/AlertInline.tsx
```diff
@@ -46,7 +74,23 @@ export const AlertInline = forwardRef<HTMLDivElement, Props>(function Alert(
...
 ) {
   const classes = useStyles()
-  const { variant, children, className, iconPadding } = props
+  const { variant, children, className, iconPadding, alertVariant } = props
+
+  let typographyColor = variant as ColorType
+
+  if (variant === 'blue') {
+    typographyColor = 'light-blue'
+  }
```
LashaJini, line 83: We could replace `blue` with `light-blue` in the `VariantType` (which would have been more consistent with actual colors being passed), but it would have been a breaking change.

.changeset/red-socks-fly.md
```diff
@@ -0,0 +1,9 @@
+---
+'@toptal/picasso-shared': minor
+---
+
+---
+
+### Color Type
+
+- Add `ligt-blue` to shared color types
```
TomasSlama, line 9:
```suggestion
- Add `light-blue` to shared color types
```

.changeset/selfish-nails-lick.md
```diff
@@ -0,0 +1,22 @@
...
+
+---
+
+### AlertInline
+
+- Use `AlertInline` specific icons
+- Use correct colors for typography
+- Change font weight from `400` to `600`
+- Change font size from `14px` to `13px`
+- Change line height from `22px` to `20px`
```
TomasSlama, line 13: Do you think this is important for users to know? 🤔 

How about something like `- Updated styles to be compliment with BASE design`? WDYT?

packages/picasso/src/AlertInline/AlertInline.tsx
```diff
@@ -1,39 +1,67 @@
...
 
+type AlertVariant = 'inline' | 'block'
+
 export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
   /** Main content of the Alert */
   children?: ReactNode
   /** Style variant of Alert */
   variant?: VariantType
   iconPadding?: SizeType<'xsmall' | 'small'> // undocumented prop, only for internal usage
+  alertVariant?: AlertVariant // undocumented prop, only for internal usage
```
TomasSlama, line 30: Since `inline` and `block` are values of the `display` property in CSS, how about naming this property `displayVariant`?

packages/picasso/src/AlertInline/AlertInline.tsx
```diff
@@ -1,39 +1,67 @@
...
 
     case 'blue':
-      return <Info16 color='light-blue' />
+      return <InfoSolid16 color='light-blue' />
 
     case 'yellow':
-      return <Exclamation16 color='yellow' />
+      return <ExclamationSolid16 color='yellow' />
   }
 }
```
TomasSlama, line 66:
```
const AlertIconsMap = {
  block: {
    red: <Exclamation16 color='red' />,
    // ...
  },
  // inline: { ... }
}

// usage
const Icon = AlertIconMap[alertVariant][variant]
```

What do you think about this approach? 🤔

packages/picasso/src/AlertInline/AlertInline.tsx
```diff
@@ -46,7 +74,23 @@ export const AlertInline = forwardRef<HTMLDivElement, Props>(function Alert(
...
+
+  let size = 'small' as TypographyProps['size']
+  let weight = 'semibold' as TypographyProps['weight']
+  let color = typographyColor
+
+  if (alertVariant === 'block') {
+    size = 'medium'
+    color = 'black'
+    weight = 'regular'
+  }
```
TomasSlama, line 93: const SizeMap = {
 inline: 'small',
 block: 'medium'
}

TomasSlama (commented):
Good job Lasha, I have just checked the code and left some comments. I will be checking design little bit later.

packages/picasso/src/AlertInline/AlertInline.tsx
LashaJini: hmm, but it's not quite related to `display` property in CSS. I just got the "inspiration" from [figma's](https://www.figma.com/file/q2nvjiyO2CLqBv4DeJnU3i/Product-Library-Documentation?node-id=8%3A2) variants.

packages/picasso/src/AlertInline/AlertInline.tsx
LashaJini: Better :+1:

packages/picasso/src/AlertInline/AlertInline.tsx
TomasSlama: @LashaJini yeah, in figmat it is right, because it is actually `variant`, but we used it already for `color`

TomasSlama: but you are right, maybe `display` is not the right naming either. Maybe someone else will have better idea 🤷

.changeset/red-socks-fly.md
```diff
@@ -0,0 +1,9 @@
+---
+'@toptal/picasso-shared': minor
+---
+
+---
+
+### Color Type
+
+- Add `light-blue` to shared color types
```
denieler, line 9:
```suggestion
- Add `light-blue` color
```
🤷‍♂️

.changeset/selfish-nails-lick.md
```diff
@@ -0,0 +1,18 @@
...
+- Updated styles to be compliant with BASE design
+
+### Icons
+
+- Add `DoneSolid16`, `DoneSolid24` icons
+- Add `InfoSolid16`, `InfoSolid24` icons
+
+### Typography
+
+- Add `lightBlue` style
```
denieler, line 18:
```suggestion
- Add `lightBlue` color
```
because the prop name is `color` in `Typography` component has been added

packages/picasso/src/AlertInline/__snapshots__/test.tsx.snap
```diff
@@ -12,17 +12,17 @@ exports[`Alert renders 1`] = `
...
       >
         <svg
-          class="PicassoSvgExclamation16-root PicassoSvgExclamation16-yellow"
+          class="PicassoSvgExclamationSolid16-root PicassoSvgExclamationSolid16-yellow"
           style="min-width: 16px; min-height: 16px;"
           viewBox="0 0 16 16"
         >
           <path
-            d="M7.5 15a7.5 7.5 0 1 1 0-15 7.5 7.5 0 0 1 0 15Zm0-1a6.5 6.5 0 1 0 0-13 6.5 6.5 0 0 0 0 13ZM7 3h1v6H7V3Zm0 8h1v1H7v-1Z"
+            d="M8 0a8 8 0 1 1 0 16A8 8 0 0 1 8 0Zm0 11a1 1 0 1 0 0 2 1 1 0 0 0 0-2Zm0-8a1 1 0 0 0-1 1v5a1 1 0 1 0 2 0V4a1 1 0 0 0-1-1Z"
```
dmaklygin, line 20: fix me, but according to Figma design, icons should be kept as is. no?

packages/picasso/src/AlertInline/AlertInline.tsx
dmaklygin: let's maybe follow Figma design and name it `variant` — then it will be a breaking change

packages/picasso/src/AlertInline/AlertInline.tsx
dmaklygin: wdyt if we avoid adding this workaround and use exactly what specified in the Figma design?

packages/picasso/src/AlertInline/AlertInline.tsx
TomasSlama: ☝️ might be the optimal idea. `variant` and `color`. But then we are increasing the scope of the ticket a lot 🤔

packages/picasso/src/AlertInline/AlertInline.tsx
```diff
@@ -56,13 +89,14 @@ export const AlertInline = forwardRef<HTMLDivElement, Props>(function Alert(
         alignItems='center'
         className={classes.iconWrapper}
       >
-        {renderAlertIcon(variant)}
+        {icon}
       </Container>
       <Typography
-        size='medium'
+        size={alertVariant && typographyProps[alertVariant].size}
```
TomasSlama, line 95:
```suggestion
        size={typographyProps[alertVariant!].size}
```

Since we have `alertVariant` in `defaultProps`, we know that they will always be around. We will need to disable the eslint rule locally for this, but we do it a lot in the codebase.

TomasSlama: and we can even use `const { size, weight, color } = typographyProps[AlertVariant!]` so we have it only in one place

packages/picasso/src/AlertInline/AlertInline.tsx
TomasSlama: @dmaklygin after checking the storybook, I think this might be unnecessary, because we don't expose the `alertVariant` to public. We use `Alert` component when we need block variant and `Alert.Inline` for inline. So the breaking change is probably not necessary, WDYT?

TomasSlama (changes requested):
@LashaJini it seems we are using different color for success alert
image
image

packages/picasso/src/AlertInline/__snapshots__/test.tsx.snap
LashaJini: Well, `AlertInline` component has solid icons while `Alert` has just icons.

packages/picasso/src/AlertInline/AlertInline.tsx
LashaJini: could you specify wdym, please?

packages/picasso/src/AlertInline/AlertInline.tsx
dmaklygin: I'm completely confused :)
`Alert.Inline` has `alertVariant` prop which may be either `block` or `inline`, correct?

LashaJini: Yes.

LashaJini: And `Alert` extends `Alert.Inline`'s props as well.

packages/picasso/src/AlertInline/AlertInline.tsx
dmaklygin: I mean, if we want to keep backward compatibility, then this workaround is okay. 
But, we have to put a comment to keep in mind that `blue` is a leftover, and it exists only because we want to keep legacy here.
I just suggested to rework it, and use colors that are specified in design. And remove this `blue` color, because it's not specified in design anymore

LashaJini: Yes, I also agree. And in the design we also have dark green, instead of regular green. I just didn't want to complicate this simple task, I will create a ticket when this one is merged:

- Replace `blue` with `light-blue`
- Replace `green` with `dark-green`
- Add `light-blue` and `dark-green` to Container (cuz we don't have them there, and we are passing colors to the container)
- Remove redundant `blue` and `green` colors
- Write codemod to do the replacing

packages/picasso/src/AlertInline/AlertInline.tsx
OndrejTuma: I dont understand why we need `block` variant for inline Alert? 🤔 
But if it's only these two variants, maybe just keep a boolean variable `inline` with true/false

OndrejTuma: OK I checked it once more and now I think we should do separate implementation for block and for inline variant. Think seems to much tweaking for me. WDYT?

LashaJini: It is indeed to much, cuz it's weird to handle `block` variant inside inline `Alert`. It needs quite amount of refactoring + breaking change + codemod.

TomasSlama (changes requested):
image

The wrapper of SVG has 22px, it should be 20px, so it is vertically centered with text. Maybe we could discuss during our daily, if we should not refactor these alerts into two separate Components as @OndrejTuma suggested.

packages/picasso/src/AlertInline/AlertInline.tsx
LashaJini: https://github.com/toptal/picasso/pull/2701#discussion_r861643898 + separating `block` and `inline`.

packages/picasso/src/Alert/Alert.tsx
```diff
@@ -38,16 +65,33 @@ export const Alert = forwardRef<HTMLDivElement, Props>(function Alert(
...
+      </Container>
       {onClose && renderAlertCloseButton({ onClose })}
     </Container>
   )
 })
 
 Alert.defaultProps = {
-  variant: 'yellow'
+  variant: 'yellow',
+  iconPadding: 'small'
```
TomasSlama, line 94: Why do we need this new internal prop? 🤔

LashaJini: Good point :+1: . I guess `AlertInline` will always have `xsmall`, while `Alert` will have `small` for `iconPadding`

dmaklygin (approved):

packages/picasso/src/Alert/Alert.tsx
```diff
@@ -21,11 +34,24 @@ const renderAlertCloseButton = ({ onClose }: Pick<Props, 'onClose'>) => (
...
 
+const icons = {
+  red: <Exclamation16 color='red' />,
+  green: <Done16 color='dark-green' />,
+  blue: <Info16 color='light-blue' />,
+  yellow: <Exclamation16 color='yellow' />
+}
+
+const useStyles = makeStyles<Theme>(styles, {
+  name: 'PicassoAlertInline'
```
OndrejTuma, line 45: Probably just:
```suggestion
  name: 'PicassoAlert'
```
?

augustobmoura (approved):

OndrejTuma (approved):
Nice work!

.changeset/selfish-nails-lick.md
```diff
@@ -0,0 +1,18 @@
...
+---
+
+### AlertInline
+
+- Updated styles to be compliant with BASE design
+
+### Icons
+
+- Add `DoneSolid16`, `DoneSolid24` icons
+- Add `InfoSolid16`, `InfoSolid24` icons
```
OndrejTuma, line 14:
```suggestion
- Added `DoneSolid16`, `DoneSolid24` icons
- Added `InfoSolid16`, `InfoSolid24` icons
- Added `darkGreen` color 
```

packages/picasso/src/Alert/Alert.tsx
LashaJini: :+1: , good catch. Thank you.

.changeset/selfish-nails-lick.md
LashaJini: Do we use past simple while writing changeset, :thinking: ? Or it does not matter?

TomasSlama: Good question @LashaJini, maybe we should use `present simple` as in commits, we can discuss in daily

LashaJini: We can quickly discuss it on daily :+1:

PR #2757, opened 2022-05-11
chore: remove .yarn files from repo

TomasSlama:
[FX-NNNN]

### Description

add needed yarn files to the `.gitattributes` file as it is listed in [docs](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored)

### Development checks

- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


denieler (approved):
It is bad to keep in the repo files we don't know why exist or we don't have explicit use from them. The reason is they confuse people. They take additional effort in maintenance.

TomasSlama:
> It is bad to keep in the repo files we don't know why exist or we don't have explicit use from them. The reason is they confuse people. They take additional effort in maintenance.

@denieler so you think, we should remove `/.yarn/releases` from git? 🤔

separatio (approved):
@denieler In this case, I take these binaries as I take `yarn.lock` partially.

Ignoring that I know what `yarn.lock` does, I also know that it is best to keep it in .git.
So I'd go with the official docs recommendation from yarn, even though it's not super clear on their purpose. We could potentially see other failures later if we don't keep the binaries in git.

So I'm going to go for a :+1: on this PR.

denieler:
> One more comment. As the documentation states:
> 
> image
> 
> Overall, keeping them won't hurt.

@separatio those docs are for `yarn v2` which we don't use at Toptal

augustobmoura:
Even though this is on the docs, I never see it in the wild, and no major project uses it. If we want to keep the version of yarn consistent among the team we should just use an external versioning tool for the job

augustobmoura:
Even though this is on the docs, I never see it in the wild, and no major project uses it. If we want to keep the version of yarn consistent among the team we should just use an external versioning tool for the job

separatio:
Since the docs are for yarn v2, then we might as well ditch this change and the file, just adding it to .gitignore.

dmaklygin (approved):

denieler:
I agree with @augustobmoura that we need consistency. And we know - we use yarn v1.

ozgurkececioglu (approved):

OndrejTuma (approved):

PR #2770, opened 2022-05-16
chore: [FX-2739] Add madge circularity check

denieler:
[FX-2739]

### Description

Add circularity check via [Madge](https://github.com/pahen/madge)

### How to test

- `yarn install`
- `yarn circularity`

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2739]: https://toptal-core.atlassian.net/browse/FX-2739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.madgerc
+++ b/.madgerc
@@ -1,0 +1,4 @@
+{
+  "fileExtensions": ["js", "ts", "jsx", "tsx"],
+  "tsConfig": "./tsconfig.json"
+}
--- a/.madgerc
+++ b/.madgerc
@@ -1,5 +1,4 @@
 {
   "fileExtensions": ["js", "ts", "jsx", "tsx"],
-  "tsConfig": "./tsconfig.json",
-  "excludeRegExp": ["dist-package"]
+  "tsConfig": "./tsconfig.json"
 }
--- a/.madgerc
+++ b/.madgerc
@@ -1,4 +1,5 @@
 {
   "fileExtensions": ["js", "ts", "jsx", "tsx"],
-  "tsConfig": "./tsconfig.json"
+  "tsConfig": "./tsconfig.json",
+  "excludeRegExp": ["dist-package"]
 }
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
   "scripts": {
     "build:package": "lerna run build:package --stream --parallel",
     "build:storybook": "build-storybook -c .storybook -o build/storybook --quiet",
-    "commit": "git-cz",
+    "circularity": "madge --circular packages",
     "generate:component": "davinci code new component",
     "generate:example": "davinci code new example",
     "generate:icons": "./bin/generate-icons",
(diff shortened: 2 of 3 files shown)


separatio:
Is this PR going to also fix circularity issues? I found 91 locally.

denieler:
> Is this PR going to also fix circularity issues? I found 91 locally.

@separatio no, only adding the command to check them 😃

separatio (approved):
Straightforward. Looks cool. Can't wait to use it more.

.madgerc
```diff
@@ -0,0 +1,4 @@
+{
+  "fileExtensions": ["js", "ts", "jsx", "tsx"],
+  "tsConfig": "./tsconfig.json"
```
separatio, line 3: Should we exclude `dist-package` directories via `excludeRegExp` or something else?

denieler: yeah, let me check 👍

denieler: @separatio done 👍 `dist-package` folders are excluded

.madgerc
```diff
@@ -0,0 +1,5 @@
+{
+  "fileExtensions": ["js", "ts", "jsx", "tsx"],
+  "tsConfig": "./tsconfig.json",
+  "excludeRegExp": ["dist-package"]
```
TomasSlama, line 4: should we also exclude stories and tests? 🤔

separatio: Probably. I just encountered an edge case that included them.

denieler: @TomasSlama let's do that action whenever we will see the problem exists. So far I haven't seen any tests of stories in the report

TomasSlama (approved):

.madgerc
separatio: If you run `yarn circularity -s`, which gives a summary of all files checked, you can see the stories, with 0 circularities offenses. The same goes for styles.ts.

separatio: As for tests, we actually have offenses for tests. For example, I found 4 circularity dependencies offenses in `picasso/src/NativeSelect/test.tsx`

denieler: ok @separatio I've checked and it looks like `excludeRegExp` option doesn't exclude files from the check, but exclude them only from the output. So it's pretty hard to properly filter files from the check in `madge`. I stopped at the current solution
```
madge --circular packages/*/src
```

it works under 6 sec, so looks like good enough 👍 if we would have circularity in our tests or stories - it will report that, but so far we didn't have any cases like that, so I would wait until that moment. Efforts to fix or remove some further files would be noticed and the code qualify and maintenance will be not good enough. So I don't think it worth it.

PR #2778, opened 2022-05-16
fix: [FX-2739] Circularity in compound components - part 3

denieler:
[FX-2739]

### Description

Fix circularity in different compound components with 2 fixes:
- Move compound to `*Compound` component folder
- Remove a type and replace it with just a union of string literals in multiple components

### How to test

- `yarn circularity`
- To have -17 circularity issues

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2739]: https://toptal-core.atlassian.net/browse/FX-2739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/picasso/src/List/List.tsx
```diff
@@ -6,12 +6,10 @@ import cx from 'classnames'
 import ListItem from '../ListItem'
 import styles from './styles'
 
-export type ListVariant = 'ordered' | 'unordered'
-
 export type Props = BaseProps & {
   children: ReactNode
   /** The variant to use */
-  variant: ListVariant
+  variant: 'ordered' | 'unordered'
```
ozgurkececioglu, line 12: I'm curious, what is the motivation behind this? We will lose the type safety across the other compounded components for `List`. And to be more consistent, it should be applied to some other places, right?

dmaklygin: I kinda agree with @ozgurkececioglu . We could pick up the `ListVariant` type and move it to the some place, so it might be easily imported to several components

denieler: @ozgurkececioglu let's assume we have Base and User component. In code we will have 2 situations:

1. We defined in Base `variant: 'ordered' | 'unordered'` type, then we try to type User `variant` prop

```
Base Props {
  /* docs */
  variant: 'ordered' | 'unordered'
}

User Props {
  variant: ?
}

Base Component {
  const { variant } = baseProps
  ...
  <User variant={variant} ... />
}
```

in this case, if we do a mistake in User `variant` type and set it to type `variant: 'ordered'` - TS would raise an error, that extra value possible for this prop. If we type with an additional value - `variant: 'ordered' | 'unordered' | 'fake'` - we would have an extra possible code inside the User component, which would be processing the `fake` value. It would not be a huge mistake, just unusable code, but it's pretty easily detectable.

2. We defined in User `variant: 'ordered' | 'unordered'` type, then we try to type Base `variant` prop

```
Base Props {
  /* docs */
  variant: ?
}

User Props {
  variant: 'ordered' | 'unordered'
}

Base Component {
  const { variant } = baseProps
  ...
  <User variant={variant} ... />
}
```

in this case, if we do a mistake in Base `variant` type and set it to `variant: 'ordered'` (this is hard to do because Base props usually are in storybook documentation) - User would have extra logic processing extra value. If we set it to `variant: 'ordered' | 'unordered' | 'fake'` in Base - TS will raise an error, that extra value can not be passed to User.

So, in those 4 variants it is pretty hard to make a human mistake, but even if we do - this mistake basically would bring us not a bug, but extra unnecessary code 🤷‍♂️ So I wouldn't say it's so harmful and also easy detectable

ozgurkececioglu: Thanks for the explanation. This is totally true, I agree.

So, this PR is actually about making a decision to not have enum-like custom types inside components anymore. Am I wrong?
And next step would be removing prop proxy(the ones I added recently and the others) and adding all types internally for Props interfaces, right? (even if there will be duplication like you did here ☝️)

denieler: @ozgurkececioglu not sure about all cases like that. I think we can still have some use cases when extracting type for prop might make sense

dmaklygin: I think it depends on the case, and we should implement it depending on that case. For example, if the Base component just  forwards `variant` prop and does nothing with variant inside itself, we can reuse `Variant` type in base:

```
// user.ts
type Variant = 'ordered' | 'unordered'
type UserProps = {
  variant: Variant
}
const User = ({variant}: UserProps) => {
}

// base.ts
import {Variant} from 'user.ts'
type BaseProps = {
  variant: Variant
}

const Base = ({variant}: BaseProps) => {
  <User variant={variant} />
}

```

1. If the user made a mistake in variant prop: when TS will raise an error, so the mistake is caught
2. If a new possible value was added to `Variant` type when nothing we should do in the Base component because we reuse `Variant`

dmaklygin: The case, when `variant` property is used inside `Base` component is more complex. In that case, I would use separated types as @denieler proposed

dmaklygin: So, I would rely on the second option you, @denieler suggested, but only then Base component forwards the property to children.

dmaklygin (approved):

packages/picasso/src/List/List.tsx
ozgurkececioglu: Ok got it. Thanks for the clarification guys 👍

augustobmoura (approved):

OndrejTuma (approved):

PR #2790, opened 2022-05-18
docs: add visual testing tooling RFC

separatio:
[FX-2534]

### Description

This is the RFC for visual tests in Picasso.
Should be self-explanatory.

### How to test

- review the doc

### Screenshots

### Development checks

- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2534]: https://toptal-core.atlassian.net/browse/FX-2534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+1. Puppeteer with Storybook
+2. Happo with Cypress
+
+## Proposal
+
+Move all visual tests to Cypress with Happo.
+
+The reasons for this are:
+- Picasso will support only one provider (Happo) and framework (Cypress) for visual tests
+- PRs will have only one visual tests check
```
denieler, line 15: I don't think this will be true, because we probably would still want to distinguish Cypress tests from Storybook tests

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+- Storybook's purpose in Picasso is not for doing visual tests
+- supporting two different plugins (Storybook and Happo) brings additional complexity and maintenance effort
```
denieler, line 17: we probably need to say that maintaining `jest-image-snapshot` (with visual test result files pushed to git) and Happo at the same time is complex. Also, we need to check visual screenshots using 2 different tools, which is confusing

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Problem
+
+Two different frameworks for visual testing are currently supported in Picasso:
+1. Puppeteer with Storybook
+2. Happo with Cypress
+
+## Proposal
+
+Move all visual tests to Cypress with Happo.
```
denieler, line 11: Maybe removing completely Storybook tests would not play that well for us, because Storybook is one of the outcomes, and products, which we propose to our users. We would like to make sure that our products look great. But at the same time, we don't want to have 2 different technologies. Maye the better way to go would be to do visual tests of Storybook via Happo? WDYT?

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+- it will take some time to migrate tests to Cypress
+- manual intervention in adding visual tests later
+- the Puppeteer stack (and tests) can't be removed immediately
+
+## Alternatives
+
+1. Simply replace the Puppeteer stack with the Happo provider for visual tests in Storybook:
+- this is the easy way out, but gives us less control and visibility over what's covered
+- maintenance over Cypress visual tests is still needed, as they are covering different areas
+- it requires supporting and maintaining two configurations for visual tests
```
denieler, line 36: maybe we can merge those two configurations and use the same Happo engine on both types of tests 🤔

docs/decisions/04-single-visual-tests-tool.md
separatio: For the current proposal, there are no Storybook tests. :slightly_smiling_face: 
Of course, if we have both, we'll get two checks in PRs.

docs/decisions/04-single-visual-tests-tool.md
separatio: I've considered it as well, hence the Alternatives section.
tl;dr It's going to be hard to take decisions what to test in which side if we do that.

These are some bad possible outcomes I am foreseeing:
- component is visually tested in both Storybook and Cypress, or at least they have overlapping areas being tested
- extending examples to increase visual test coverage instead of increasing test coverage via Cypress
- increased amount of time when deciding the "correct" way to visually test a component, bringing more discussions and opposing views, rather than unity

Yes, the product is showcased in Storybook. Hence, I am seeing Storybook as our Production environment. We shouldn't be testing on Production environments.

On the other hand, if we abandon Cypress visual tests, we'll end up with many potentially unnecessary examples in Storybook, bloating it and making it less usable for our end users.

docs/decisions/04-single-visual-tests-tool.md
separatio: We need to configure two separate plugins regardless. To my knowledge, the Storybook plugin is too different from the Cypress plugin to merge the configs. Check these for more information:

https://docs.happo.io/docs/cypress
https://docs.happo.io/docs/storybook

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
+# Single Visual Tests tool
+
+## Problem
+
+Two different frameworks for visual testing are currently supported in Picasso:
```
ozgurkececioglu, line 5: I'm not quite sure what is the problem 🙂 I mean it could be more clear

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+- the Puppeteer stack (and tests) can't be removed immediately
+
+## Alternatives
+
+1. Simply replace the Puppeteer stack with the Happo provider for visual tests in Storybook:
+- this is the easy way out, but gives us less control and visibility over what's covered
+- maintenance over Cypress visual tests is still needed, as they are covering different areas
+- it requires supporting and maintaining two configurations for visual tests
+
+2. A mix between Storybook and Cypress visual tests, both using the Happo provider plugins:
```
TomasSlama, line 38: I don't see the difference between the first alternative and this one. 🤔,

When writing cypress tests many times it seems we just copy-paste examples from stories. So in those cases, it would be great if we could just take a screenshot in the storybook.

docs/decisions/04-single-visual-tests-tool.md
dmaklygin: This is an existing plugin (https://docs.happo.io/docs/storybook) to use Happo with Storybook.
* Cypress will test only behavioral tests, meanwhile visual tests in Storybook will take a screenshot of the component state
* As far as I know, we are not going to increase the number of tests. Instead, we are willing to have a standard approach to writing/maintaining visual tests: keeping only Storybook+Happo will let us achieve this.
* I think, this is the most simple task, we will take a screenshot of each story in the Storybook, as we currently do

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+Move all visual tests to Cypress with Happo.
+
+The reasons for this are:
+- Picasso will support only one provider (Happo) and framework (Cypress) for visual tests
+- PRs will have only one visual tests check
+- Storybook's purpose in Picasso is not for doing visual tests
+- supporting two different plugins (Storybook and Happo) brings additional complexity and maintenance effort
+
+Considering this move, we will gain:
+- control over which visual tests we add
```
dmaklygin, line 20: Hmm, honestly, I would say the same for every single technology.

docs/decisions/04-single-visual-tests-tool.md
OleksandrNechai: Yeah, I guess real problem is Puppeteer being super slow and hence we have slow feedback from quality gate which slows down development and reduce developer's focus.

docs/decisions/04-single-visual-tests-tool.md
OleksandrNechai: As I understand, the difference is that in first option there is no place for Cypress tests.

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,48 @@
...
+### Drawbacks and limitations
+
+There are two drawbacks present to the above proposal:
+- it will take some time to migrate tests to Cypress
+- manual intervention in adding visual tests later
+- the Puppeteer stack (and tests) can't be removed immediately
+
+## Alternatives
+
+1. Simply replace the Puppeteer stack with the Happo provider for visual tests in Storybook:
```
OleksandrNechai, line 33: To me alternative (1) where we do not use Cypress simply is not possible. Take Dropdown for example. We need to open it and make screenshot. It can only be done in Cypress (maybe also in Storybook with some scripts, but is not it abusing Storybook?). Next, we need to test components composition. For example, Dropdown on Modal. Only Cypress for this.

docs/decisions/04-single-visual-tests-tool.md
OleksandrNechai: Tomas makes good point. If we go with Cypress only, we will have many code duplicates in stories and tests. Not sure if this is a problem though.

docs/decisions/04-single-visual-tests-tool.md
OleksandrNechai: >I am seeing Storybook as our Production environment. We shouldn't be testing on Production environments.

Right. Yet we test employs. And it makes total sense to test if what we potentially deploy looks reasonable. What if all components are rendered without styles? We will not notice it.

docs/decisions/04-single-visual-tests-tool.md
OleksandrNechai: I see only 2 real options: Cypress vs Cypress+Storybook. Now, is this `complexity and maintenance effort` so huge that we should not test Storybook to avoid it? Pleas, help us understand what are these efforts?

docs/decisions/04-single-visual-tests-tool.md
OleksandrNechai: I think the idea here is that we actively create Cypress tests instead of having all automatically covered in Storybook and tests kind appear against our active participation. Sounds like a good thing, we all learned hard way that `magic` feels good at first and then it is hard to maintain.

docs/decisions/04-single-visual-tests-tool.md
OndrejTuma: I agree, some summarization of headaches that we have with using these two frameworks together would be nice 🙂

separatio: Is the problem clearly stated now @ozgurkececioglu @OleksandrNechai @OndrejTuma ? Is there anything that might be a good addition to this section?

docs/decisions/04-single-visual-tests-tool.md
separatio: I understand your concern, @LashaJini . However, it still remains true that some of us and contributors are having headaches due to permissions in Linux, giving us headaches in running the tests, not to mention update screenshots.

Furthermore, we've already decided to go with Happo as the go-to visual tests provider.

denieler: yeah, I also agree that if we remove Storybook visual tests completely - we automatically add a huge maintenance effort of covering all those use-cases by Cypress tests. And the second - we might not be sure that deployed storybook looks correct

docs/decisions/04-single-visual-tests-tool.md
denieler: yeah, I've checked both pages and it looks like the effort pretty small - configs are just 1-2 liners looks like

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Problem
+
+Two different frameworks for visual testing are currently supported in Picasso:
+1. Puppeteer with Storybook
+2. Happo with Cypress
+
+The Puppeteer based stack has some downsides:
+- screenshots are hard to update
+- diffs inspection is not straightforward
```
denieler, line 11: idk, I never had issues with diffs inspections. I even like it more in GitHub rather than in Happo UI

separatio: I should've said local inspection, rather than on Github. :thinking:

docs/decisions/04-single-visual-tests-tool.md
separatio: While the effort is small, it's still true that we can't merge the configs, and we need two separate configurations.

docs/decisions/04-single-visual-tests-tool.md
separatio: It's possible via testing Storybook. But it still adds a layer of complexity. I've added the information in the RFC.

docs/decisions/04-single-visual-tests-tool.md
separatio: Yes, that is the idea. I've tried clarifying.

docs/decisions/04-single-visual-tests-tool.md
separatio: Yeah, this was confusing. I re-wrote this section.

And even if we copy-paste examples that we need testing, that's fine, as long as we _know_ what we are testing.
Right now, we are mostly assuming all examples are tested and we don't have any reporting for visual tests in Storybook, which is a downside.

OleksandrNechai (approved):

TomasSlama (approved):

docs/decisions/04-single-visual-tests-tool.md
```diff
@@ -0,0 +1,74 @@
...
+
+This section was added to clarify the options for Picasso with the Happo provider.
+The options are:
+- pure Storybook plugin, all visual tests being done in Storybook
+- mix between Storybook and Cypress plugins, visual test being split, most likely with two separate CI jobs
+- pure Cypress plugin, all visual tests being done in Cypress
+
+## Proposal
+
+Replace the Puppeteer stack with the Happo provider plugin for Storybook, while **keeping** Cypress tests.
```
denieler, line 29:
```suggestion
Replace the Puppeteer stack with the Happo provider plugin for Storybook, while **keeping** Cypress tests with Happo too.
```

denieler (approved):

PR #2642, opened 2022-04-14
chore: use davinci for unit tests

TomasSlama:
[FX-NNNN]

### Description

Use davinci-qa for unit testing

### How to test

- FIXME: Add the steps describing how to verify your changes

### Screenshots

| Before.                                 | After.                                  |
| --------------------------------------- | --------------------------------------- |
| Insert screenshots or screen recordings | Insert screenshots or screen recordings |

### Review

- [ ] Read [CONTRIBUTING.md](https://github.com/toptal/picasso/blob/master/CONTRIBUTING.md) and [Component API principles](https://github.com/toptal/picasso/blob/master/docs/contribution/component-api.md)
- [ ] Annotate all `props` in component with documentation
- [ ] Create `examples` for component
- [ ] Ensure that deployed demo has expected results and good examples
- [ ] Ensure that tests pass by running `yarn test`
- [ ] Ensure that visuals tests pass by running `yarn test:visual`. If not - check the documentation [how to fix visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md#fixing-broken-visual-tests-inside-a-pr)
- [ ] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).

<details>
<summary>PR commands</summary>
<br />

List of available commands:

- `@toptal-bot run all` - Run whole pipeline
- `@toptal-bot run build` - Check build
- `@toptal-bot run visual` - Run visual tests
- `@toptal-bot run deploy:documentation` - Deploy documentation
- `@toptal-bot run package:alpha-release` - Release alpha version

</details>

<details>
<summary>PR Review Guidelines</summary>
<br />

#### When to approve? ✅

**You are OK** with merging this PR and

1. You have no extra requests.
2. You have optional requests.
   1. Add `nit:` to your comment. (ex. `nit: I'd rename this variable from makeCircle to getCircle`)

#### When to request changes? ❌

**You are not OK** with merging this PR because

1. Something is broken after the changes.
2. Acceptance criteria is not reached.
3. Code is dirty.

#### When to comment (neither ✅ nor ❌)

**You want your comments to be addressed** before merging this PR in cases like:

1. There are leftovers like unnecessary logs, comments, etc.
2. You have an opinionated comment regarding the code that requires a discussion.
3. You have questions.

#### How to handle the comments?

1. An owner of a comment is the only one who can resolve it.
[...]

PR #2818, opened 2022-05-26
chore: use davinci for unit tests

TomasSlama:
[FX-2702]

### Description

This is the first part of using Davinci to run unit tests.

In this part, we use custom settings in the config, because it was the way how it was set up on the current master and we want to be this PR as small as possible.
```js
{
  resetMocks: false,
  restoreMocks: false
}
```

### How to test

- CI should be green

### Development checks

-  Add changeset according to [guidelines](https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md) (if needed)
- Ensure that deployed demo has expected results and good examples
- Ensure that visuals tests pass by running `yarn test:visual`. If not - check the documentation [how to fix visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md#fixing-broken-visual-tests-inside-a-pr)
- [x] Self reviewed
- Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2702]: https://toptal-core.atlassian.net/browse/FX-2702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/picasso/src/utils/Formatters/format-amount.test.ts
```diff
@@ -9,7 +9,9 @@ describe('formatAmount', () => {
 
   describe('when `locale` is `fr-FR`', () => {
     it('formats `fr-FR` locale', () => {
-      expect(formatAmount({ amount: 1500, locale: 'fr-FR' })).toBe('$1,500.00')
+      expect(formatAmount({ amount: 1500, locale: 'fr-FR' })).toBe(
+        '1 500,00 $US'
+      )
```
TomasSlama, line 14: image

This test file didn't match the tested pattern, so it was never tested before. I have checked manually and this is the correct output.

packages/picasso/src/RichTextEditor/hooks/useCounter/test.ts
```diff
@@ -30,7 +30,7 @@ describe('useCounter', () => {
 
     const { counterMessage } = result.current
 
-    expect(counterMessage).toContain(4)
+    expect(counterMessage).toContain('4')
```
TomasSlama, line 33: image

augustobmoura:
Should we change the test pattern? I thought only having tests with the name `test.ts` were done on purpose actually

TomasSlama:
@augustobmoura we can have a discussion about it, but since it is set up this way in Davinci, I think we are ok. Maybe a better approach could be to rename the test file so it is consistent with other test files.

jest.spec.js
```diff
@@ -0,0 +1,20 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
denieler, line 1: why this file's name is `.spec.js`?

jest.spec.js
```diff
@@ -0,0 +1,20 @@
...
+    '^@toptal/picasso/(.*)$': '<rootDir>/packages/picasso/src/$1',
+    '^@toptal/picasso-shared$': '<rootDir>packages/shared/src/index.ts',
+    '^@toptal/picasso$': '<rootDir>packages/picasso/src/index.ts',
+    '^@toptal/picasso-provider$':
+      '<rootDir>packages/picasso-provider/src/index.ts',
+    '^@toptal/picasso-root/(.*)$': '<rootDir>/$1'
+  },
+
+  resetMocks: false,
+  restoreMocks: false
```
denieler, line 19: we did not use them in the original config, why they are necessary now?

packages/picasso/src/utils/Formatters/test.ts
```diff
@@ -9,7 +9,9 @@ describe('formatAmount', () => {
 
   describe('when `locale` is `fr-FR`', () => {
     it('formats `fr-FR` locale', () => {
-      expect(formatAmount({ amount: 1500, locale: 'fr-FR' })).toBe('$1,500.00')
+      expect(formatAmount({ amount: 1500, locale: 'fr-FR' })).toBe(
+        '1 500,00 $US'
```
denieler, line 13: this doesn't look correct 😃 we probably need to change internalization settings for Jest instead

packages/picasso/src/utils/Formatters/format-amount.test.ts
denieler: why do we even test for `fr-FR` locale? 🤔 maybe we should replace it with `US`?

jest.spec.js
TomasSlama: if we use name `jest.config.js` davinci throws a warning message that it will not use custom config and ask FX team how to proceed ;-) So I have looked how staff portal does it and this is their solution.

jest.spec.js
TomasSlama: TBH I have no idea :D, this is even the default configuration by jest, so there had to be a reason to change it. I will try to find some information in confluence about it.

packages/picasso/src/utils/Formatters/test.ts
TomasSlama: @denieler 
image

I don't understand 🤷, it has same result in my browser.

TomasSlama: and even in CI

packages/picasso/src/utils/Formatters/format-amount.test.ts
TomasSlama: there is a test even for `US`, so I guess the purpose of the test was to test, that it passes the variable correctly

jest.spec.js
TomasSlama: It was added 2 years ago without any explanation ;-)

https://github.com/toptal/davinci/pull/318

TomasSlama: we set those two options back to [default value](https://jestjs.io/docs/configuration#restoremocks-boolean)

jest.spec.js
denieler: I think then we need to fix Davinci 😃

denieler (approved):

jest.spec.js
dmaklygin: I think we need to use from the default configuration. It will bring additional efforts to fix test cases. But anyway, they should be `true` by default

jest.spec.js
TomasSlama: @denieler I have created [ticket](https://toptal-core.atlassian.net/browse/FX-2820) for that.

jest.spec.js
TomasSlama: @dmaklygin this was the intention of this PR, to make it smaller than we had it in [previous PR](https://github.com/toptal/picasso/pull/2642).

In the next PR, I will enable these rules and update the tests accordingly. ( I will just copy-paste solution from old PR)

dmaklygin: got it, then OKay :)

augustobmoura (approved):
It looks like `yarn.lock` has been updated as well, can you commit it?
Aside from that, I run it on my side and it worked flawless! Thanks @TomasSlama :+1:

ozgurkececioglu (approved):

PR #2942, opened 2022-06-27
fix(RichTextEditor): use `FieldLabel`

TomasSlama:
[FX-2918]

### Description

Currently we don't use `FieldLabel` to render `label`, we just render it as plain text and it has different styling.

### How to test

- check [RichTextEditor story](https://picasso.toptal.net/fix-rich-text-label/?path=/story/picasso-forms-form--form#rich-text-editor) inside picasso-forms. The label should have the same styling as other inputs
- you can test it out in [codesandbox](https://codesandbox.io/s/currying-cherry-dkbuu7?file=/package.json) with aplha package

### Screenshots

| image | image |

### Development checks

- [x] Self reviewed
-  Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2918]: https://toptal-core.atlassian.net/browse/FX-2918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
+++ b/packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
@@ -8,6 +8,7 @@ export type Props = {
   name?: string
   label?: string
   required?: boolean
+  onClick?: () => void
 } & TextLabelProps
 
 const getRequiredDecoration = (
@@ -29,7 +30,7 @@ const getRequiredDecoration = (
 }
 
 const FieldLabel = (props: Props) => {
-  const { label, required, titleCase, name } = props
+  const { label, required, titleCase, name, onClick } = props
 
   const formConfig = useFormConfig()
   const requiredDecoration = getRequiredDecoration(
@@ -42,6 +43,7 @@ const FieldLabel = (props: Props) => {
       requiredDecoration={requiredDecoration}
       htmlFor={name}
       titleCase={titleCase}
+      onClick={onClick}
     >
       {label}
     </PicassoForm.Label>
--- a/packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
+++ b/packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
@@ -8,7 +8,6 @@ export type Props = {
   name?: string
   label?: string
   required?: boolean
-  onClick?: () => void
 } & TextLabelProps
 
 const getRequiredDecoration = (
@@ -30,7 +29,7 @@ const getRequiredDecoration = (
 }
 
 const FieldLabel = (props: Props) => {
-  const { label, required, titleCase, name, onClick } = props
+  const { label, required, titleCase, name } = props
 
   const formConfig = useFormConfig()
   const requiredDecoration = getRequiredDecoration(
@@ -43,7 +42,6 @@ const FieldLabel = (props: Props) => {
       requiredDecoration={requiredDecoration}
       htmlFor={name}
       titleCase={titleCase}
-      onClick={onClick}
     >
       {label}
     </PicassoForm.Label>
--- a/packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
+++ b/packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
@@ -8,6 +8,7 @@ import { Except } from 'type-fest'
 
 import { FieldProps } from '../FieldWrapper'
 import InputField from '../InputField'
+import FieldLabel from '../FieldLabel'
 
 type OverriddenProps = {
   defaultValue?: ASTType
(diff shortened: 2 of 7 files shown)


augustobmoura (commented):
One thing that I noticed, when you click on the label on other types of `Field` it focuses their input. But that doesn't happen with `RichTextEditor`. Do you think is something easy to make work?

Peek 2022-06-29 09-27

TomasSlama:

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -34,14 +37,37 @@ export const RichTextEditor = ({ onChange, defaultValue, ...rest }: Props) => {
     [onChange, setValue]
   )
 
+  const handleLabelClick = () => {
+    if (!editorRef) {
+      return
+    }
+
+    editorRef.current?.focus()
+  }
```
augustobmoura, line 46: NIT: we can simplify it to a single line, because editorRef will never be `null`
```suggestion
  const handleLabelClick = () => editorRef.current?.focus()
```

packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
```diff
@@ -8,6 +8,7 @@ export type Props = {
   name?: string
   label?: string
   required?: boolean
+  onClick?: () => void
```
denieler, line 11: do we really want a label to be clickable? is it a good UX?

diogolessa (commented):
I've manually checked and it looks good. Only thing I found is that when clicking on label of rich text with error state, it doesn't show red outline as others, but blue:
image

TomasSlama:
> I've manually checked and it looks good. Only thing I found is that when clicking on label of rich text with error state, it doesn't show red outline as others, but blue: image

great catch, thanks, we were fixing it for other inputs last week, I guess we forgot about RichTextEditor 👍 I will fix it, thanks.

packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
TomasSlama: @denieler as I understand it, it is native browser behavior of usual inputs when used like this:
```HTML
<label for='foo'>some label</label>
<input id='foo' />
```
you can test it yourself in this [codesandbox](https://codesandbox.io/s/priceless-poitras-7ibpby)
So I believe it is ok, to simulate this behavior for RichTextEditor because it is actually only `div`

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: ✅

packages/picasso-forms/src/FieldLabel/FieldLabel.tsx
denieler: @TomasSlama I agree that it should work for RichTextEditor, as you described, what I disagree with is that if we look at the `FieldLabel` component in isolation - we giving it an opportunity to become a clickable component. Which to me seems wrong. So maybe we can try to solve this in a different way 🤔 maybe leveraging `for/id` from HTML if it would work

TomasSlama: @denieler I was trying this approach, but because `RichTextEditor` is `div`, this native behavior does not work.

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -232,11 +235,23 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
...
         )}
+        {hiddenInputId && enableFocusOnLabelClick(hiddenInputId)}
       </Container>
     )
   }
 )
 
+const hiddenInputStyle: React.CSSProperties = {
+  position: 'absolute',
+  opacity: 0,
```
denieler, line 246: nit:
```suggestion
  visibility: hidden,
```
would it work?

denieler (approved):
The code and approach look good 👍 Haven't tested them. Maybe I would add also a cypress tests to check this functionality working

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
TomasSlama: @denieler nope, neither `visiblity: hidden` or `display: none` works. Browser is too clever to focus not visible element ;-)

ozgurkececioglu (approved):
Tested on temploy. Looking awesome, good job 💪

diogolessa (approved):
Thanks for taking care of it 👍🏻

PR #3008, opened 2022-07-21
fix(TagSelector): fix using TagSelector with initialValue

ozgurkececioglu:
[FX-2913]

### Description

Due to the default prop value of initialValue, it wasn't possible to set initialValues for TagSelector field.

### How to test

- use Form's default story to see TagSelector working fine
- bug was visible after creating a new app from `davinci new app`

### Screenshots

From newly created davinci app
Screen Shot 2022-07-21 at 09 19 34
Screen Shot 2022-07-21 at 09 17 47

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2913]: https://toptal-core.atlassian.net/browse/FX-2913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso-forms/src/Form/story/Default.example.tsx
+++ b/packages/picasso-forms/src/Form/story/Default.example.tsx
@@ -37,6 +37,11 @@ const filterOptions = (str = '', options: Item[] = []): Item[] | null => {
   return result.length > 0 ? result : null
 }
 
+const initialValues = {
+  'default-gender': 'female',
+  'default-skills': [skills[0]],
+}
+
 const Example = () => {
   const [skillInputValue, setSkillInputValue] =
     useState<string>(EMPTY_INPUT_VALUE)
@@ -52,7 +57,7 @@ const Example = () => {
     <Form
       autoComplete='off'
       onSubmit={values => window.alert(JSON.stringify(values, undefined, 2))}
-      initialValues={{ 'default-gender': 'female' }}
+      initialValues={initialValues}
     >
       <Form.Input
         enableReset
--- a/packages/picasso-forms/src/Form/story/Default.example.tsx
+++ b/packages/picasso-forms/src/Form/story/Default.example.tsx
@@ -39,7 +39,6 @@ const filterOptions = (str = '', options: Item[] = []): Item[] | null => {
 
 const initialValues = {
   'default-gender': 'female',
-  'default-skills': [skills[0]],
 }
 
 const Example = () => {
--- a/packages/picasso-forms/src/Form/test.tsx
+++ b/packages/picasso-forms/src/Form/test.tsx
@@ -55,6 +55,37 @@ const renderForm = (
   )
 }
 
+interface FormData {
+  firstName: string
+  skills: any[]
+}
+
+const skillOptions = [
+  { value: '0', text: 'HTML' },
+  { value: '1', text: 'CSS' },
+  { value: '2', text: 'Javascript' },
+]
+
+const initialValues: FormData = {
+  firstName: 'Bruce',
+  skills: [skillOptions[0]],
+}
+
+const renderFormWithInitialValues = (onSubmit: (values: FormData) => void) => {
+  return render(
+    <Form onSubmit={values => onSubmit(values)} initialValues={initialValues}>
+      <Form.Input name='firstName' placeholder='test input' />
+      <Form.TagSelector
+        name='skills'
+        label='Skills'
+        options={skillOptions}
+        inputValue=''
+      />
+      <Button type='submit'>Submit</Button>
+    </Form>
+  )
+}
+
 const scrollToMock = scrollTo as jest.Mock
 
 describe('Form', () => {
@@ -153,6 +184,20 @@ describe('Form', () => {
           expect(getByTestId('valid-icon')).toBeInTheDocument()
         })
       })
+    })
+  })
+
+  describe('when initial values provided', () => {
+    it('fills up the fields with provided value', async () => {
+      const onSubmit = jest.fn()
+
+      const { getByText } = renderFormWithInitialValues(onSubmit)
+
+      await act(() => {
+        fireEvent.click(getByText('Submit'))
+      })
+
+      expect(onSubmit).toHaveBeenCalledWith(initialValues)
     })
   })
 })
--- a/packages/picasso-forms/src/Form/test.tsx
+++ b/packages/picasso-forms/src/Form/test.tsx
@@ -56,8 +56,7 @@ const renderForm = (
 }
 
 interface FormData {
-  firstName: string
-  skills: any[]
+  skills: { value: string; text: string }[]
 }
 
 const skillOptions = [
@@ -67,14 +66,14 @@ const skillOptions = [
 ]
 
 const initialValues: FormData = {
-  firstName: 'Bruce',
   skills: [skillOptions[0]],
 }
 
-const renderFormWithInitialValues = (onSubmit: (values: FormData) => void) => {
+const renderTagSelectorWithInitialValue = (
+  onSubmit: (values: FormData) => void
+) => {
   return render(
     <Form onSubmit={values => onSubmit(values)} initialValues={initialValues}>
-      <Form.Input name='firstName' placeholder='test input' />
       <Form.TagSelector
         name='skills'
         label='Skills'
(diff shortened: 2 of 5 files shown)


ozgurkececioglu:

packages/picasso-forms/src/Form/story/Default.example.tsx
```diff
@@ -52,7 +57,7 @@ const Example = () => {
     <Form
       autoComplete='off'
       onSubmit={values => window.alert(JSON.stringify(values, undefined, 2))}
-      initialValues={{ 'default-gender': 'female' }}
+      initialValues={initialValues}
```
ozgurkececioglu, line 60: I changed this line to not courage users to follow the old approach. It should be a static reference to pass to this prop because otherwise, nested objects/arrays are not working properly. For this PR's case, it is TagSelector.

packages/picasso/src/TagSelector/TagSelector.tsx
```diff
@@ -133,7 +133,7 @@ export const TagSelector = forwardRef<HTMLInputElement, Props>(
       renderLabel: customRenderLabel,
       renderOption,
       showOtherOption,
-      value: values = [],
+      value,
```
ozgurkececioglu, line 136: The old version is not covering `empty string case`.  That's why I added another line to make sure that `values` is array now.

denieler: to be honest, an empty string shouldn't be a valid value type to pass to this component, so I would not change implementation because of that

packages/picasso-forms/src/Form/story/Default.example.tsx
```diff
@@ -37,6 +37,11 @@ const filterOptions = (str = '', options: Item[] = []): Item[] | null => {
   return result.length > 0 ? result : null
 }
 
+const initialValues = {
+  'default-gender': 'female',
+  'default-skills': [skills[0]],
```
denieler, line 42: instead of adding this value here, let's add a test that will be checking this functionality. Probably unit test should be enough 🤔

packages/picasso/src/TagSelector/TagSelector.tsx
ozgurkececioglu: We can also solve this problem on `Form.TagSelector` level. Let me do that 👍

packages/picasso/src/TagSelector/TagSelector.tsx
```diff
@@ -143,6 +143,9 @@ export const TagSelector = forwardRef<HTMLInputElement, Props>(
       ...rest
     } = props
 
+    // being sure that values will be an array
+    const values = value ? value : []
```
dmaklygin, line 147: One moment. According to the prop definition, it might be either `Item[]` or `undefined` 
How it's possible to pass an empty string? It must be restricted by static analyzer, no?

ozgurkececioglu: As @denieler mentioned https://github.com/toptal/picasso/pull/3008#discussion_r926319721, I can handle this conversion outside of TagSelector. So, nothing will change here.

packages/picasso-forms/src/Form/test.tsx
```diff
@@ -55,6 +55,37 @@ const renderForm = (
...
+  { value: '1', text: 'CSS' },
+  { value: '2', text: 'Javascript' },
+]
+
+const initialValues: FormData = {
+  firstName: 'Bruce',
+  skills: [skillOptions[0]],
+}
+
+const renderFormWithInitialValues = (onSubmit: (values: FormData) => void) => {
```
denieler, line 74: I would specify that it's form with TagSelector specifically, because for us it's important

packages/picasso-forms/src/Form/test.tsx
```diff
@@ -155,4 +186,18 @@ describe('Form', () => {
       })
     })
   })
+
+  describe('when initial values provided', () => {
+    it('fills up the fields with provided value', async () => {
```
denieler, line 191: same here, I would specify that we test specifically TagSelector in this case

packages/picasso-forms/src/TagSelector/TagSelector.tsx
```diff
@@ -28,16 +28,16 @@ export const TagSelector = (props: Props) => {
       }
     >
       {(inputProps: TagSelectorProps) => {
-        return <PicassoTagSelector {...inputProps} />
+        const { value, ...restOfInputProps } = inputProps
+        // avoid passing empty string to TagSelector
+        const valueAsArray = value ? value : []
```
denieler, line 33: idk, to me it seems it's not worth doing this change. Instead we need to improve type-safety for such a component 🤔

denieler (approved):
Code looks good 👍 tested this change, it works as expected 👍

packages/picasso-forms/src/TagSelector/TagSelector.tsx
```diff
@@ -28,16 +28,16 @@ export const TagSelector = (props: Props) => {
...
+
+        return <PicassoTagSelector {...restOfInputProps} value={valueAsArray} />
```
dmaklygin, line 35: I'm just wondering. the `Form.TagSelector` inherits the types from `TagSelectorProps` and according to `TagSelectorProps`, the `value` must be either `Item[]` or `undefined`. 
Why do we expect here an empty string?

packages/picasso-forms/src/TagSelector/TagSelector.tsx
ozgurkececioglu: `final-form` provides an empty string for all default values instead of `undefined` to avoid seeing errors like `don't convert uncontrolled inputs to controlled inputs` etc. We can provide allowNull and pass null for non-existing values. But, TagSelector itself doesn't expect a value except an empty array.

packages/picasso-forms/src/TagSelector/TagSelector.tsx
dmaklygin: If `inputProps.value` might be an **empty** string, then we also should expect the same behavior for `Form.Select`, coz it also might expect value as an array: https://github.com/toptal/picasso/blob/2d76bbf2e1b565829145f3fd338ff86c9d198f6b/packages/picasso/src/SelectBase/types.ts#L71

packages/picasso-forms/src/TagSelector/TagSelector.tsx
dmaklygin: Got it. 
I would be more specific here, to compare `value` with an empty string:
```
value === '' ? value : []
```
to highlight the reason why we do this
wdyt?

packages/picasso-forms/src/TagSelector/TagSelector.tsx
ozgurkececioglu: I tried to explain here: https://github.com/toptal/picasso/pull/3008#discussion_r926446167

We already expect empty string for `Select` here https://github.com//toptal/picasso/blob/2d76bbf2e1b565829145f3fd338ff86c9d198f6b/packages/picasso/src/NonNativeSelect/NonNativeSelect.tsx#L60 so I don't think that's the same issue 🙂

packages/picasso-forms/src/TagSelector/TagSelector.tsx
OndrejTuma: Wow so this was the problem all along 😱 great that you figured it out 💪

OndrejTuma: what about using `Array.isArray(value)`

ozgurkececioglu: In this case, TS will argue that value can't be a string 😄

packages/picasso-forms/src/TagSelector/TagSelector.tsx
dmaklygin: if it's covered, okay then! Thanks!

dmaklygin (approved):
Checked the code, looks good!

packages/picasso-forms/src/TagSelector/TagSelector.tsx
ozgurkececioglu: That can be better since TS will argue about
```
value === '' ? value : []
```
because value can't be a string

TomasSlama (approved):
Looks good

ozgurkececioglu:

separatio (approved):
Tested. :+1:

PR #3041, opened 2022-08-11
feat(AvatarUpload): implement AvatarUpload

ozgurkececioglu:
[FX-2889]

### Description

- implemented new component AvatarDropzone
- I decided to skip the title and file description parts for now. They can be introduced in another new component if needed.
- I solved the `focused` state differently than the designs to simplify the implementation.

### How to test

- you can use temploy to test the component. `Default` story is mimicking the uploading UX

### Screenshots

Screen Shot 2022-08-11 at 15 15 19
Screen Shot 2022-08-11 at 15 15 33
Screen Shot 2022-08-11 at 15 15 37

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2889]: https://toptal-core.atlassian.net/browse/FX-2889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/blue-eels-compete.md
+++ b/.changeset/blue-eels-compete.md
@@ -1,0 +1,9 @@
+---
+'@toptal/picasso': minor
+---
+
+---
+
+### AvatarUpload
+
+- add new component for avatar uploads
--- a/cypress/component/AvatarUpload.spec.tsx
+++ b/cypress/component/AvatarUpload.spec.tsx
@@ -1,0 +1,66 @@
+import React from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const component = 'AvatarUpload'
+
+describe('AvatarUpload', () => {
+  describe('when warning message provided', () => {
+    it('renders warning icon and tooltip', () => {
+      cy.mount(
+        <Container padded='small'>
+          <AvatarUpload
+            warningMessage={
+              <span data-testid='warning-content'>File is too big</span>
+            }
+            testIds={{
+              imageAvatar: 'image-avatar',
+              warningIcon: 'warning-icon',
+            }}
+          />
+        </Container>
+      )
+
+      cy.getByTestId('warning-icon').realHover()
+      cy.getByTestId('warning-content').should('be.visible')
+
+      cy.get('body').happoScreenshot({
+        component,
+        variant: 'with-warning-message/after-hover',
+      })
+    })
+  })
+
+  describe('when source file exists and component is hovered', () => {
+    let src = ''
+
+    beforeEach(() => {
+      // eslint-disable-next-line max-nested-callbacks, promise/catch-or-return
+      cy.fixture('pablo.jpg').then(image => {
+        src = 'data:image/jpg;base64,' + image
+
+        return image
+      })
+    })
+
+    it('renders upload icon over image avatar', () => {
+      cy.mount(
+        <Container padded='small'>
+          <AvatarUpload
+            src={src}
+            hovered
+            testIds={{
+              uploadIcon: 'upload-icon',
+            }}
+          />
+        </Container>
+      )
+
+      cy.getByTestId('upload-icon').should('be.visible')
+
+      cy.get('body').happoScreenshot({
+        component,
+        variant: 'with-source-file/after-hover',
+      })
+    })
+  })
+})
--- a/packages/picasso/src/AvatarUpload/AvatarUpload.tsx
+++ b/packages/picasso/src/AvatarUpload/AvatarUpload.tsx
@@ -1,0 +1,237 @@
+import React, { forwardRef, ReactNode, useCallback } from 'react'
+import { makeStyles, Theme } from '@material-ui/core'
+import { BaseProps, SizeType } from '@toptal/picasso-shared'
+import cx from 'classnames'
+import { useDropzone } from 'react-dropzone'
+
+import Container from '../Container'
+import { ExclamationSolid16, Upload24 } from '../Icon'
+import styles from './styles'
+import { AvatarUploadOptions, DropEvent, FileRejection } from './types'
+import Loader from '../Loader'
+import AvatarWrapper from '../Avatar/AvatarWrapper/AvatarWrapper'
+import ImageAvatar from '../Avatar/ImageAvatar/ImageAvatar'
+import Tooltip from '../Tooltip'
+import useAvatarUpload from './use-avatar-upload'
+import { Status } from '../OutlinedInput'
+
+export interface Props extends BaseProps {
+  /**
+   * Set accepted file types. See https://github.com/okonet/attr-accept for more information.
+   */
+  accept?: AvatarUploadOptions['accept']
+  /** Alt text */
+  alt?: string
+  /** Image URL */
+  src?: string
+  /** Size of the avatar */
+  size?: SizeType<'small' | 'large'>
+  /** Enable/disable the dropzone */
+  disabled?: boolean
+  /** Maximum file size (in bytes) */
+  maxSize?: number
+  /** Minimum file size (in bytes) */
+  minSize?: number
+  /**
+   * Callback for when the drop event occurs. Note that if file is not accepted, this callback is not invoked.,
+   * @type <T extends File>(files: T, event: DropEvent) => void
+   */
+  onDropAccepted?: AvatarUploadOptions['onDropAccepted']
+  /**
+   * Callback for when the drop event occurs. Note that if file is not rejected, this callback is not invoked.
+   * @type (fileRejection: FileRejection, event: DropEvent) => void
+   */
+  onDropRejected?: AvatarUploadOptions['onDropRejected']
+  /**
+   * Callback for when the drop event occurs. Note that the onDrop callback will always be invoked regardless if the dropped file was accepted or rejected.
+   * @type <T extends File>(acceptedFile: T | null, fileRejection: FileRejection | null, event: DropEvent) => void
+   */
+  onDrop?: AvatarUploadOptions['onDrop']
+  /**
+   * Custom validation function
+   * (file: File) => FileError | FileError[] | null
+   */
+  validator?: AvatarUploadOptions['validator']
+  /** Warning message to be used as a tooltip */
+  warningMessage?: ReactNode
+  /** Indicate `AvatarUpload` is in `error` or `default` state */
+  status?: Extract<Status, 'error' | 'default'>
+  /** Indicate whether the selected file is being uploaded */
+  uploading?: boolean
+  /** Indicate whether component has focused state */
+  focused?: boolean
+  /** Indicate whether component has hovered state */
+  hovered?: boolean
+  testIds?: {
+    imageAvatar?: string
+    warningIcon?: string
+    uploadIcon?: string
+    loader?: string
+  }
+}
+
+const useStyles = makeStyles<Theme, Props>(styles, {
+  name: 'PicassoAvatarUpload',
+})
+
+export const AvatarUpload = forwardRef<HTMLInputElement, Props>(
+  function AvatarUpload(props, ref) {
+    const {
+      className,
+      style,
+      status,
+      size = 'small',
+      uploading,
+      src,
+      alt,
+      warningMessage,
+      testIds,
+      focused: initiallyFocused,
+      hovered: initiallyHovered,
+      'data-testid': dataTestId,
+
+      // dropzoneOptions
+      accept,
+      minSize,
+      maxSize,
+      disabled,
+      onDrop,
+      onDropAccepted,
+      onDropRejected,
+      validator,
+    } = props
+
+    // callback overrides to return only one file to the parent component
+    const handleDrop = useCallback(
+      (
+        acceptedFiles: File[],
+        fileRejections: FileRejection[],
+        event: DropEvent
+      ) => {
+        if (onDrop) {
+          onDrop(acceptedFiles[0] ?? null, fileRejections[0] ?? null, event)
+        }
+      },
+      [onDrop]
+    )
+
+    const handleDropAccepted = useCallback(
+      (files: File[], event: DropEvent) => {
+        if (onDropAccepted) {
+          onDropAccepted(files[0], event)
+        }
+      },
+      [onDropAccepted]
+    )
+
+    const handleDropRejected = useCallback(
+      (fileRejections: FileRejection[], event: DropEvent) => {
+        if (onDropRejected) {
+          onDropRejected(fileRejections[0], event)
+        }
+      },
+      [onDropRejected]
+    )
+
+    const error = status === 'error'
+    const { getRootProps, isDragActive, getInputProps } = useDropzone({
+      accept,
+      minSize,
+      maxSize,
+      multiple: false,
+      disabled,
+      onDrop: handleDrop,
+      onDropAccepted: handleDropAccepted,
+      onDropRejected: handleDropRejected,
+      validator,
+    })
+
+    const {
+      callbacks,
+      focused,
+      hovered,
+      showImageAvatar,
+      showUploadIcon,
+      showLoadingIcon,
+      sourceFileExist,
+    } = useAvatarUpload({
+      focused: initiallyFocused,
+      hovered: initiallyHovered,
+      src,
+      isDragActive,
+      uploading,
+    })
+
+    const classes = useStyles({ ...props, focused })
+
+    const uploadedImageAvatar = showImageAvatar && (
+      <ImageAvatar
+        className={classes.imageAvatar}
+        size={size}
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        src={src!}
+        alt={alt}
+        data-testid={testIds?.imageAvatar}
+      />
+    )
+
+    const loadingIcon = showLoadingIcon && (
+      <Loader size='small' variant='inherit' data-testid={testIds?.loader} />
+    )
+    const uploadIcon = showUploadIcon && (
+      <Upload24 data-testid={testIds?.uploadIcon} />
+    )
+
+    const warningTooltip = warningMessage && (
+      <Tooltip content={warningMessage}>
+        <div className={classes.warningIcon}>
+          <ExclamationSolid16 data-testid={testIds?.warningIcon} />
+        </div>
+      </Tooltip>
+    )
+
+    return (
+      <Container
+        style={style}
+        ref={ref}
+        className={cx(classes.root, classes.size, className)}
+      >
+        <AvatarWrapper variant='square' size={size}>
+          <Container
+            {...getRootProps({
+              className: cx(classes.dropzone, classes.size, classes.corner, {
+                [classes.dragActive]: isDragActive,
+                [classes.disabled]: disabled,
+                [classes.error]: error,
+                [classes.focused]: focused,
+                [classes.reupload]: sourceFileExist,
+                [classes.hovered]: hovered,
+              }),
+            })}
+            {...callbacks}
+            data-testid={dataTestId}
+          >
+            <input {...getInputProps()} />
+            {loadingIcon}
+            {uploadIcon}
+          </Container>
+          {uploadedImageAvatar}
+          {!showImageAvatar && <div className={classes.leftBottomCorner} />}
+        </AvatarWrapper>
+        {warningTooltip}
+      </Container>
+    )
+  }
+)
+
+AvatarUpload.displayName = 'AvatarUpload'
+
+AvatarUpload.defaultProps = {
+  size: 'small',
+  disabled: false,
+  maxSize: Infinity,
+  minSize: 0,
+  accept: 'image/*',
+}
+
+export default AvatarUpload
--- a/packages/picasso/src/AvatarUpload/story/Default.example.tsx
+++ b/packages/picasso/src/AvatarUpload/story/Default.example.tsx
@@ -1,0 +1,12 @@
+import React from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const Example = () => {
+  return (
+    <Container padded='medium'>
+      <AvatarUpload />
+    </Container>
+  )
+}
+
+export default Example
--- a/packages/picasso/src/AvatarUpload/story/Sizes.example.tsx
+++ b/packages/picasso/src/AvatarUpload/story/Sizes.example.tsx
@@ -1,0 +1,20 @@
+import React from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const Example = () => (
+  <Container flex gap='medium' padded='medium'>
+    <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' />
+    <AvatarUpload
+      alt='Jacqueline Roque. Pablo Picasso, 1954'
+      src='./jacqueline-with-flowers-1954-square.jpg'
+    />
+    <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' size='large' />
+    <AvatarUpload
+      alt='Jacqueline Roque. Pablo Picasso, 1954'
+      size='large'
+      src='./jacqueline-with-flowers-1954-square.jpg'
+    />
+  </Container>
+)
+
+export default Example
--- a/packages/picasso/src/AvatarUpload/story/States.example.tsx
+++ b/packages/picasso/src/AvatarUpload/story/States.example.tsx
@@ -1,0 +1,43 @@
+import React from 'react'
+import { AvatarUpload, Container, Typography } from '@toptal/picasso'
+
+const Example = () => (
+  <Container flex padded='medium' gap='medium'>
+    <Container flex direction='column' alignItems='center'>
+      <Typography variant='heading' size='small'>
+        Focused
+      </Typography>
+      <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' focused />
+    </Container>
+
+    <Container flex direction='column' alignItems='center'>
+      <Typography variant='heading' size='small'>
+        Error
+      </Typography>
+      <AvatarUpload
+        alt='Jacqueline Roque. Pablo Picasso, 1954'
+        status='error'
+      />
+    </Container>
+
+    <Container flex direction='column' alignItems='center'>
+      <Typography variant='heading' size='small'>
+        Focused & Error
+      </Typography>
+      <AvatarUpload
+        alt='Jacqueline Roque. Pablo Picasso, 1954'
+        status='error'
+        focused
+      />
+    </Container>
+
+    <Container flex direction='column' alignItems='center'>
+      <Typography variant='heading' size='small'>
+        Loading
+      </Typography>
+      <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' uploading />
+    </Container>
+  </Container>
+)
+
+export default Example
--- a/packages/picasso/src/AvatarUpload/story/Warning.example.tsx
+++ b/packages/picasso/src/AvatarUpload/story/Warning.example.tsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const Example = () => (
+  <Container flex padded='medium' gap='medium'>
+    <AvatarUpload
+      alt='Jacqueline Roque. Pablo Picasso, 1954'
+      warningMessage='This is a warning message'
+    />
+    <AvatarUpload
+      alt='Jacqueline Roque. Pablo Picasso, 1954'
+      size='large'
+      warningMessage='This is a warning message'
+    />
+  </Container>
+)
+
+export default Example
--- a/packages/picasso/src/AvatarUpload/story/index.jsx
+++ b/packages/picasso/src/AvatarUpload/story/index.jsx
@@ -1,0 +1,30 @@
+import PicassoBook from '~/.storybook/components/PicassoBook'
+import { AvatarUpload } from '../AvatarUpload'
+
+const page = PicassoBook.section('Components').createPage(
+  'AvatarUpload',
+  'Gets the image from file input and displays it in the Avatar component.'
+)
+
+page.createTabChapter('Props').addComponentDocs({
+  component: AvatarUpload,
+  name: 'AvatarUpload',
+})
+
+page
+  .createChapter()
+  .addExample('AvatarUpload/story/Default.example.tsx', {
+    title: 'Default',
+    takeScreenshot: false,
+  })
+  .addExample('AvatarUpload/story/WithUpload.example.tsx', {
+    title: 'With Upload',
+    takeScreenshot: false,
+    description: 'Simulates upload of an image on drop',
+  })
+  .addExample('AvatarUpload/story/Sizes.example.tsx', 'Sizes')
+  .addExample('AvatarUpload/story/Warning.example.tsx', {
+    title: 'Warning',
+    description: 'Hover warning icon to see the message',
+  })
+  .addExample('AvatarUpload/story/States.example.tsx', 'States')
--- a/packages/picasso/src/AvatarUpload/styles.ts
+++ b/packages/picasso/src/AvatarUpload/styles.ts
@@ -1,0 +1,167 @@
+import { alpha, SizeType } from '@toptal/picasso-shared'
+import { createStyles, Theme } from '@material-ui/core/styles'
+
+import { Status } from '../OutlinedInput'
+
+const getHypotenuseOfEqualSides = (sideLength: number) => {
+  return sideLength * Math.sqrt(2)
+}
+
+const smallCornerSize = 1
+const largeCornerSize = 1.5
+
+const SETTINGS = {
+  small: {
+    dimensions: 5,
+    cornerSize: smallCornerSize,
+    hypotenuse: getHypotenuseOfEqualSides(smallCornerSize),
+  },
+  large: {
+    dimensions: 10,
+    cornerSize: largeCornerSize,
+    hypotenuse: getHypotenuseOfEqualSides(largeCornerSize),
+  },
+} as const
+
+interface Props {
+  size?: SizeType<'small' | 'large'>
+  src?: string
+  status?: Extract<Status, 'error' | 'default'>
+  focused?: boolean
+}
+
+export default ({ palette, sizes, transitions }: Theme) => {
+  const dragActiveColor = alpha(palette.blue.main, 0.16)
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const hoverColor = alpha(palette.grey.darker!, 0.7)
+
+  return createStyles({
+    root: {
+      position: 'relative',
+    },
+    dropzone: {
+      display: 'flex',
+      justifyContent: 'center',
+      alignItems: 'center',
+      position: 'relative',
+      backgroundColor: palette.blue.lighter,
+      fontSize: '1rem',
+      flexShrink: 0,
+      flexGrow: 0,
+      zIndex: 1,
+      outline: 'none',
+      borderWidth: sizes.borderWidth,
+      borderColor: palette.blue.main,
+      color: palette.blue.main,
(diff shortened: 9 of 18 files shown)


packages/picasso/src/AvatarDropzone/styles.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      return {
+        clipPath,
+        // we can remove this prefix as soon as this issue will
+        // be resolved - https://github.com/cssinjs/css-vendor/issues/74
+        '-webkit-clip-path': clipPath,
+      }
+    },
+
+    leftBottomCorner: ({ size = 'small', error }: Props) => {
```
ozgurkececioglu, line 98: I had to provide a different class and additional `div` because pseudo-classes are not working with functions.

.changeset/blue-eels-compete.md
```diff
@@ -0,0 +1,8 @@
+---
+'@toptal/picasso': patch
+---
+
+---
+### AvatarDropzone
```
denieler, line 6: in [Figma](https://www.figma.com/file/5SCTOPrCDcHuk5We091GBn/Product-Library?node-id=3886%3A16524) the name of the component is `AvatarUpload`. Because we have to have common design language I think we should have the same name for the component

ozgurkececioglu: Yeah, I wanted to provide Dropzone first and then I will create `AvatarUpload` and use `AvatarDropzone` inside with other text component parts (heading and file upload description). Does it makes sense 🤔

ozgurkececioglu:
@toptal-bot run all

packages/picasso/src/AvatarDropzone/styles.ts
```diff
@@ -0,0 +1,128 @@
...
+    reupload: {},
+
+    imageAvatar: {
+      zIndex: 0,
+    },
+
+    size: ({ size = 'small' }: Props) => {
+      const { dimensions } = SETTINGS[size]
+      const widthRatio = 1
+      const heightRatio = 1
```
OndrejTuma, line 78: I wonder if this is a leftover or not?

packages/picasso/src/AvatarDropzone/AvatarDropzone.tsx
```diff
@@ -0,0 +1,189 @@
...
+          <Container
+            flex
+            justifyContent='center'
+            alignItems='center'
+            {...getRootProps({
+              className: cx(classes.dropzone, classes.size, classes.corner, {
+                [classes.dragActive]: isDragActive,
+                [classes.disabled]: disabled,
+                [classes.error]: error,
+                [classes.reupload]: !!src,
```
OndrejTuma, line 164: I think the re-upload visual should be applied on hover only? 🤔
I'm not 100% sure though

packages/picasso/src/AvatarDropzone/AvatarDropzone.tsx
```diff
@@ -0,0 +1,189 @@
...
+  disabled?: boolean
+  /** Maximum file size (in bytes) */
+  maxSize?: number
+  /** Minimum file size (in bytes) */
+  minSize?: number
+  /** Callback for when the drop event occurs */
+  onDropAccepted?: AvatarDropzoneOptions['onDropAccepted']
+  /** Callback for when the drop event occurs */
+  onDropRejected?: AvatarDropzoneOptions['onDropRejected']
+  /** Callback for when the drop event occurs */
```
OndrejTuma, line 33: Those three callbacks have the same description 🙂

ozgurkececioglu: Oh, I forgot to add additional docs for the story file. Thanks for the reminder 🙇‍♂️

packages/picasso/src/AvatarDropzone/AvatarDropzone.tsx
ozgurkececioglu: Designs are not clear about this. I asked the same question to Matt here https://www.figma.com/file/q2nvjiyO2CLqBv4DeJnU3i?node-id=22:676#246707809

packages/picasso/src/AvatarDropzone/styles.ts
ozgurkececioglu: I copied the code from dropzone and these should be cleaned up. Thank you 🙇‍♂️

.changeset/blue-eels-compete.md
ozgurkececioglu: I will rename the component and all usages to fit with the same name in the designs. If we decide to extend the component to have title and description parts, we can provide another component for it.

packages/picasso/src/AvatarDropzone/AvatarDropzone.tsx
OndrejTuma: I'm also missing a focused state here - I'm not sure how tricky it can be 🤞

ozgurkececioglu:
@toptal-anvil ping reviewers

jg:
@toptal-anvil ping reviewers

.changeset/blue-eels-compete.md
```diff
@@ -0,0 +1,8 @@
+---
+'@toptal/picasso': patch
```
TomasSlama, line 2: I think we are adding new feature, so this should be probably minor bump 🤔

packages/picasso/src/AvatarUpload/story/Sizes.example.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' size='large' />
+    </Container>
+    <Container padded='medium'>
+      <AvatarUpload
+        alt='Jacqueline Roque. Pablo Picasso, 1954'
+        size='large'
+        src='./jacqueline-with-flowers-1954-square.jpg'
+      />
+    </Container>
+  </Container>
```
TomasSlama, line 25:
```suggestion
  <Container flex gap='medium'>
    <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' />
    <AvatarUpload
      alt='Jacqueline Roque. Pablo Picasso, 1954'
      src='./jacqueline-with-flowers-1954-square.jpg'
    />
   <AvatarUpload alt='Jacqueline Roque. Pablo Picasso, 1954' size='large' />
   <AvatarUpload
     alt='Jacqueline Roque. Pablo Picasso, 1954'
     size='large'
     src='./jacqueline-with-flowers-1954-square.jpg'
    />
  </Container>
```

packages/picasso/src/AvatarUpload/story/States.example.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { AvatarUpload, Container, Typography } from '@toptal/picasso'
+
+const Example = () => (
+  <Container flex direction='column'>
+    <Typography variant='heading' size='small'>
+      Focused
+    </Typography>
+    <Container flex>
```
TomasSlama, line 9: I would also use here `<Container flex gap="medium">` so we don't need the inner containers.

packages/picasso/src/AvatarUpload/story/Warning.example.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const Example = () => (
+  <Container flex>
```
TomasSlama, line 5: also `gap="medium"`

packages/picasso/src/AvatarUpload/styles.ts
```diff
@@ -0,0 +1,131 @@
...
+  small: {
+    dimensions: 5,
+    cornerSize: 1,
+    hipotenusa: Math.sqrt(2) * 1,
+    cornerCoordinate: 0.65,
+  },
+  large: {
+    dimensions: 10,
+    cornerSize: 1.5,
+    hipotenusa: Math.sqrt(2) * 1.5,
```
TomasSlama, line 14: I'm a little bit confused by this calculation. The hypotenuse is calculated `c² = a² + a²`. I would maybe move this calculation to a separate function
```ts
// c² = a² + a²
const getHypotenuse = (pendant: number) => Math.sqrt(Math.pow(pendant) * 2)
```

```suggestion
    hypotenuse: getHypotenuse(1.5),
```

packages/picasso/src/AvatarUpload/test.tsx
```diff
@@ -0,0 +1,66 @@
...
+      }}
+    />
+  )
+}
+
+describe('AvatarUpload', () => {
+  it('renders', () => {
+    const { container } = renderAvatarUpload({})
+
+    expect(container).toMatchSnapshot()
```
TomasSlama, line 24: do we really need this snapshot? 🤔 I think we should use snapshots when using MUI components to help us migrate to new versions. In this case, I don't think it is necessary.

packages/picasso/src/AvatarUpload/test.tsx
```diff
@@ -0,0 +1,66 @@
...
+    it('renders image in background', () => {
+      const { container, getByTestId } = renderAvatarUpload({
+        alt: 'Photo alt text',
+        src: 'http://example.png',
+      })
+
+      const imageAvatar = getByTestId('image-avatar')
+
+      expect(imageAvatar).toBeVisible()
+      expect(container).toMatchSnapshot()
```
TomasSlama, line 37: I would do a visual test instead of these snapshots 🤔 WDYT?

packages/picasso/src/AvatarUpload/story/Sizes.example.tsx
ozgurkececioglu: I try to give at least one container padding to not cut in the story snapshots like this https://happo.io/a/675/components?component=Avatar&variant=Default&target=chrome-desktop&projectId=1189 because sometimes borders need more spaces but we can't see it. Wdyt?

packages/picasso/src/AvatarUpload/test.tsx
ozgurkececioglu: But our changes can affect these components just like MUI migrations, right? let's say we decided to use tailwind + mui/base and in this case, I can't use visual testing to be sure that nothing is broken visually. wdyt?

ozgurkececioglu: ok, I got you wrong first. You are saying that they can be visual tests instead of snapshots. to I can move them to cypress 👍 makes sense

packages/picasso/src/AvatarUpload/styles.ts
```diff
@@ -0,0 +1,156 @@
...
+      if (focused) {
+        borderAttributes.borderStyle = 'solid'
+        borderAttributes.borderWidth = '3px'
+
+        coordinate -= focusedCornerCorrection
+      }
+
+      return {
+        position: 'absolute',
+        left: `-${coordinate}em`,
```
TomasSlama, line 133: In the storybook, it looks not good. Shouldn't this just be half of the hypotenuse? 🤔

TomasSlama: image

packages/picasso/src/AvatarUpload/story/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+      type: {
+        name: 'function',
+        description:
+          '(acceptedFiles: File, fileRejections: FileRejection, event:DragEvent | Event) => void',
+      },
+      description: `
+Callback for when the drop event occurs. Note that this callback is invoked after the getFilesFromEvent callback is done.
+
+Files are accepted or rejected based on the accept, multiple, minSize and maxSize props. accept must be a valid MIME type according to input element specification or a valid file extension. If multiple is set to false and additional files are dropped, all files besides the first will be rejected. Any file which does not have a size in the minSize, maxSize range, will be rejected as well.
+
```
dmaklygin, line 51:
```suggestion
- If multiple is set
+ If `multiple` prop is set
```
maybe?

cypress/component/AvatarUpload.spec.tsx
```diff
@@ -0,0 +1,33 @@
...
+        </Container>
+      )
+
+      cy.getByTestId('warning-icon').realHover()
+      cy.getByTestId('warning-content').should('be.visible')
+
+      cy.get('body').happoScreenshot({
+        component,
+        variant: 'with-warning-message/after-hover',
+      })
```
dmaklygin, line 30: 👍

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <Container
+        style={style}
+        ref={ref}
+        className={cx(classes.root, classes.size, className)}
+      >
+        <AvatarWrapper variant='square' size={size}>
+          <Container
+            flex
+            justifyContent='center'
+            alignItems='center'
```
dmaklygin, line 162: isn't better move these styles to `styles.ts` ?

packages/picasso/src/AvatarUpload/test.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+const renderAvatarUpload = () => render(<AvatarUpload />)
+
+describe('AvatarUpload', () => {
+  it('renders', () => {
+    const { container } = renderAvatarUpload()
+
+    expect(container).toMatchSnapshot()
+  })
+})
```
dmaklygin, line 14: besides, snapshot, I would also test functionality, namely: 
1. `onDropRejected`, `onDropAcepted`, `onDrop` callbacks.
2. test that `useDropzone` is properly called
3. test `isDragActive` in different states

packages/picasso/src/AvatarUpload/styles.ts
ozgurkececioglu: Screen Shot 2022-08-16 at 10 32 20
Half of the hypotenuse is not working for me 😞 this needs a really hard calculation to make it perfect.

My versions are working fine. I tried to check different screen sizes as well and none of them looked bad.

packages/picasso/src/AvatarUpload/test.tsx
ozgurkececioglu: But don't you think these functionalities are coming from react-dropzone 🤔 I mean these tests will start testing react-dropzone instead of our component

plus: we don't have these tests for dropzone as well

ozgurkececioglu:
@toptal-anvil ping reviewers

packages/picasso/src/AvatarUpload/story/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+      },
+      description: 'Custom validation function',
+      defaultvalue: 'null',
+    },
+    onDrop: {
+      name: 'onDrop',
+      type: {
+        name: 'function',
+        description:
+          '(acceptedFiles: File, fileRejections: FileRejection, event:DragEvent | Event) => void',
```
OndrejTuma, line 45: The accepted file can actually be null as well
```suggestion
          '(acceptedFile: File | null, fileRejection: FileRejection | null, event:DragEvent | Event) => void',
```

packages/picasso/src/AvatarUpload/story/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+      },
+      description: `
+Callback for when the drop event occurs. Note that this callback is invoked after the getFilesFromEvent callback is done.
```
OndrejTuma, line 48: I'd probably skip `getFilesFromEvent` callback as it is not mentioned here anywhere
```suggestion
Callback for when the drop event occurs.
```

packages/picasso/src/AvatarUpload/story/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+page.createTabChapter('Props').addComponentDocs({
+  component: AvatarUpload,
+  name: 'AvatarUpload',
+  additionalDocs: {
+    onDropAccepted: {
+      name: 'onDropAccepted',
+      type: {
+        name: 'function',
+        description: '(files: File, event) => void',
```
OndrejTuma, line 17:
```suggestion
        description: '(file: File, event) => void',
```

packages/picasso/src/AvatarUpload/types.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export type AvatarUploadOptions = {
+  /**
+   * Set accepted file types. See https://github.com/okonet/attr-accept for more information.
+   */
+  accept?: string | string[]
+  /** Callback for when the drop event occurs */
+  onDrop?: <T extends File>(
+    acceptedFile: T,
+    fileRejection: FileRejection,
```
OndrejTuma, line 16:
```suggestion
    acceptedFile: T | null,
    fileRejection: FileRejection | null,
```

packages/picasso/src/AvatarUpload/story/Default.example.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const useFile = () => {
+  const [uploading, setUploading] = useState<boolean>(false)
+  const [src, setSrc] = useState<string | undefined>()
+
+  const addFile = (acceptedFile: File) => {
+    const reader = new FileReader()
```
OndrejTuma, line 9: If people are going to use this as an example its good to handle acceptedFile if it is null. Otherwise, reading the file would throw an error and didnt trigger `onDropRejected` callback
```suggestion
  const addFile = (acceptedFile: File | null) => {
    if (!acceptedFile) {
      return
    }
    const reader = new FileReader()
```

TomasSlama (approved):
Nice Ozgur, lot of good work, tested manually.

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,226 @@
...
+    )
+
+    return (
+      <Container
+        style={style}
+        ref={ref}
+        className={cx(classes.root, classes.size, className)}
+      >
+        <AvatarWrapper variant='square' size={size}>
+          <Container
```
denieler, line 187: can't we pass all those props to `AvatarWrapper` component? it seems we can remove one not necessary nesting component - `Container` 🤔

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,226 @@
...
+    const handleDropRejected = useCallback(
+      (fileRejections: FileRejection[], event: DropEvent) => {
+        if (onDropRejected) {
+          onDropRejected(fileRejections[0], event)
+        }
+      },
+      [onDropRejected]
+    )
+
+    const handleFocus = () => {
```
denieler, line 123:
```suggestion
    const handleFocus = useCallback(() => {
```
?

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,226 @@
...
+      onDropAccepted: handleDropAccepted,
+      onDropRejected: handleDropRejected,
+      validator,
+    })
+
+    const classes = useStyles({ ...props, focused })
+
+    const sourceFileExist = src !== undefined && src !== ''
+    const showImageAvatar = sourceFileExist && !isDragActive
+    const uploadedImageAvatar = showImageAvatar && (
```
denieler, line 155: wouldn't it be `false` in case of `showImageAvatar === false`? so it might render `false` string 🤔

packages/picasso/src/AvatarUpload/story/Default.example.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react'
+import { AvatarUpload, Container } from '@toptal/picasso'
+
+const useFile = () => {
```
denieler, line 4: nit: maybe move this hook implementation below the example, so example itself would be more visible?

packages/picasso/src/AvatarUpload/story/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+const page = PicassoBook.section('Components').createPage(
+  'AvatarUpload',
+  'Gets the image from file input and displays it in the Avatar component.'
+)
+
+page.createTabChapter('Props').addComponentDocs({
+  component: AvatarUpload,
+  name: 'AvatarUpload',
+  additionalDocs: {
+    onDropAccepted: {
```
denieler, line 13: if type is not correctly recognized I would suggest using inline comments to specify it correctly - https://github.com/toptal/picasso/blob/2d76bbf2e1b565829145f3fd338ff86c9d198f6b/packages/topkit-analytics-charts/src/AnalyticsChart/AnalyticsChart.tsx#L17, to me it seems better, because closer to the place of the prop definition, so it would be less cases when type got updated, but the doc has not been

packages/picasso/src/AvatarUpload/types.ts
```diff
@@ -0,0 +1,50 @@
...
+  | Event
+
+export const ErrorCode = {
+  FileInvalidType: 'file-invalid-type',
+  FileTooLarge: 'file-too-large',
+  FileTooSmall: 'file-too-small',
+  TooManyFiles: 'too-many-files',
+} as const
+
+type ErrorCodeType = typeof ErrorCode[keyof typeof ErrorCode]
```
denieler, line 40: don't we want to export this type as well?

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: I was thinking that we would pass the callbacks to div so there shouldn't be needed to use `useCallback` but you are right. `Container` usage can change and then we may need these memorizing 👍

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: If I understood correctly, I don't think so. It is one of the most common conditional rendering usages in react, right?

This should also print "false" in this case but it doesn't.
```
return (
    <>
        {x > 1 && <div>testme</div>}
    </>
)
```

packages/picasso/src/AvatarUpload/story/Default.example.tsx
ozgurkececioglu: I used our Dropzone implementation as a reference and Dropzone itself also needs refactoring 🙂 
I will update Dropzone later separately.

packages/picasso/src/AvatarUpload/types.ts
ozgurkececioglu: Again, this is from Dropzone types here https://github.com//toptal/picasso/blob/2d76bbf2e1b565829145f3fd338ff86c9d198f6b/packages/picasso/src/Dropzone/types.ts#L46.

I mostly duplicated them because one of the components can be updated independently so there is no need to make them coupled for types. Makes sense?

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: I tried this last night but fixing the issues about layering the components against z-index is not easy. I mean your suggestion will make them equal in depth and dropzone should be on the top always. I need to keep imageAvatar outside of the Dropzone but inside AvatarWrapper. That's why there needs to be one more layer.

I'm open to alternative solutions 🙂

OndrejTuma (approved):
Thanks Oz, great work you did there 🙇

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
denieler: yep, you are right 👍 tested, looks like none of `true, false, null` values render anything

packages/picasso/src/AvatarUpload/types.ts
denieler: `ErrorCodeType` is used in `FileError` type, which is exported, so I think it make sense to export this one as well 👍

ozgurkececioglu: @denieler done ✅

cypress/component/AvatarUpload.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+            src={src}
+            hovered
+            testIds={{
+              uploadIcon: 'upload-icon',
+            }}
+          />
+        </Container>
+      )
+
+      cy.getByTestId('upload-icon').should('be.visible')
```
dmaklygin, line 58: isn't enough to take only a screenshoot?

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+  size?: SizeType<'small' | 'large'>
+  /** Enable/disable the dropzone */
+  disabled?: boolean
+  /** Maximum file size (in bytes) */
+  maxSize?: number
+  /** Minimum file size (in bytes) */
+  minSize?: number
+  /**
+   * Callback for when the drop event occurs. Note that if file is not accepted, this callback is not invoked.,
+   * @type <T extends File>(files: T, event: DropEvent) => void
```
dmaklygin, line 35: hm, is it intentional to repeat type in comments?

ozgurkececioglu: I have added the type as a fallback for docs section https://github.com/toptal/picasso/pull/3041#discussion_r946907296

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    // callback overrides to return only one file to the parent component
+    const handleDrop = useCallback(
+      (
+        acceptedFiles: File[],
+        fileRejections: FileRejection[],
+        event: DropEvent
+      ) => {
+        if (onDrop) {
+          onDrop(acceptedFiles?.[0] ?? null, fileRejections?.[0] ?? null, event)
```
dmaklygin, line 108: according to definition of `acceptedFiles` and `fileRejections` — they cannot be `undefined`.
so, I think `?` is useless here, isn't it?

cypress/component/AvatarUpload.spec.tsx
ozgurkececioglu: without an assert, it didn't work. I mean, the image was not representing the hovering state. I thought it is innocent.

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+        }
+      },
+      [onDrop]
+    )
+
+    const handleDropAccepted = useCallback(
+      (files: File[], event: DropEvent) => {
+        if (onDropAccepted && files.length === 1) {
+          onDropAccepted(files[0], event)
+        }
```
dmaklygin, line 118: Could you please explain this logic. Why we rely only on case when `files.length` is 1?

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+        if (onDropRejected) {
+          onDropRejected(fileRejections[0], event)
+        }
+      },
+      [onDropRejected]
+    )
+
+    const handleFocus = useCallback(() => {
+      setFocused(true)
+    }, [])
```
dmaklygin, line 134: at least one dependency: `setFocused`

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+      setFocused(false)
+    }, [])
+
+    const handleMouseEnter = useCallback(() => {
+      setHovered(true)
+    }, [])
+
+    const handleMouseLeave = useCallback(() => {
+      setHovered(false)
+    }, [])
```
dmaklygin, line 146: I would also add dependencies for useCallback`. Honestly, I would remove `useCallback` here :)

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: React accepts that setState functions are static and they won't be changing. No need to add them as a dependency. See: https://reactjs.org/docs/hooks-reference.html#usestate

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: As I mentioned, setState functions are stable so no need to add them as dependencies. The reason for useCallback is that I am passing these callbacks to `Container` component. One day, if we decide to use a `useEffect` inside `Container` with one of these callbacks above, it will retrigger each time and it will be impossible to collect all usages like these. So, it is for being on the safe side. Wdyt?

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,235 @@
...
+        data-testid={testIds?.imageAvatar}
+      />
+    )
+
+    const showUploadIcon = (sourceFileExist && hovered) || !sourceFileExist
+    const icon = uploading ? (
+      <Loader size='small' variant='inherit' data-testid={testIds?.loader} />
+    ) : (
+      showUploadIcon && <Upload24 data-testid={testIds?.uploadIcon} />
+    )
```
dmaklygin, line 179: I would create a separate hook (`useAvatarUpload`) and move such conditions there, otherwise, testing the `AvatarUpload` Component is getting more difficult. 
The component should be silly, just for rendering :)

packages/picasso/src/AvatarUpload/story/index.jsx
dmaklygin: Hm, this is something new. I've checked and we specify `@type` only in `picasso-charts`

packages/picasso/src/AvatarUpload/test.tsx
dmaklygin: I would rely on unit tests, it means, we should mock the `useDropzone` and return a specified response from this hook and test the result output

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: After providing `multiple=false`, there is no way to get more than one item in the files array so I will remove the additional length condition 👍

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: Good catch! Removing... 👍

ozgurkececioglu: @dmaklygin done ✅

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: @dmaklygin done ✅

packages/picasso/src/AvatarUpload/test.tsx
ozgurkececioglu: @dmaklygin done ✅

packages/picasso/src/AvatarUpload/styles.ts
```diff
@@ -0,0 +1,161 @@
...
+      borderWidth: sizes.borderWidth,
+      borderColor: palette.blue.main,
+      color: palette.blue.main,
+      borderStyle: 'dashed',
+      transition: `all ${transitions.duration.short}ms ${transitions.easing.easeOut}`,
+      transitionProperty: 'background-color',
+      cursor: 'pointer',
+      '&$dragActive': {
+        background:
+          'linear-gradient(0deg, rgba(32, 78, 207, 0.16), rgba(32, 78, 207, 0.16))',
```
denieler, line 56: are those `rgba` colors part of theme.palette?

packages/picasso/src/AvatarUpload/story/index.jsx
ozgurkececioglu: I kept it for a reason but I should only keep the `onDrop` prop description inside storybook docs because it has a code template and it won't work inside `AvatarUpload > Props` section.

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,230 @@
...
+  /** Custom validation function */
+  validator?: AvatarUploadOptions['validator']
+  /** Warning message to be used as a tooltip */
+  warningMessage?: ReactNode
+  /** Indicate whether input is in error state */
+  error?: boolean
+  /** Indicate whether the selected file is being uploaded */
+  uploading?: boolean
+  focused?: boolean
+  hovered?: boolean
```
denieler, line 58: these props seems undocumented and also seems they are used only like `initial` props, but it's not represented in the prop name

packages/picasso/src/AvatarUpload/story/index.jsx
denieler: @ozgurkececioglu what do you mean by code template? why we can't move it to `Props` as well?

ozgurkececioglu: I meant this:
Screen Shot 2022-08-18 at 10 48 51

when I copy it under `Props` section, it doesn't work

packages/picasso/src/AvatarUpload/styles.ts
ozgurkececioglu: Aligning with Matt. He said they are from our palette but I need to find correct ones and replace the names 👍

packages/picasso/src/AvatarUpload/story/index.jsx
denieler: what if we would create a special example for using this prop and remove the code from prop documentation?

ozgurkececioglu: I think you meant `what about`? I think I can prepare a new story and trim this prop description. Makes sense 👍

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: @denieler done ✅

packages/picasso/src/AvatarUpload/story/index.jsx
ozgurkececioglu: @denieler done ✅

packages/picasso/src/AvatarUpload/test.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+      expect(mockOnDrop).toHaveBeenCalledWith(
+        dummyFile,
+        null,
+        expect.any(Object)
+      )
+      expect(mockOnDropAccepted).toHaveBeenCalledWith(
+        dummyFile,
+        expect.any(Object)
+      )
```
dmaklygin, line 57: I would also add:
```
expect(mockOnDropAccepted).toHaveBeenCalledTimes(1)
```

ozgurkececioglu: @dmaklygin done ✅ 🙇‍♂️

packages/picasso/src/AvatarUpload/test.tsx
dmaklygin: I would also check whether `useDropzone` is properly called or not. 
And I would mock the `useDropzone` as well

dmaklygin (approved):
Two minor comments

ozgurkececioglu:
Closing this due to not confuse with new designs and implementations here https://github.com/toptal/picasso/pull/3068

PR #3069, opened 2022-08-26
feat(AvatarUpload): implement AvatarUpload

ozgurkececioglu:
[FX-2889]

### Description

- added functionality as simple as possible. other requirements like sizes and visual states will follow this PR.

### How to test

- go to [component page](https://picasso.toptal.net/fx-2889-implement-basic-avatar-upload/?path=/story/components-avatarupload--avatarupload)
- there are two stories: default and with-upload. You can check them both to check the functionality works

### Screenshots

Screen Shot 2022-08-29 at 11 54 18
Screen Shot 2022-08-29 at 11 54 13

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-2889]: https://toptal-core.atlassian.net/browse/FX-2889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/AvatarUpload/AvatarUpload.tsx
+++ b/packages/picasso/src/AvatarUpload/AvatarUpload.tsx
@@ -131,6 +131,7 @@ export const AvatarUpload = forwardRef<HTMLDivElement, Props>(
     const showLoader = Boolean(uploading)
     const showAvatar = !showLoader && Boolean(src)
     const showUploadIcon = !showAvatar && !showLoader
+    const showEditIcon = Boolean(onEdit)
 
     const classes = useStyles()
 
@@ -171,7 +172,7 @@ export const AvatarUpload = forwardRef<HTMLDivElement, Props>(
         {showAvatar ? (
           <Avatar
             size='small'
-            onEdit={handleEdit}
+            onEdit={showEditIcon ? handleEdit : undefined}
             // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
             src={src!}
             alt={alt}
@@ -180,9 +181,9 @@ export const AvatarUpload = forwardRef<HTMLDivElement, Props>(
           />
         ) : (
           <>
+            <DropzoneSvg data-testid={testIds?.dropzoneSvg} />
             {loadingIcon}
             {uploadIcon}
-            <DropzoneSvg data-testid={testIds?.dropzoneSvg} />
           </>
         )}
       </div>
@@ -194,7 +195,7 @@ AvatarUpload.displayName = 'AvatarUpload'
 
 AvatarUpload.defaultProps = {
   disabled: false,
-  maxSize: Infinity,
+  maxSize: 104857600, // 100MB in bytes (100 * 1024 * 1024)
   minSize: 0,
   accept: 'image/*',
 }
--- a/packages/picasso/src/AvatarUpload/AvatarUpload.tsx
+++ b/packages/picasso/src/AvatarUpload/AvatarUpload.tsx
@@ -1,0 +1,202 @@
+import React, { forwardRef, useCallback } from 'react'
+import { makeStyles, Theme } from '@material-ui/core'
+import { BaseProps } from '@toptal/picasso-shared'
+import { useDropzone } from 'react-dropzone'
+import cx from 'classnames'
+
+import Avatar from '../Avatar'
+import styles from './styles'
+import { AvatarUploadOptions, DropEvent, FileRejection } from './types'
+import DropzoneSvg from './DropzoneSvg/DropzoneSvg'
+import Loader from '../Loader'
+import { Upload24 } from '../Icon'
+
+export interface Props extends BaseProps {
+  /**
+   * Set accepted file types. See https://github.com/okonet/attr-accept for more information.
+   */
+  accept?: AvatarUploadOptions['accept']
+  /** Alt text */
+  alt?: string
+  /** Image URL */
+  src?: string
+  /** Enable/disable the dropzone */
+  disabled?: boolean
+  /** Maximum file size (in bytes) */
+  maxSize?: number
+  /** Minimum file size (in bytes) */
+  minSize?: number
+  /**
+   * Callback for when there is already a source and user clicks on the avatar.
+   */
+  onEdit?: AvatarUploadOptions['onEdit']
+  /**
+   * Callback for when the drop event occurs. Note that if file is not accepted, this callback is not invoked.,
+   * @type <T extends File>(files: T, event: DropEvent) => void
+   */
+  onDropAccepted?: AvatarUploadOptions['onDropAccepted']
+  /**
+   * Callback for when the drop event occurs. Note that if file is not rejected, this callback is not invoked.
+   * @type (fileRejection: FileRejection, event: DropEvent) => void
+   */
+  onDropRejected?: AvatarUploadOptions['onDropRejected']
+  /**
+   * Callback for when the drop event occurs. Note that the onDrop callback will always be invoked regardless if the dropped file was accepted or rejected.
+   * @type <T extends File>(acceptedFile: T | null, fileRejection: FileRejection | null, event: DropEvent) => void
+   */
+  onDrop?: AvatarUploadOptions['onDrop']
+  /**
+   * Custom validation function
+   * (file: File) => FileError | FileError[] | null
+   */
+  validator?: AvatarUploadOptions['validator']
+  /** Indicate whether the selected file is being uploaded */
+  uploading?: boolean
+  testIds?: {
+    avatar?: string
+    dropzoneSvg?: string
+    uploadIcon?: string
+    loader?: string
+  }
+}
+
+const useStyles = makeStyles<Theme>(styles, {
+  name: 'PicassoAvatarUpload',
+})
+
+export const AvatarUpload = forwardRef<HTMLDivElement, Props>(
+  function AvatarUpload(props, ref) {
+    const {
+      uploading,
+      onEdit,
+      'data-testid': dataTestId,
+      testIds,
+      src,
+      alt,
+
+      // dropzoneOptions
+      accept,
+      minSize,
+      maxSize,
+      disabled,
+      onDrop,
+      onDropAccepted,
+      onDropRejected,
+      validator,
+    } = props
+
+    // callback overrides to return only one file to the parent component
+    const handleDrop = useCallback(
+      (
+        acceptedFiles: File[],
+        fileRejections: FileRejection[],
+        event: DropEvent
+      ) => {
+        if (onDrop) {
+          onDrop(acceptedFiles[0] ?? null, fileRejections[0] ?? null, event)
+        }
+      },
+      [onDrop]
+    )
+
+    const handleDropAccepted = useCallback(
+      (files: File[], event: DropEvent) => {
+        if (onDropAccepted) {
+          onDropAccepted(files[0], event)
+        }
+      },
+      [onDropAccepted]
+    )
+
+    const handleDropRejected = useCallback(
+      (fileRejections: FileRejection[], event: DropEvent) => {
+        if (onDropRejected) {
+          onDropRejected(fileRejections[0], event)
+        }
+      },
+      [onDropRejected]
+    )
+
+    const handleEdit = (
+      event: React.MouseEvent<HTMLDivElement, MouseEvent>
+    ) => {
+      // to avoid dropzone to be triggered
+      event.stopPropagation()
+
+      if (onEdit) {
+        onEdit(event)
+      }
+    }
+
+    const showLoader = Boolean(uploading)
+    const showAvatar = !showLoader && Boolean(src)
+    const showUploadIcon = !showAvatar && !showLoader
+
+    const classes = useStyles()
+
+    const loadingIcon = showLoader && (
+      <Loader
+        className={classes.icon}
+        size='small'
+        variant='inherit'
+        data-testid={testIds?.loader}
+      />
+    )
+    const uploadIcon = showUploadIcon && (
+      <Upload24 className={classes.icon} data-testid={testIds?.uploadIcon} />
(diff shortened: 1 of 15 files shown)


packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,202 @@
...
+      onDrop: handleDrop,
+      onDropAccepted: handleDropAccepted,
+      onDropRejected: handleDropRejected,
+      validator,
+    })
+
+    return (
+      <div
+        {...getRootProps({
+          ref,
```
ozgurkececioglu, line 164: Ref forwarding to expose it for users. They can use it to reach to the component and trigger a click if needed.

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,202 @@
...
+        }
+      },
+      [onDropRejected]
+    )
+
+    const handleEdit = (
+      event: React.MouseEvent<HTMLDivElement, MouseEvent>
+    ) => {
+      // to avoid dropzone to be triggered
+      event.stopPropagation()
```
ozgurkececioglu, line 124: There is no way to keep dropzone in the DOM but avoid being clicked. We have to keep dropzone because while editing, users may want to open a new file selection dialog for start over.

packages/picasso/src/AvatarUpload/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  validator?: <T extends File>(file: T) => FileError | null
+}
+
+export type DropEvent =
+  | React.DragEvent<HTMLElement>
+  | React.ChangeEvent<HTMLInputElement>
+  | DragEvent
+  | Event
+
+export const ErrorCode = {
```
ozgurkececioglu, line 27: Most of these types coming from `react-dropzone`. It is just local duplication to avoid type clashing in a major `react-dropzone` version.

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
```diff
@@ -0,0 +1,202 @@
...
+      </div>
+    )
+  }
+)
+
+AvatarUpload.displayName = 'AvatarUpload'
+
+AvatarUpload.defaultProps = {
+  disabled: false,
+  maxSize: Infinity,
```
denieler, line 197: nit: I would probably set some value here, like 100 Mb or something

packages/picasso/src/AvatarUpload/DropzoneSvg/DropzoneSvg.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { getBackgroundShape, getBordersShape, getOutlineShape } from './utils'
+
+const BASE_FONT_SIZE = 16
+
+const dimensions = 5 * BASE_FONT_SIZE
+const cornerSize = BASE_FONT_SIZE
+
+const backgroundShape = getBackgroundShape(dimensions, cornerSize)
+const bordersShape = getBordersShape(dimensions, cornerSize)
+const outlineShape = getOutlineShape(dimensions, cornerSize)
```
denieler, line 15: this is not clear to me why we don't just use an icon from picasso-icons and need to create svg almost manually?

packages/picasso/src/AvatarUpload/DropzoneSvg/styles.ts
```diff
@@ -0,0 +1,31 @@
+import { createStyles, Theme } from '@material-ui/core/styles'
+
+export default ({ palette, transitions }: Theme) =>
+  createStyles({
+    root: {
+      position: 'relative',
+      width: '80px',
+      height: '80px',
```
denieler, line 8: why we don't use `rem` size here?

packages/picasso/src/AvatarUpload/DropzoneSvg/DropzoneSvg.tsx
ozgurkececioglu: The problem is, on the left bottom of the avatar there is a cut. It should be applied to both background, borders, and outline (for focus). And these values are changing for different Avatar sizes, small and large. It is not scaling with percentage. So, it should be handled as a component instead of a static SVG.

packages/picasso/src/AvatarUpload/DropzoneSvg/styles.ts
ozgurkececioglu: SVG's `d` parameter works with percentage and pixels only. So. I have decided to keep everything pixel inside DropzoneSvg.

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
ozgurkececioglu: The same mistake from our Dropzone component. Let me handle 👍

packages/picasso/src/AvatarUpload/story/WithUpload.example.tsx
```diff
@@ -0,0 +1,62 @@
...
+      console.log(fileRejection.errors.join(', '))
+    }
+  }
+
+  const handleEdit = () => {
+    // free to do whatever you want with editing event
+    // e.g. open a modal to change zoom or selected area of the uploaded image
+
+    // you can use the ref to access the root component
+    avatarUploadRef.current?.click()
```
denieler, line 45: why do users need to click on avatarUpload on edit?

packages/picasso/src/AvatarUpload/story/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+  component: AvatarUpload,
+  name: 'AvatarUpload',
+})
+
+page
+  .createChapter()
+  .addExample('AvatarUpload/story/Default.example.tsx', 'Default')
+  .addExample('AvatarUpload/story/WithUpload.example.tsx', {
+    title: 'With upload',
+    description: 'Simulates file upload and editing',
```
denieler, line 19:
```suggestion
    description: 'Shows how the component can be used with FileReader to upload a file',
```

packages/picasso/src/AvatarUpload/styles.ts
```diff
@@ -0,0 +1,24 @@
...
+      outline: 'none',
+    },
+
+    size: {
+      width: '5rem',
+      height: '5rem',
+    },
+
+    icon: {
+      position: 'absolute',
```
denieler, line 20: can we position the icon in a non-absolute positioned way?

packages/picasso/src/AvatarUpload/test.tsx
```diff
@@ -0,0 +1,127 @@
...
+      })
+
+      const avatar = getByTestId('avatar')
+      const dropzoneSvg = queryByTestId('dropzone-svg')
+
+      expect(avatar).toBeVisible()
+      expect(dropzoneSvg).not.toBeInTheDocument()
+    })
+  })
+})
```
denieler, line 127: nice tests 👍

denieler (approved):
Haven't tested it, but the code looks good to me, except could small corrections 👍

packages/picasso/src/AvatarUpload/story/WithUpload.example.tsx
ozgurkececioglu: We discussed as a team of engineers and Matt how to handle common use cases of AvatarUpload before implementation. One of them has this flow that makes sense:

Screen Shot 2022-08-26 at 11 27 41
Screen Shot 2022-08-26 at 11 28 08
Screen Shot 2022-08-26 at 11 27 52

generally, people want to adjust their avatars before saving. And we need to enable re-selection of the avatar after selecting one.

denieler (commented):
btw, maybe we should add also a visual test for this component? when we uploaded some image how the uploaded state would look like

packages/picasso/src/AvatarUpload/story/WithUpload.example.tsx
denieler: but in that real-world example you would probably open a new edit-photo modal in onEdit callback, this would make sense. But it still doesn't make sense to me to fire a click on avatar upload component from onEdit callback 🤔

packages/picasso/src/AvatarUpload/AvatarUpload.tsx
denieler: I think it would be just a bit safer

packages/picasso/src/AvatarUpload/DropzoneSvg/DropzoneSvg.tsx
denieler: I don't like this code, because it is not clear what it does and why. But I'm not against it. If you guys already discussed multiple variants of implementation and chose this one - definitely up to you, I don't have much experience here

packages/picasso/src/AvatarUpload/DropzoneSvg/styles.ts
denieler: I would probably mention it somewhere in a comment

packages/picasso/src/AvatarUpload/story/WithUpload.example.tsx
ozgurkececioglu: Oh, it was just an example to tell user that they can use ref to reach the component. I can remove it if you think it is confusing.

denieler: I would make even a separate example for `onEdit` 🤔

packages/picasso/src/AvatarUpload/DropzoneSvg/DropzoneSvg.tsx
ozgurkececioglu: I tried some other solutions but those just failed. After checking out how SVG paths work, I think the code would become easier to get. And I don't think avatar shape will change frequently so I guess we are safe.

packages/picasso/src/AvatarUpload/story/index.jsx
ozgurkececioglu: Let me add a description to this component because users may think that component can be used without an external upload service. The component should use value as `string` representing `src` and it will not expose `File` type except `onDrop` callback.

packages/picasso/src/AvatarUpload/styles.ts
ozgurkececioglu: Idk how to do that. Do you have something in your mind?

denieler: you are right 🤔 I was confused with `zIndex: 1`, probably we can remove it at least and just change the order in JSX?

packages/picasso/src/AvatarUpload/DropzoneSvg/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export const getOutlineShape = (dimensions: number, cornerSize: number) => {
+  const centerShift = 2 // shift for the outline stroke
+  const outlineStrokeWidth = 2 // width of the outline stroke
+
+  return `
+    M ${centerShift} ${centerShift}
+    H ${dimensions + centerShift + outlineStrokeWidth} 
+    V ${dimensions + centerShift + outlineStrokeWidth} 
+    H ${cornerSize + 1}
```
dmaklygin, line 30: do you also need `centerShift` here?

packages/picasso/src/AvatarUpload/DropzoneSvg/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  const outlineStrokeWidth = 2 // width of the outline stroke
+
+  return `
+    M ${centerShift} ${centerShift}
+    H ${dimensions + outlineStrokeWidth} 
+    V ${dimensions + outlineStrokeWidth} 
+    H ${cornerSize + outlineStrokeWidth} 
+    L ${centerShift} ${dimensions - cornerSize + outlineStrokeWidth} Z
+  `
+}
```
dmaklygin, line 51: wdyt to create a single function which builds a shape. Other functions will reuse it and pass parameters?

packages/picasso/src/AvatarUpload/types.ts
dmaklygin: wdyt to rename it to `AvatarUloadErrorCode` to avoid collission in types names

dmaklygin (approved):
Haven't checked. Checked the code!

packages/picasso/src/AvatarUpload/types.ts
ozgurkececioglu: I have already exported them as you suggested to Picasso clients. Since the error codes won't be used by other internal components, I think we are safe, wdyt?

packages/picasso/src/AvatarUpload/DropzoneSvg/utils.ts
ozgurkececioglu: very simple and wonderful idea 🙇‍♂️  I will put some documentation for that function as well so it may help us to understand how the path parameters work better.

packages/picasso/src/AvatarUpload/DropzoneSvg/utils.ts
ozgurkececioglu: I thought hard about this as well but I couldn't find a good formula for this value. I found '+1' while testing the app. 

This has to stay static for now 😞 If you have a suggestion I would definitely try that.

separatio (approved):
Manually tested. Works as expected. No edit state present on the component, but the avatar can be edited/ changed.

PR #3096, opened 2022-09-12
feat(Tabs): introduce `orientation` prop

iatanas0v:
[TACO-1038]

### Description

This introduces the `orientation` prop on `Tabs`. This is already supported by MUI itself so I just had to expose it and style the varient.

I bounced this off BASE already - https://toptal-core.slack.com/archives/CCC3GP6CC/p1662703900679159

### How to test

- Check the new vertical tabs example in storybook

### Screenshots

image

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

- n/a ~~codemod is created and showcased in the changeset~~
- n/a ~~test alpha package of Picasso in StaffPortal~~

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[TACO-1038]: https://toptal-core.atlassian.net/browse/TACO-1038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/Tab/Tab.tsx
+++ b/packages/picasso/src/Tab/Tab.tsx
@@ -8,6 +8,7 @@ import React, {
 import { makeStyles, Theme } from '@material-ui/core/styles'
 import MUITab, { TabProps } from '@material-ui/core/Tab'
 import { BaseProps, TextLabelProps, useTitleCase } from '@toptal/picasso-shared'
+import cx from 'classnames'
 
 import Typography from '../Typography'
 import styles from './styles'
@@ -75,8 +76,14 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       selected={selected}
       onChange={onChange}
       onClick={onClick}
-      classes={classes}
-      className={`PicassoTab-${orientation}`}
+      classes={{
+        root: cx({
+          [classes.horizontal]: orientation === 'horizontal',
+          [classes.vertical]: orientation === 'vertical',
+        }),
+        selected: classes.selected,
+        wrapper: classes.wrapper,
+      }}
     />
   )
 })
--- a/packages/picasso/src/Tab/Tab.tsx
+++ b/packages/picasso/src/Tab/Tab.tsx
@@ -8,7 +8,6 @@ import React, {
 import { makeStyles, Theme } from '@material-ui/core/styles'
 import MUITab, { TabProps } from '@material-ui/core/Tab'
 import { BaseProps, TextLabelProps, useTitleCase } from '@toptal/picasso-shared'
-import cx from 'classnames'
 
 import Typography from '../Typography'
 import styles from './styles'
@@ -77,10 +76,7 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       onChange={onChange}
       onClick={onClick}
       classes={{
-        root: cx({
-          [classes.horizontal]: orientation === 'horizontal',
-          [classes.vertical]: orientation === 'vertical',
-        }),
+        root: classes[orientation],
         selected: classes.selected,
         wrapper: classes.wrapper,
       }}
--- a/packages/picasso/src/Tab/Tab.tsx
+++ b/packages/picasso/src/Tab/Tab.tsx
@@ -61,7 +61,7 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       {titleCase ? toTitleCase(label) : label}
     </Typography>
   )
-  const orientation = useContext(TabsOrientationContext) || 'horizontal'
+  const orientation = useContext(TabsOrientationContext)
 
   return (
     <MUITab
--- a/packages/picasso/src/Tab/Tab.tsx
+++ b/packages/picasso/src/Tab/Tab.tsx
@@ -3,6 +3,7 @@ import React, {
   ReactNode,
   HTMLAttributes,
   ReactElement,
+  useContext,
 } from 'react'
 import { makeStyles, Theme } from '@material-ui/core/styles'
 import MUITab, { TabProps } from '@material-ui/core/Tab'
@@ -11,6 +12,7 @@ import { BaseProps, TextLabelProps, useTitleCase } from '@toptal/picasso-shared'
 import Typography from '../Typography'
 import styles from './styles'
 import toTitleCase from '../utils/to-title-case'
+import { TabsOrientationContext } from '../Tabs/Tabs'
 
 export interface Props
   extends BaseProps,
@@ -59,6 +61,7 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       {titleCase ? toTitleCase(label) : label}
     </Typography>
   )
+  const orientation = useContext(TabsOrientationContext) || 'horizontal'
 
   return (
     <MUITab
@@ -73,6 +76,7 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       onChange={onChange}
       onClick={onClick}
       classes={classes}
+      className={`PicassoTab-${orientation}`}
     />
   )
 })
--- a/packages/picasso/src/Tab/styles.ts
+++ b/packages/picasso/src/Tab/styles.ts
@@ -1,10 +1,8 @@
 import { Theme, createStyles } from '@material-ui/core/styles'
 import { rem } from '@toptal/picasso-shared'
-import { PicassoProvider, palette } from '@toptal/picasso-provider'
+import { PicassoProvider } from '@toptal/picasso-provider'
 
-const ACTIVE_SHADOW = `4px 0 0 ${palette.grey.lightest}, 0 0 4px rgba(0, 0, 0, 0.08)`
-
-PicassoProvider.override(({ breakpoints }: Theme) => ({
+PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
   MuiTab: {
     root: {
       minHeight: 0,
@@ -47,7 +45,7 @@ PicassoProvider.override(({ breakpoints }: Theme) => ({
   },
 }))
 
-export default ({ sizes }: Theme) =>
+export default ({ sizes, palette, shadows }: Theme) =>
   createStyles({
     horizontal: {
       '&:not(:last-child)': {
@@ -79,7 +77,7 @@ export default ({ sizes }: Theme) =>
     },
     selected: {
       '&$vertical': {
-        boxShadow: ACTIVE_SHADOW,
+        boxShadow: shadows[1],
 
         '&::before': {
           content: '""',
--- a/packages/picasso/src/Tab/styles.ts
+++ b/packages/picasso/src/Tab/styles.ts
@@ -2,7 +2,7 @@ import { Theme, createStyles } from '@material-ui/core/styles'
 import { rem } from '@toptal/picasso-shared'
 import { PicassoProvider } from '@toptal/picasso-provider'
 
-PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
+PicassoProvider.override(({ breakpoints, palette, sizes }: Theme) => ({
   MuiTab: {
     root: {
       minHeight: 0,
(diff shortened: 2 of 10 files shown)


packages/picasso/src/Tab/Tab.tsx
```diff
@@ -59,6 +61,7 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       {titleCase ? toTitleCase(label) : label}
     </Typography>
   )
+  const orientation = useContext(TabsOrientationContext) || 'horizontal'
```
ozgurkececioglu, line 64: Do you still need the default value even after defining it in context creation?

ozgurkececioglu:
I think there are some visual differences between implementation and designs we need to fix
Screen Shot 2022-09-12 at 15 13 34
Screen Shot 2022-09-12 at 15 13 15
I'm saying this because we are trying to pixel-perfect based on the designs to eliminate 

As far as I see:
- hovering on the non-selected tab is missing
- there is a shadow difference 
- heights are different than designs
- right side of the edge of a Tab is transparent. Do we want that? We can ask Matt's opinion about this if needed.

packages/picasso/src/Tab/Tab.tsx
```diff
@@ -73,6 +76,7 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       onChange={onChange}
       onClick={onClick}
       classes={classes}
+      className={`PicassoTab-${orientation}`}
```
ozgurkececioglu, line 79: I wouldn't follow this approach. I think it would be better to use classes for the classname from `styles` file. Please check this example out for more info https://github.com//toptal/picasso/blob/2d76bbf2e1b565829145f3fd338ff86c9d198f6b/packages/picasso/src/Button/Button.tsx#L156

packages/picasso/src/Tab/styles.ts
```diff
@@ -18,8 +18,17 @@ PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
...
+        '&:not(:last-child)': {
+          marginRight: '2em',
+        },
+      },
+
+      '&.PicassoTab-vertical': {
+        borderRadius: '0.5rem 0 0 0.5rem',
+        margin: '0.125rem 0',
+        overflow: 'hidden',
+        padding: `${rem('8px')} 0 ${rem('8px')}`,
```
ozgurkececioglu, line 31: Is there a specific reason to not use `0.5rem` directly?

packages/picasso/src/Tab/styles.ts
```diff
@@ -39,6 +48,20 @@ PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
     },
     selected: {
       color: palette.common.black,
+
+      '&.PicassoTab-vertical': {
+        boxShadow: `0.25rem 0 0 ${palette.grey.lightest}, 0 0 0.25rem rgba(0, 0, 0, 0.08)`,
```
ozgurkececioglu, line 53: We have already defined boxShadows inside theme. Please check them out first before defining a custom one. And please make sure it is the same with figma 🙇‍♂️

packages/picasso/src/Tab/styles.ts
iatanas0v: No, not really. I just tried to be a smartass, I guess :sweat_smile:

packages/picasso/src/Tab/styles.ts
iatanas0v: I checked `packages/picasso-provider/src/Picasso/config/shadows.ts` but didn't find similar shadow. Is that the right place to look ?

Also you mentioned the shadow is different than the design. I double checked it and it's the very same :thinking: Maybe I'm missing something here, can you point out my mistake, please :pray:

iatanas0v:
> hovering on the non-selected tab is missing
fixed
> there is a shadow difference
see my other comment  https://github.com/toptal/picasso/pull/3096#discussion_r968475915
> heights are different than designs
fixed
> right side of the edge of a Tab is transparent. Do we want that? We can ask Matt's opinion about this if needed.
I didn't understand you :see_no_evil: :thinking: Can you elaborate, please

packages/picasso/src/Tab/Tab.tsx
```diff
@@ -72,7 +76,14 @@ export const Tab = forwardRef<HTMLDivElement, Props>(function Tab(props, ref) {
       selected={selected}
       onChange={onChange}
       onClick={onClick}
-      classes={classes}
+      classes={{
+        root: cx({
+          [classes.horizontal]: orientation === 'horizontal',
+          [classes.vertical]: orientation === 'vertical',
+        }),
```
ozgurkececioglu, line 83:
```suggestion
        root: classes[orientation],
```

packages/picasso/src/Tab/styles.ts
ozgurkececioglu: Hmm, I think you are right. We don't have this shadow in the codebase. But still, please define a variable and use it for `boxShadow` to increase readability 🙇‍♂️ 

About where to store this value (theme or just this file), we can discuss when others are reviewing this PR

ozgurkececioglu:
@iatanas0v 
I was referring to this part
Screen Shot 2022-09-13 at 10 46 22

In the example stories, there is no edge. So, I'm having doubt that it should be stay like that or we should provide a default border 🤔

packages/picasso/src/Tabs/Tabs.tsx
```diff
@@ -18,38 +18,51 @@ export interface Props
...
-      onChange={onChange}
-      value={value}
-      variant='scrollable'
-      action={action}
-      scrollButtons='auto'
-      ScrollButtonComponent={TabScrollButton}
-    >
-      {children}
-    </MUITabs>
+    <TabsOrientationContext.Provider value={orientation || 'horizontal'}>
```
ozgurkececioglu, line 44: Again, I don't think covering the default value is not necessary. Wdyt?

iatanas0v: In this case, it's here to please the TS. Another option is to use a not-null assertion :thinking: `value={orientation!}` as this should never be `null`. WDYT?

image

packages/picasso/src/Tab/styles.ts
```diff
@@ -46,10 +42,53 @@ PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
...
+    horizontal: {
+      '&:not(:last-child)': {
+        marginRight: '2em',
+      },
+    },
+    vertical: {
+      borderRadius: `${sizes.borderRadius.medium} 0 0 ${sizes.borderRadius.medium}`,
+      margin: '0.5rem 0',
+      overflow: 'hidden',
+      padding: `${rem('9px')} 0 ${rem('9px')}`,
```
ozgurkececioglu, line 59: I think you made this change to make it 38px height exactly, right?

iatanas0v: Yes, exactly. In the design there's less padding but the `line-height` of the label is `22px` and I saw `22px` is not something supported by `<Typography>`. I thought about using a custom css to force it to 22px but that would create visual inconsistency in cases of long labels (wrapping on few lines)

packages/picasso/src/Tabs/story/index.jsx
```diff
@@ -26,6 +26,10 @@ page
     title: 'Default',
     takeScreenshot: false,
   })
+  .addExample('Tabs/story/Vertical.example.tsx', {
+    title: 'Vertical',
+    takeScreenshot: false,
```
ozgurkececioglu, line 31: I think we can have this as a screenshot. We don't have vertical tabs in happo

packages/picasso/src/Tabs/test.tsx
```diff
@@ -61,6 +62,21 @@ describe('Tabs', () => {
     expect(container).toMatchSnapshot()
   })
 
+  it('renders in vertical orientation', () => {
```
ozgurkececioglu, line 65: What is the purpose of this test? I mean, are you trying to test that none of the tabs are visible when there is no value?

packages/picasso/src/Tabs/Tabs.tsx
ozgurkececioglu: Yeah, we generally use not-null assertion (until we decide to handle `defaultProps` in a better TS way). I think you can follow that for now

packages/picasso/src/Tabs/test.tsx
iatanas0v: I just wanted to have a snapshot test for the vertical option, that's all. I literally copied the above test case and changed orientation.

Looking now at it, I guess, we can strip it down to just `expect(container).toMatchSnapshot()`

ozgurkececioglu (approved):
Tested on my local. Looking good 👍 
Thanks for the contribution once again ❤️

packages/picasso/src/Tab/styles.ts
```diff
@@ -1,8 +1,10 @@
 import { Theme, createStyles } from '@material-ui/core/styles'
 import { rem } from '@toptal/picasso-shared'
-import { PicassoProvider } from '@toptal/picasso-provider'
+import { PicassoProvider, palette } from '@toptal/picasso-provider'
 
-PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
+const ACTIVE_SHADOW = `4px 0 0 ${palette.grey.lightest}, 0 0 4px rgba(0, 0, 0, 0.08)`
```
denieler, line 5: this doesn't seem right if this shadow is not taken from the `theme`

iatanas0v: There's no similar shadow in the theme. Do you think we should add this one to the theme ? :thinking:

packages/picasso/src/Tab/styles.ts
```diff
@@ -46,10 +44,53 @@ PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
...
+    horizontal: {
+      '&:not(:last-child)': {
+        marginRight: '2em',
+      },
+    },
+    vertical: {
+      borderRadius: `${sizes.borderRadius.medium} 0 0 ${sizes.borderRadius.medium}`,
+      margin: '0.5rem 0',
+      overflow: 'hidden',
+      padding: `${rem('9px')} 0 ${rem('9px')}`,
```
denieler, line 61: can we use `em` here?

packages/picasso/src/Tab/styles.ts
```diff
@@ -46,10 +44,53 @@ PicassoProvider.override(({ breakpoints, palette }: Theme) => ({
...
+          top: 0,
+          bottom: 0,
+          width: '3px',
+        },
+      },
+    },
+    wrapper: {
+      '$vertical &': {
+        marginLeft: '1rem',
+        marginRight: '2rem',
```
denieler, line 93: can we move those styles to the `vertical` class?

packages/picasso/src/Tab/styles.ts
iatanas0v: Sure. Just for the `padding`? Can you give me a quick explanation when do you use `rem` and when `em` so I know for the future :pray:

denieler: Yes, we use `rem` only for `font-size` and `px` for `1px` cases (1px border, for example). All the rest places should use `em`

packages/picasso/src/Tab/styles.ts
denieler: @iatanas0v if this shadow is still not in Picasso - we need to check whether this shadow is in BASE. If yes - we should add this shadow to the list of other shadows in theme 👍 If not - we need to change this shadow to one, which is in the BASE most likely

OndrejTuma (approved):
I agree with Daniel about the standardized shadow,
otherwise it looks good to me 👍 
Thanks for contribution 🙇

packages/picasso/src/Tab/styles.ts
iatanas0v: @denieler After a chat with BASE it was decided to reuse an existing shadow: https://toptal-core.slack.com/archives/CCC3GP6CC/p1663224856285379?thread_ts=1662703900.679159&cid=CCC3GP6CC

packages/picasso/src/Tabs/styles.ts
```diff
@@ -18,6 +18,15 @@ PicassoProvider.override(({ palette }: Theme) => ({
...
       },
     },
+    vertical: {
+      '&::after': {
+        display: 'none',
+      },
+
+      '& $indicator': {
+        display: 'none',
+      },
```
denieler, line 28: nit: idk, I have a bit mixed feelings about those classes, because it seems to me that the abstraction of `Tabs` is not good enough if in some special cases we need to hide some elements 🤔 usually it's better to do opposite - to show something based on a condition, instead of hiding 🤔

denieler (approved):
Haven't tested, but code looks good to me now 👍

iatanas0v:
@toptal/frontend-experience-eng can someone review and approve the Happo snapshot as well :pray:

packages/picasso/src/Tabs/styles.ts
iatanas0v: @denieler I hear you. I managed to avoid removing the "separator" line https://github.com/toptal/picasso/pull/3096/commits/6410c5ebb3c32c9004355565cd4ebeb564ed5e49 but I don't think it's easy to achieve the same for the indicator :unamused: 

I tried to reuse the indicator, just to move it to the left side (beginning of the tabs), but this way we can't match the design exactly as we can't have the exact same border radius and clip-effect from the tab :\

PR #3168, opened 2022-10-23
fix: get rid of style functions

ozgurkececioglu:
[FX-3191]

### Description

This PR is about removing dynamic style functions due to not working after the latest storybook update locally.

### How to test

- Green CI and happo reports
- Nothing should be different visually

### Screenshots

- no visual changes

### Development checks

- [x] Self reviewed
- Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-3191]: https://toptal-core.atlassian.net/browse/FX-3191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/Avatar/AvatarWrapper/styles.ts
+++ b/packages/picasso/src/Avatar/AvatarWrapper/styles.ts
@@ -52,7 +52,7 @@ const generateSizeClassNames = () => {
 const generateCornerClassNames = () => {
   const classNames: Record<string, CSSProperties> = {}
 
-  for (const [size, { cornerSize }] of Object.entries(SETTINGS)) {
+  Object.entries(SETTINGS).forEach(([size, { cornerSize }]) => {
     const className = `corner${capitalize(size)}`
     const clipPath = `polygon(0 0, 100% 0, 100% 100%, ${cornerSize} 100%, 0 calc(100% - ${cornerSize}))`
 
@@ -62,7 +62,7 @@ const generateCornerClassNames = () => {
       // be resolved - https://github.com/cssinjs/css-vendor/issues/74
       '-webkit-clip-path': clipPath,
     }
-  }
+  })
 
   return classNames
 }
--- a/packages/picasso/src/Avatar/AvatarWrapper/styles.ts
+++ b/packages/picasso/src/Avatar/AvatarWrapper/styles.ts
@@ -1,5 +1,7 @@
 import { SizeType } from '@toptal/picasso-shared'
 import { createStyles, Theme } from '@material-ui/core/styles'
+import capitalize from '@material-ui/core/utils/capitalize'
+import { CSSProperties } from '@material-ui/core/styles/withStyles'
 
 const SETTINGS = {
   xxsmall: {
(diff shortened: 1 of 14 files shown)


packages/picasso/src/Dropdown/Dropdown.tsx
```diff
@@ -258,9 +258,9 @@ export const Dropdown = forwardRef<HTMLDivElement, Props>(function Dropdown(
             <div onClick={close}>
               <Grow in={isOpen} appear>
                 <Paper
+                  style={contentStyle}
```
ozgurkececioglu, line 261: I'm not sure which one is more efficient than the other. I couldn't come up with a different approach for `contentStyle` style function. Any suggestions?

TomasSlama (approved):
The code looks great!, No Happo changes ❤️.  I have tested all affected files in development

packages/picasso/src/Avatar/AvatarWrapper/styles.ts
```diff
@@ -24,6 +26,47 @@ const SETTINGS = {
...
+      const ratio = size === 'large' ? 3 / 4 : 2 / 3
+      const widthRatio = variant === 'portrait' ? ratio : 1
+      const heightRatio = variant === 'landscape' ? ratio : 1
+
+      classNames[className] = {
+        width: `${dimensions * widthRatio}em`,
+        height: `${dimensions * heightRatio}em`,
+      }
+    })
+  })
```
LashaJini, line 47: This decreases performance in terms that we are running unnecessary calculations (called more than once, other function calls, float number multiplication) when state is updated in React component. For demonstration purposes, I've inserted `console.log('oi')` inside second inner iteration, just to show that it is called `VARIANTS` * `SETTINGS.keys().length` times (in this case 15)

https://user-images.githubusercontent.com/72510037/197479632-7548e9e8-7d02-44a5-8389-f87a7f00eb87.mp4



Maybe we could lazily initialize it using  lazy initial state in AvatarWrapper: 

```jsx
const [state, _] = useState(() => {
  const initialState = someExpensiveComputation(props);
  return initialState;
});
```

packages/picasso/src/Avatar/AvatarWrapper/styles.ts
```diff
@@ -24,6 +26,47 @@ const SETTINGS = {
...
+
+  return classNames
+}
+
+const generateCornerClassNames = () => {
+  const classNames: Record<string, CSSProperties> = {}
+
+  Object.entries(SETTINGS).forEach(([size, { cornerSize }]) => {
+    const className = `corner${capitalize(size)}`
+    const clipPath = `polygon(0 0, 100% 0, 100% 100%, ${cornerSize} 100%, 0 calc(100% - ${cornerSize}))`
```
LashaJini, line 57: Same here. And it might be worse here :thinking: , depending on time complexity of `corner` and `polygon`

packages/picasso/src/Avatar/AvatarWrapper/styles.ts
OndrejTuma: Maybe it doesnt need to be function at all, or called once in root level and store reference into a const

LashaJini: Yeah, that would be a better solution :+1: . Don't see any problems with that

LashaJini: But it might problematic because we are calling it inside `createStyles`.

OndrejTuma (approved):
code seems legit 👌

packages/picasso/src/Avatar/AvatarWrapper/styles.ts
ozgurkececioglu: excellent catch 👍 I replaced the generator functions with const and reducer. Please take a look once again @LashaJini @OndrejTuma 

at least we won't be iterating over and over again.

packages/picasso/src/Avatar/AvatarWrapper/styles.ts
ozgurkececioglu: we are just generating CSS here, so I don't think it is different than `display: flex`. But about iterations, I would definitely agree 👍

LashaJini (approved):
Looks good. Thank you Oz for the fix :muscle:

PR #3180, opened 2022-10-26
chore: comply dropzone with base design

LashaJini:
[FX-3177](https://toptal-core.atlassian.net/browse/FX-3177?atlOrigin=eyJpIjoiNmMzNDcwZjVmMDdhNDkwMmIyNjkyNjlhODJlZTllNTYiLCJwIjoiaiJ9)

### Description

- In **BASE** we have a different design of failed upload than we have implemented in Picasso.

### How to test

- `Dropzone`'s [Upload Failed example](https://picasso.toptal.net/fx-3177-dropzone-error-state/?path=/story/components-dropzone--dropzone#upload-failed) should be complied with the one in BASE design
- also: Screenshot from 2022-10-28 10-12-40, error will be shown this way, from now on, because we are deprecating `errorMessages` (I've commented below why)

### Screenshots

### Development checks

- [x] Self reviewed
- Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

### DIFF
--- a/packages/picasso/src/Dropzone/Dropzone.tsx
+++ b/packages/picasso/src/Dropzone/Dropzone.tsx
@@ -44,7 +44,7 @@ export interface Props extends BaseProps {
   value?: FileUpload[]
   /**
    * @deprecated **Use value.error instead.**
-   * Provides reasons why files couldn't be dropped into dropzone
+   * Provide reasons why files couldn't be dropped into dropzone
    */
   errorMessages?: string[]
   focused?: boolean
--- a/packages/picasso/src/Dropzone/Dropzone.tsx
+++ b/packages/picasso/src/Dropzone/Dropzone.tsx
@@ -43,8 +43,8 @@ export interface Props extends BaseProps {
   /** Value uses the File interface. */
   value?: FileUpload[]
   /**
-   * @deprecated Use value.error instead
-   * Reasons why files couldn't be droped into dropzone
+   * @deprecated **Use value.error instead.**
+   * Provides reasons why files couldn't be dropped into dropzone
    */
   errorMessages?: string[]
   focused?: boolean
--- a/packages/picasso/src/Dropzone/Dropzone.tsx
+++ b/packages/picasso/src/Dropzone/Dropzone.tsx
@@ -11,6 +11,7 @@ import Container from '../Container'
 import FileList from '../FileList'
 import Typography from '../Typography'
 import { FileUpload, DropzoneOptions } from './types'
+import { usePropDeprecationWarning } from '../utils/use-deprecation-warnings'
 import styles from './styles'
 
 export interface Props extends BaseProps {
@@ -43,7 +44,7 @@ export interface Props extends BaseProps {
   /** Value uses the File interface. */
   value?: FileUpload[]
   /**
-   * @deprecated **Use value.error instead.**
+   * @deprecated **Use value[n].error instead.**
    * Provide reasons why files couldn't be dropped into dropzone
    */
   errorMessages?: string[]
@@ -98,6 +99,14 @@ export const Dropzone = forwardRef<HTMLInputElement, Props>(function Dropzone(
   })
 
   const classes = useStyles()
+
+  usePropDeprecationWarning({
+    props,
+    name: 'errorMessages',
+    componentName: 'Dropzone',
+    description:
+      'Use the `value[n].error` prop instead. `errorMessages` is deprecated and will be removed in the next major release.',
+  })
 
   return (
     <Container style={style} ref={ref} className={className}>
--- a/packages/picasso/src/Dropzone/Dropzone.tsx
+++ b/packages/picasso/src/Dropzone/Dropzone.tsx
@@ -42,7 +42,10 @@ export interface Props extends BaseProps {
   validator?: DropzoneOptions['validator']
   /** Value uses the File interface. */
   value?: FileUpload[]
-  /** Reasons why files couldn't be droped into dropzone */
+  /**
+   * @deprecated Use value.error instead
+   * Reasons why files couldn't be droped into dropzone
+   */
   errorMessages?: string[]
   focused?: boolean
   hovered?: boolean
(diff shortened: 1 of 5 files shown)


packages/picasso/src/Dropzone/Dropzone.tsx
```diff
@@ -42,7 +42,10 @@ export interface Props extends BaseProps {
   validator?: DropzoneOptions['validator']
   /** Value uses the File interface. */
   value?: FileUpload[]
-  /** Reasons why files couldn't be droped into dropzone */
+  /**
+   * @deprecated Use value.error instead
+   * Reasons why files couldn't be droped into dropzone
+   */
```
LashaJini, line 48: We are using `FileUpload`'s `value` prop which already has `error` field inside and _error message placement complies with base design_. We should remove all the `errorMessage` references in this file and all related styles, which would be a BREAKING change ofc.

packages/picasso/src/Dropzone/story/Uploader.example.tsx
```diff
@@ -69,18 +75,18 @@ const useFiles = ({ maxFiles }: { maxFiles: number }) => {
...
     if (rejectedFiles.length) {
-      setError(
-        rejectedFiles.map(
-          ({ errors, file }) =>
-            `${file.name}: ${errors.map(error => error.message).join(', ')}`
-        )
-      )
-    } else {
-      setError([])
+      setFiles((prevFiles: DropzoneFileUpload[]) => [
```
ozgurkececioglu, line 78: Previously, acceptedFiles and rejectedFiles were different state variables. Now, they will be kept in the same state `files`. In this function, you may call the `setFiles` function twice with different values (both acceptedFiles and rejectedFiles exist) which is not a good thing. Most probably react will use batching to reflect one update on state but still, it may cause a bug in the future after some following updates on this function.

Do you think it would be a better idea to handle everything and call `setFiles` once at the end?

ozgurkececioglu (commented):
I'm pretty sure that Dropzone has more compliance issues like hovering and focusing states on the input. But this PR is just covering error state, so nice work 👍 

I have only one comment 🙇‍♂️

packages/picasso/src/Dropzone/story/Uploader.example.tsx
LashaJini: > Do you think it would be a better idea to handle everything and call setFiles once at the end like a real transaction?
This sounds good, but I don't think we will be able to have only 1 call to `setFiles` in this function.

ozgurkececioglu: > acceptedFiles and rejectedFiles were different state variables
by this, I meant that `acceptedFiles` were being kept as `files` and `rejectedFiles` were being kept as `errors` so we weren't setting them as one state variable: `files`.

I'm saying that in a function, if you call the same `setState` function multiple times, react will batch(merge) them and apply the update only once, which is good. but I'm more in favor of doing the update only once, instead of believing in react. So, I would prefer defining a `files` array in the beginning, doing all changes on it, and calling `setFiles` at the end once. My reasoning is that this way, the changes you are applying would be out of async state updates. Think about this example: https://codesandbox.io/s/strange-tereshkova-6gflxp?file=/src/App.js. You are using both `files` and `setFiles` and it can cause a side effect while calling second `setFiles`. More clear?

ozgurkececioglu: ok after a good discussion, @LashaJini made me realize that there are interval usages and the code is actually working as expected. Sorry for being blind 🙇

ozgurkececioglu (approved):
Tested via temploy, working as expected
Thank you @LashaJini 🙇‍♂️

TomasSlama (approved):
Tested locally, looks good 👍

packages/picasso/src/Dropzone/story/Uploader.example.tsx
```diff
@@ -69,18 +75,18 @@ const useFiles = ({ maxFiles }: { maxFiles: number }) => {
...
+          const error = errors.map(error => error.message).join(', ')
+
+          return { error, file }
+        }),
+      ])
     }
   }
 
-  const removeFile = (fileName: string, fileIndex: number) => {
+  const removeFile = (_fileName: string, fileIndex: number) => {
```
separatio, line 89: Is it really necessary to keep this `_filename` variable? :thinking: It seems redundant to me.

packages/picasso/src/Dropzone/Dropzone.tsx
separatio, line 48: This looks kinda bad in the documentation. Can you please fix it somehow?
I'd be fine with removing `Reasons why files couldn't be droped into dropzone` altogether.

image

separatio (commented):
I'd also touch up the documentation for `value` since you are referring it. `Value uses the File interface.` This is not very descriptive to me. Do you think this should be done separately?

packages/picasso/src/Dropzone/story/Uploader.example.tsx
LashaJini: It is passed to `onRemove` prop in `Dropzone`, which internally is passing to `FileList''s prop `onItemRemove`, and it is required by it's type: `onItemRemove?: (fileName: string, index: number) => void`

https://github.com/toptal/picasso/blob/master/packages/picasso/src/FileList/FileList.tsx#L13

packages/picasso/src/Dropzone/Dropzone.tsx
separatio: I mean, you can just put a `.` to separate the two sentences and reformulate them a bit for them to make sense. No need for anything too drastic.

LashaJini: Done.

separatio (approved):

PR #3224, opened 2022-11-12
Fx 3139 page hamburger

OndrejTuma:
[FX-3139]

### Description

Add `PageHamburger` as a standalone component and use it to display `SidebarMenu` on small screens through portal.

### How to test

- check the code - hamburger from SidebarMenu is moved to the PageHamburger component 
- check locally that the [Layout Page example with SidebarMenu](http://localhost:9001/?path=/story/layout-page--page#default) looks and works the same

### Development checks

- Add changeset according to [guidelines](https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md) (if needed)
- Ensure that deployed demo has expected results and good examples
- [x] Self reviewed
- Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-3139]: https://toptal-core.atlassian.net/browse/FX-3139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/picasso/src/PageHamburger/PageHamburgerPortal.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { getHamburgerContainer } from './utils'
+
+type Props = {
+  children: React.ReactNode
+}
+
+const PageHamburgerPortal = ({ children }: Props) => {
+  const { hamburgerId } =
+    useContext<PageHamburgerContextProps>(PageHamburgerContext)
+  const container = useCallback(
```
ozgurkececioglu, line 16: I'm confused, your purpose is to memoize the result right? Wouldn't it be meaningful to use `useMemo`?

Another idea, wdyt about using a `useRef(null)` and assign it in `useEffect` with `[hamburgerId]` dependency array? I remember a comment related to using ref in your previous (big) PR but can't remember now. I think using ref would be more convenient 🙇‍♂️

packages/picasso/src/PageHamburger/hooks/use-hamburger.ts
```diff
@@ -0,0 +1,20 @@
...
+  const [showHamburger, setShowHamburger] = useState<boolean>(false)
+
+  const context = useMemo(
+    (): PageHamburgerContextProps => ({
+      hamburgerId,
+      showHamburger,
+      setShowHamburger,
+    }),
+    [hamburgerId, showHamburger, setShowHamburger]
+  )
```
ozgurkececioglu, line 15: Are you sure this memoization helps to avoid re-rendering for children? I mean the components that use this context should be re-rendered when `hamburgerId` and `showHamburger` changed (btw, `setShowHamburger` is not necessary because react assures us that `setState` functions are immutable already) right? I would double-check that it will be meaningful to do this memorization.

packages/picasso/src/PageHamburger/hooks/use-hamburger.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  return { context }
```
ozgurkececioglu, line 17: Why are we returning an object instead of the context itself?

packages/picasso/src/PageHamburger/styles.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+    responsiveWrapper: {
+      position: 'fixed',
+      top: '0.75em',
+      left: '0.75em',
+      zIndex: zIndex.appBar,
+    },
+    responsiveWrapperContent: {
+      backgroundColor: palette.grey.lighter,
+      boxShadow: `inset -1px 0px 0px 0px ${palette.grey.lighter2}`,
```
ozgurkececioglu, line 19: As I saw comments related to these inline values multiple times in my PRs that it would be better to extract them as a variable so people would understand what was the intention to provide inline values.

packages/picasso/src/PageSidebar/PageSidebar.tsx
```diff
@@ -114,6 +83,10 @@ export const PageSidebar = forwardRef<HTMLDivElement, Props>(function Sidebar(
     }
   }, [setHasSidebar])
 
+  useEffect(() => {
+    setShowHamburger?.(true)
```
ozgurkececioglu, line 87: nit: I'm not a fan of conditional calling functions like this due to decrease readability. I would prefer 

```suggestion
    if (setShowHamburger) {
        setShowHamburger(true)
    }
```

I remember discussing this with Daniel and he also said that it is not that common to use this conditional function call approach, so I have an architect opinion at least 🙂 still up to you 🙇‍♂️

ozgurkececioglu (commented):
Checked temploy and looks cool 👍 

I have some comments to discuss 🙇‍♂️

packages/picasso/src/PageHamburger/PageHamburgerPortal.tsx
```diff
@@ -0,0 +1,24 @@
...
+} from './PageHamburgerContext'
+import { getHamburgerContainer } from './utils'
+
+type Props = {
+  children: React.ReactNode
+}
+
+const PageHamburgerPortal = ({ children }: Props) => {
+  const { hamburgerId } =
+    useContext<PageHamburgerContextProps>(PageHamburgerContext)
```
ozgurkececioglu, line 15: I saw that this context will be used multiple times so wdyt about

```suggestion
    usePageHamburgerContext()
```

like here: https://github.com//toptal/picasso/blob/2d76bbf2e1b565829145f3fd338ff86c9d198f6b/packages/picasso/src/PageSidebar/SidebarContextProvider.tsx#L45

packages/picasso/src/Page/Page.tsx
```diff
@@ -15,6 +16,7 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
   centered?: boolean
   /** Children components (`Page.TopBar`, `Page.Content`, `Page.Footer`) */
   children: ReactNode
+  hamburgerId?: string
```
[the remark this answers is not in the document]
TomasSlama: I think it is not necessary, it is primarily for our storybook when we need to showcase multiple bars on one page. In applications, there will most likely be only one TopBar 🤔

packages/picasso/src/PageHamburger/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const getHamburgerContainer = (hamburgerId: string) => {
```
TomasSlama, line 1: with a better name, this function could be actually used for any other HTML element. How about just `getElementById` 🤔

TomasSlama (commented):
Looks good, left some comments.

packages/picasso/src/Page/Page.tsx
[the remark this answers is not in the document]
OndrejTuma: Exactly ☝️  there should be only one TopBar on the page. Its an internal prop just to make our storybook examples easier

packages/picasso/src/PageHamburger/PageHamburger.tsx
```diff
@@ -0,0 +1,70 @@
...
+import ButtonCircular from '../ButtonCircular'
+import PageHamburgerContext, {
+  PageHamburgerContextProps,
+} from './PageHamburgerContext'
+import Dropdown from '../Dropdown'
+import { Close16, Overview16 } from '../Icon'
+import { useBreakpoint } from '../utils'
+import styles from './styles'
+
+type HamburgerContentProps = {
```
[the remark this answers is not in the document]
OndrejTuma: Good catch. I think this was a name from a different place and got copy+pasted

packages/picasso/src/PageSidebar/PageSidebar.tsx
OndrejTuma: Done 👍

ozgurkececioglu (approved):
Thanks for the updates, tested on my local and looking good 👍

packages/picasso/src/PageHamburger/styles.ts
OndrejTuma: ahh, nice 👍 I'll add that, thanks for the pointer 🙇

packages/picasso/src/PageHamburger/PageHamburgerContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+  children,
+  hamburgerId,
+}: Props) => {
+  const [showHamburger, setShowHamburger] = useState<boolean>(false)
+
+  const context = {
+    hamburgerId,
+    showHamburger,
+    setShowHamburger,
+  }
```
TomasSlama, line 28:
```suggestion
  const [isHamburgerVisible, setHamburgerVisiblity] = useState<boolean>(false)
  
  const showHamburger = () => setHamburgerVisiblity(true)

  const context = {
    hamburgerId,
    showHamburger,
    isHamburgerVisible,
  }
```

WDYT? `showHamburger` evocates a function in me 🤔

packages/picasso/src/PageHamburger/styles.ts
OndrejTuma: Done ✅

packages/picasso/src/PageHamburger/PageHamburgerContext.tsx
OndrejTuma: I changed the naming, also added the hook `usePortalToHamburger` please check it out 🙇

packages/picasso/src/PageHamburger/PageHamburger.tsx
```diff
@@ -0,0 +1,67 @@
...
+  id: string
+}
+
+const useStyles = makeStyles<Theme>(styles, {
+  name: 'PageHamburger',
+})
+
+const PageHamburgerContent = ({ id }: Props) => {
+  return <div id={id} />
+}
```
augustobmoura, line 22: This component feels a bit unnecessary, maybe we can remove it, wdyt?

packages/picasso/src/PageHamburger/hooks/use-portal-to-hamburger.ts
```diff
@@ -0,0 +1,19 @@
...
+ * hamburger menu.
+ * Should be used in conjunction with PageHamburgerPortal component.
+ * It sets hamburger menu to be visible in compact screens.
+ */
+const usePortalToHamburger = () => {
+  const { setIsHamburgerVisible } = useHamburgerContext()
+
+  useEffect(() => {
+    setIsHamburgerVisible(true)
+  }, [])
```
augustobmoura, line 16: Why do we need this hook? Sorry if I missed something

packages/picasso/src/PageHamburger/styles.ts
```diff
@@ -0,0 +1,39 @@
...
+        maxHeight: 'calc(100vh - 3rem)', // viewport minus header height
+      },
+
+      // height under which maxHeight menu starts to overflow
+      // and needs to reduce height dynamically to avoid overflow
+      '@media screen and (max-height: 585px)': {
+        maxHeight: 'calc(100vh - 4.5rem)', // viewport minus header height
+
+        [screens('small', 'medium')]: {
+          maxHeight: 'calc(100vh - 3rem)', // viewport minus header height
```
augustobmoura, line 34: Do we need these values in `rem`? Can we use `em` instead? We had some problems while using `rem` in some users scenarios before

augustobmoura (commented):
I have some questions about the implementation, but code overall looks good!

packages/picasso/src/PageHamburger/hooks/use-portal-to-hamburger.ts
TomasSlama: @augustobmoura My bad, I had hard time to explain on paper, so we had meeting yesterday with Ondra where I shared my concerns.

Having a custom hook has some benefits in this case:
1. we use it in multiple places
2. we can use a descriptive name, so it is clear what is the intent

This hook is called in SidebarMenu and PageTopBarMenu to propagate information to Hamburger, that it needs to be visible on small screens since both these components are portaled to the hamburger menu

packages/picasso/src/PageHamburger/styles.ts
TomasSlama: I think in this case, we need the exact value of header height, so for me `rem` is the correct way.

packages/picasso/src/PageHamburger/hooks/use-portal-to-hamburger.ts
OndrejTuma: @augustobmoura just to clarify the need for this _code_ altogether:
Hamburger menu itself is not visible by default. It should be visible only if there is either `SidebarMenu` or `TopBarMenu` present. Those components use this hook. And the hook is very simple, just sets the visibility to true as you see

packages/picasso/src/PageHamburger/styles.ts
OndrejTuma: This is not a new code by me, I simply moved it from SidebarMenu where it was previously implemented. And I agree that rems are solid and proven way in this component. Rem should not be a problem here, since it shouldnt be customizable by end users

packages/picasso/src/PageHamburger/PageHamburger.tsx
OndrejTuma: At first I thought this component might be more complex, but you're right that it does not have to be here at all.

OndrejTuma: Done ✅

packages/picasso/src/PageHamburger/styles.ts
augustobmoura: Got it, thanks for clearing that out @TomasSlama @OndrejTuma! :bow:

packages/picasso/src/PageHamburger/hooks/use-portal-to-hamburger.ts
augustobmoura: Ok, I think we can keep it

augustobmoura (approved):
Nice work @OndrejTuma! :+1:

PR #3251, opened 2022-11-24
feat: restyle counter component for RichTextEditor and multiline Input

OndrejTuma:
[FX-3260]

### Description

### How to test

### Screenshots

### Development checks

- [ ] Make sure that additions and changes on design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core
- [ ] Self reviewed
- [ ] Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


packages/picasso-forms/src/Field/Field.tsx
```diff
@@ -120,11 +121,19 @@ const Field = <
     shouldValidateOnSubmit,
   })
 
+  const hasCounter =
+    Boolean(
+      rest.multiline &&
+        (rest.counter === 'entered' || rest.limit) &&
+        input.value
+    ) || Boolean(rest.minLength || rest.maxLength)
```
denieler, line 129: this doesn't seem like a correct place to put the logic about the counter 🤔

OndrejTuma: The purpose of `hasCounter` is to decide where to put `autoSaveIndicator`. If **true** it goes to the component, otherwise to FormField.
Field component is a parent of FormField and Input, so I needed to make the call here

OndrejTuma: I might move the expression to some util function and describe it better.
But I didnt find any good css solution and since I didnt want to move counter from the input component - so I dont break their logic bound - I figured the best solution is to move the `autoSaveIndicator` as a kind of adnornment to either FormField or Input component itself.

packages/picasso/src/InputLimitAdornment/styles.ts
```diff
@@ -3,15 +3,13 @@ import { Theme, createStyles } from '@material-ui/core/styles'
 export default ({ palette }: Theme) =>
   createStyles({
     limiterLabel: {
-      color: palette.grey.main2,
       fontSize: '0.625rem',
-      lineHeight: '1rem',
+      lineHeight: 1,
+      marginTop: '0.25rem',
+      color: palette.grey.main2,
+      marginRight: '0.125rem',
```
TomasSlama, line 10:
```suggestion
      marginRight: rem('2px'),
```
just to improve readability

packages/picasso/src/RichTextEditorCounter/styles.ts
```diff
@@ -3,25 +3,10 @@ import { createStyles, Theme } from '@material-ui/core/styles'
...
-      lineHeight: '1.9em',
-
-      '&::before': {
-        content: '""',
-        width: 'calc(100% + 0.7em)',
-        height: '1px',
-        display: 'inline-block',
-        background: palette.grey.lighter2,
-      },
+      lineHeight: 1,
```
TomasSlama, line 8: It seems we are duplicating most of the styles with `InputLimitAdornment`, cannot we just reuse the same component? 🤔

packages/picasso-forms/src/Field/Field.tsx
OndrejTuma: @denieler I'm open to suggestions 🙂 but for the current solution we need to know in the parent - Field component - where to put the autoSaveIndicator

.changeset/fuzzy-knives-leave.md
```diff
@@ -0,0 +1,7 @@
+---
+'@toptal/picasso': minor
```
TomasSlama, line 2: 🤔 why minor? Isn't it "just" an update of styles, seems like a patch to me 🤷

.changeset/lemon-lamps-listen.md
```diff
@@ -0,0 +1,9 @@
+---
+'@toptal/picasso-forms': patch
+---
+
+---
+
+### FieldWrapper
+
+- add `autoSaveIndicator` prop for instance of `FormAutoSaveIndicator` component
```
TomasSlama, line 9: new prop seems like `minor` bump to me 🤔

ozgurkececioglu:
I'm not sure if it is a good idea to find the CSS issues before deciding the approach but the layout doesn't look correct when hint is provided 🤔 
 
https://user-images.githubusercontent.com/7733167/204792787-52d5e490-a704-4a97-ac45-0d93654efe14.mov

packages/picasso-forms/src/Field/Field.tsx
```diff
@@ -120,11 +121,19 @@ const Field = <
     shouldValidateOnSubmit,
   })
 
+  const hasCounter =
+    Boolean(
+      rest.multiline &&
+        (rest.counter === 'entered' || rest.limit) &&
```
ozgurkececioglu, line 127: Why only `entered`? Remaining is also counter right? I mean it will effect the layout as the same 🤔

ozgurkececioglu (commented):
I kept thinking about it pretty hard but couldn't find a better solution 😞 

I'm curious about naming the limit components after moving the component outside (from UI perspective). We can still introduce an internal component like `InputLimit` and use it in both Input and RichTextEditor, wdyt?

packages/picasso-forms/src/Field/Field.tsx
OndrejTuma: When counter is `entered` we dont care about `limit` prop - the counter will be visible even if you dont specify it. `limit` is required only for `remaining` counter type

packages/picasso/src/FormField/FormField.tsx
```diff
@@ -45,8 +49,13 @@ export const FormField = forwardRef<HTMLDivElement, Props>(function FormField(
...
       {children}
-      {error && <FormError className={classes.error}>{error}</FormError>}
-      {hint && <FormHint className={classes.hint}>{hint}</FormHint>}
+      <Container flex justifyContent='space-between'>
+        <Container>
+          {error && <FormError className={classes.error}>{error}</FormError>}
+          {hint && <FormHint className={classes.hint}>{hint}</FormHint>}
+        </Container>
+        {autoSaveIndicator}
+      </Container>
```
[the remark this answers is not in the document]
OndrejTuma: > I'm not sure if it is a good idea to find the CSS issues before deciding the approach but the layout doesn't look correct when hint is provided 🤔
> 
>  Screen.Recording.2022-11-30.at.15.09.49.mov

@ozgurkececioglu this is actually expected, since there has to be a space available for `autoSaveIndicator`, which has some height and margins. I agree the combination of using hint with autoSaveIndicator prop and counter does not look particularly good 🤔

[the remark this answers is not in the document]
OndrejTuma: > We can even optimize this like, if no `autoSaveIndicator` passed, we can drop a `Container` here

@ozgurkececioglu we can do that, I was afraid of increased code complexity if I introduce the condition. But I dont have a strong opinion here

packages/picasso/src/InputLimitAdornment/styles.ts
OndrejTuma: Do we have some guidelines on when to use this? Does it count only for really small values or should all the rems be written like this @TomasSlama ?

packages/picasso/src/RichTextEditorCounter/styles.ts
OndrejTuma: Good call 👍

packages/picasso/src/InputLimitAdornment/styles.ts
TomasSlama: I don't think we have a guideline, but it should be used to lower cognitive load. so for example, 1, 0.5, 0.25 rem is pretty clear, but otherwise, you need to think about it.

packages/picasso-forms/src/Field/Field.tsx
dmaklygin: I hope, @denieler means, we should create maybe a hook and put this logic inside that hook.

dmaklygin:
So, I think. It this case, we may use absolute position for the `Save` Container. It may solve problem for both cases. wdyt?

OndrejTuma:
> So, I think. It this case, we may use absolute position for the `Save` Container. It may solve problem for both cases. wdyt?

@dmaklygin we would risk the possibility of overlapping longer counter text with indicator's label (which is also customizable)
image

packages/picasso/src/FormField/FormField.tsx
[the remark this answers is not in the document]
ozgurkececioglu: Aren't we still rendering the auto-save container even if there is none? 🤔 @OndrejTuma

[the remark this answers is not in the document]
OndrejTuma: I changed it now so the container structure is there only when auto-save is present 👍

packages/picasso/src/RichTextEditorCounter/styles.ts
OndrejTuma: @TomasSlama after consideration, I decided not to do it - these components are quite different and creating an abstraction would bring more complexity. (the way error is displayed is different, there's a success state in InputLimitAdornment, content in InputLimitAdornment is displayed differently)

packages/picasso-forms/src/Form/Form.tsx
```diff
@@ -10,7 +10,9 @@ import {
   setIn,
   AnyObject,
   MutableState,
+  Mutator,
 } from 'final-form'
+import setFieldData from 'final-form-set-field-data'
```
ozgurkececioglu, line 15: In my opinion, we can introduce our own mutator as I did for `setActiveFieldTouched` in the past. I'm saying this because this way, you can specify what you want to set specifically and it would be more clear for further updates. Wdyt? 

The required code for this custom mutator would be really straight-forward

packages/picasso/src/FormField/FormField.tsx
```diff
@@ -16,21 +19,55 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
...
+>
+
+const FormFieldAdornments = ({
+  autoSaveIndicator,
+  children,
+  hasMultilineCounter,
+}: FormFieldAdornmentsProps) => {
+  const classes = useStyles()
+
+  if (!autoSaveIndicator) {
```
ozgurkececioglu, line 40: Yup, that's what I was thinking 🙂

packages/picasso/src/Input/Input.tsx
```diff
@@ -265,77 +249,97 @@ export const Input = forwardRef<HTMLInputElement, Props>(function Input(
 
   const classes = useStyles()
 
+  const showCounter = !!charsLength && hasCounter({ counter, limit })
+
+  useShowCounter({ name, showCounter, setFieldData })
```
ozgurkececioglu, line 254: I believe we all feel like this is really weird to change parent state from child like this (especially if the user decides to use Input without a form) but we have to accept this 👍

ozgurkececioglu (commented):
Tested via temploy and looks much better now 👍
I have some questions still to discuss and then I will approve.

Thanks again for this huge effort 🙇‍♂️

packages/picasso/src/RichTextEditorCounter/styles.ts
OndrejTuma: I refactored it @TomasSlama it wasnt as simple as I thought because InputLimitAdornment is not a simple displaying component, so I had to make some changes. Please check again 🙇

packages/picasso-forms/src/Form/Form.tsx
OndrejTuma: Thats a nice idea, I originally grab this one because its official mutator from final form that does exactly what I need, but writing our own will be easier to understand for devs 👍 and wont be used for anything else

packages/picasso/src/InputLimitAdornment/InputLimitAdornment.tsx
```diff
@@ -63,22 +62,14 @@ const InputLimitAdornment = (props: Props) => {
   })
 
   return (
-    <InputAdornment
+    <span
```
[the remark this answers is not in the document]
OndrejTuma: I refactored the code a bit more due to [Tomas request](https://github.com/toptal/picasso/pull/3251#discussion_r1034519944)
now its not using span anymore

packages/picasso-forms/src/Form/Form.tsx
OndrejTuma: @ozgurkececioglu done ✅

packages/picasso-forms/src/Form/Form.tsx
```diff
@@ -38,6 +38,21 @@ const setActiveFieldTouched = <
   }
 }
 
+const setHasMultilineCounter = <
```
ozgurkececioglu, line 41: You may start getting angry but should we create `mutators` folder and put these two(this and the `setActiveFieldTouched` one) custom mutators there? 😬

TomasSlama (approved):

packages/picasso-forms/src/Form/Form.tsx
```diff
@@ -38,6 +38,21 @@ const setActiveFieldTouched = <
...
+  FormValues = object,
+  InitialFormValues = Partial<FormValues>
+>(
+  args: any[],
```
augustobmoura, line 45: maybe we can type these args, wdyt? We know the first one is a name and the second is a a boolean, makes sense to type them like this?
```suggestion
  args: [name: string, hasCounter: boolean],
```

packages/picasso-forms/src/Form/Form.tsx
OndrejTuma: Done @ozgurkececioglu ✅

packages/picasso-forms/src/Form/Form.tsx
OndrejTuma: Good idea 👍 done ✅

packages/picasso/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -174,69 +174,73 @@ export const RichTextEditor = forwardRef<HTMLDivElement, Props>(
...
+            plugins={memoizedPlugins}
+          />
+          {hiddenInputId && enableFocusOnLabelClick(hiddenInputId)}
+        </Container>
         {counterMessage && (
-          <Counter error={counterError} message={counterMessage} />
+          <InputMultilineAdornment error={counterError}>
+            {counterMessage}
+          </InputMultilineAdornment>
         )}
```
OndrejTuma, line 249: This is the only real change here. I had to move the counter from the component's container, since it is now standalone component. I also reused `InputMultilineAdornment` so it inherits correct styling. It replaces `RichTextEditorCounter` component completely

ozgurkececioglu (approved):
I'm happy to see the progress along the way 🙂 
I'm even happier that we finally get rid of this ticket 😄

augustobmoura (approved):
Tested on Picasso and it works as expected! :+1: 
Thanks for all the work @OndrejTuma!

packages/picasso/src/FormField/FormField.tsx
```diff
@@ -45,9 +82,14 @@ export const FormField = forwardRef<HTMLDivElement, Props>(function FormField(
...
-      {hint && <FormHint className={classes.hint}>{hint}</FormHint>}
-      {fieldRequirements}
+      <FormFieldAdornments
+        autoSaveIndicator={autoSaveIndicator}
+        hasMultilineCounter={hasMultilineCounter}
+      >
+        {error && <FormError className={classes.error}>{error}</FormError>}
+        {hint && <FormHint className={classes.hint}>{hint}</FormHint>}
+        {fieldRequirements}
+      </FormFieldAdornments>
```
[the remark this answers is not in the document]
dmaklygin: the only thing is: inconsistency. save indicator is passed as a prop, however, `error` and `hint` are passed as children nodes

packages/picasso/src/FormField/styles.ts
```diff
@@ -13,6 +13,22 @@ export default () =>
...
+      paddingRight: '2rem',
+    },
+    autoSaveIndicator: {
+      position: 'absolute',
+      top: 0,
+      right: 0,
+
+      '&$hasMultilineCounter': {
+        top: '-0.875rem',
+      },
```
dmaklygin, line 27: 👍

packages/picasso/src/Input/Input.tsx
```diff
@@ -265,77 +289,93 @@ export const Input = forwardRef<HTMLInputElement, Props>(function Input(
...
+        }
+        endAdornment={
+          endAdornment || (
+            <EndAdornment
+              icon={icon}
+              iconPosition={iconPosition}
+              disabled={disabled}
+              limit={limit}
+              charsLength={charsLength}
+              multiline={multiline}
```
dmaklygin, line 348: can it be a multiline?

dmaklygin (approved):
Good work. Left minor comments

packages/picasso/src/FormField/FormField.tsx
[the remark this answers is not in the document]
OndrejTuma: True. It was introduced as a react node because the indicator itself can have customized label.
I'll check if it make sense to switch

packages/picasso/src/Input/Input.tsx
OndrejTuma: Input can be a multiline if its a textarea. In that case `EndAdornment` will return null

packages/picasso/src/FormField/FormField.tsx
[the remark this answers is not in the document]
OndrejTuma: OK we agreed with you and Oz to keep the solution as-is 👌

PR #3318, opened 2023-01-11
feat(Drawer): add body scroll lock

ascrazy:
[TOP-2735]

### Description

Decoupled body scroll management logic from ModalManager class to make it independent from modals and reusable. Added a prop to Drawer that allows to opt-in for body scroll lock enabled mode. 

### How to test

- Ensure modals lock body scrolling just like they would do it previously
- Ensure drawers lock body scrolling when rendered with `maintainBodyScrollLock=true`

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[TOP-2735]: https://toptal-core.atlassian.net/browse/TOP-2735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/utils/use-body-scroll-lock.ts
+++ b/packages/picasso/src/utils/use-body-scroll-lock.ts
@@ -1,0 +1,63 @@
+import { isBrowser } from '@toptal/picasso-shared'
+import { useEffect, useMemo } from 'react'
+
+const layers = new Set<number>();
+let scrollLock: { prevBodyOverflow: string } | undefined = undefined;
+
+export const useBodyScrollLock = (isLocked: boolean) => {
+  const layerId = useMemo(generateLayerId, [])
+
+  useEffect(() => {
+    if (isLocked) {
+      layers.add(layerId)
+      syncBodyScrollLock()
+    }
+
+    return () => {
+      layers.delete(layerId)
+      syncBodyScrollLock()
+    }
+  }, [layerId, isLocked])
+}
+
+const generateLayerId = (() => {
+  let count = 0
+
+  return () => {
+    count = count + 1
+    
+    return count;
+  }
+})()
+
+const syncBodyScrollLock = () => {
+  if (layers.size > 0) {
+    addBodyScrollLock()
+  } else {
+    removeBodyScrollLock()
+  }  
+}
+
+const addBodyScrollLock = () => {
+  if (!isBrowser()) {
+    return
+  }
+
+  if (!scrollLock) {
+    scrollLock = {
+      prevBodyOverflow: document.body.style.overflow,
+    }
+    document.body.style.overflow = 'hidden'
+  }
+}
+
+const removeBodyScrollLock = () => {
+  if (!isBrowser()) {
+    return
+  }
+
+  if (scrollLock) {
+    document.body.style.overflow = scrollLock.prevBodyOverflow
+    scrollLock = undefined
+  }
+}
--- a/.changeset/nice-donuts-check.md
+++ b/.changeset/nice-donuts-check.md
@@ -1,0 +1,9 @@
+---
+'@toptal/picasso': minor
+---
+
+---
+
+### Drawer
+
+- Add support for optional body scroll lock enabled mode
--- a/packages/picasso/src/Drawer/Drawer.tsx
+++ b/packages/picasso/src/Drawer/Drawer.tsx
@@ -11,6 +11,7 @@ import ButtonCircular from '../ButtonCircular'
 import Container from '../Container'
 import DrawerTitle from '../DrawerTitle'
 import { useIsomorphicLayoutEffect } from '../utils'
+import { useBodyScrollLock } from '../utils/use-body-scroll-lock'
 
 type AnchorType = 'bottom' | 'left' | 'right' | 'top'
 
@@ -33,6 +34,8 @@ export interface Props extends BaseProps {
   width?: WidthType
   /** Animation lifecycle callbacks. Backed by [react-transition-group/Transition](https://reactcommunity.org/react-transition-group/transition#Transition-props) */
   transitionProps?: TransitionProps
+  /** enable Drawer to maintain body scroll lock */
+  maintainBodyScrollLock?: boolean
 }
 
 const useStyles = makeStyles<Theme>(styles, { name: 'PicassoDrawer' })
@@ -46,12 +49,15 @@ export const Drawer = (props: Props) => {
     title,
     width = 'regular',
     transitionProps,
+    maintainBodyScrollLock,
     ...rest
   } = props
   const classes = useStyles()
   const { setHasDrawer } = useDrawer()
   const theme = useTheme()
   const container = usePicassoRoot()
+
+  useBodyScrollLock(Boolean(maintainBodyScrollLock && open))
 
   useIsomorphicLayoutEffect(() => {
     setHasDrawer(open)
--- a/packages/picasso/src/Drawer/story/WithBodyScrollLock.example.tsx
+++ b/packages/picasso/src/Drawer/story/WithBodyScrollLock.example.tsx
@@ -1,0 +1,43 @@
+import { Button, Container, List, Drawer, Checkbox } from '@toptal/picasso'
+import React, { useState } from 'react'
+
+const Example = () => {
+  const [open, setOpen] = useState(false)
+  const [maintainBodyScrollLock, setMaintainBodyScrollLock] = useState(true)
+
+  return (
+    <div>    
+      <Checkbox 
+        checked={maintainBodyScrollLock} 
+        onChange={(evt, val) => {
+          setMaintainBodyScrollLock(val)
+        }} 
+        label="Maintain body scroll lock" 
+      />
+      <Container top='small'>
+        <Button data-testid='trigger' onClick={() => setOpen(!open)}>
+          Show drawer
+        </Button>
+      </Container>
+      <Drawer
+        title='My Operational Issues'
+        open={open}
+        onClose={() => setOpen(false)}    
+        maintainBodyScrollLock={maintainBodyScrollLock} 
+      >
+        <Container data-testid='content' padded='medium'>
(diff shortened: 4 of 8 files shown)


packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,62 @@
+import { isBrowser } from '@toptal/picasso-shared'
+import { useEffect, useRef } from 'react'
+
+const layers: number[] = [];
```
OndrejTuma, line 4: nit: maybe using `Set` here would make things neater, wdyt?

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,62 @@
...
+let scrollLock: { prevBodyOverflow: string } | undefined = undefined;
+
+export const useBodyScrollLock = (isLocked: boolean) => {
+  const layerId = useRef(generateLayerId())
```
OndrejTuma, line 8: nit: just a small concern about `generateLayerId` being called every time a component using this hook is rendered. Memoization could help I think, but its probably not that big of deal. I'll leave it up to you

ascrazy: I never want this value to change and do not want it to affect lifecycle of the component, so IMO useRef is the best fit for that case. Not sure how exactly and what for I'd combine memoization with useRef...

TomasSlama (approved):
Tested in temploy, well done 👍

packages/picasso/src/utils/use-body-scroll-lock.ts
OndrejTuma: Something like this was my idea. The generated id without memoization can grow to high numbers, but it's probably not a big deal. Feel free to resolve this conversation 👌 
```suggestion
  const id = useMemo(generateLayerId, [])
  const layerId = useRef(id)
```

OndrejTuma (approved):
Tested and works as expected 👍 
Thanks for your contribution! 🎉

packages/picasso/src/utils/use-body-scroll-lock.ts
augustobmoura: I feel like Ondra's concern is valid, we are incrementing the counter on every re-render, while I don't think this will cause any starvation on the counter (`Number.MAX_SAFE_INTEGER` is pretty big), I think of it as highly inefficient. The solution of `useMemo` that Ondra proposed would fix this behavior

augustobmoura: In fact, we might not even need a ref here, just the memoized number should be enough
```suggestion
  const layerId = useMemo(generateLayerId, [])
```
And later we can use it directly
```js
const lockedLayerId = layerId.current;
// Becomes
const lockedLayerId = layerId;
```

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,55 @@
...
+  }, [isLocked])
+}
+
+const generateLayerId = (() => {
+  let count = 0
+
+  return () => ++count
+})()
+
+const dropScrollLock = () => {
```
augustobmoura, line 31: This name is kinda confusing, I would imagine that is disabling the scroll-lock, when in reality is introducing it, maybe we could have a better name for it, wdyt? Maybe `scrollLock` it's already enough because it is a verb after all, or maybe `doScrollLock`?

packages/picasso/src/utils/use-body-scroll-lock.ts
ascrazy: Updated

ascrazy: We actually have the same pattern used in [Modal.tsx#L134](https://github.com/toptal/picasso/blob/master/packages/picasso/src/Modal/Modal.tsx#L134) and I assume it was never an issue.

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,53 @@
...
+let scrollLock: { prevBodyOverflow: string } | undefined = undefined;
+
+export const useBodyScrollLock = (isLocked: boolean) => {
+  const layerId = useMemo(generateLayerId, [])
+
+  useEffect(() => {
+    if (isLocked) {
+      layers.add(layerId)
+      dropScrollLock()
+    }
```
augustobmoura, line 14: Shouldn't we also remove the layer in the case of `!isLocked`, imagine that the variable would flip mid-execution by some conditional we would need to remove the layer. (even though it should be rare, we need to think on it). We would also need to check if all the layers are removed and accordingly remove lift the scroll-lock, something like:
```suggestion
    if (isLocked) {
      layers.add(lockedLayerId)
    } else {
      layers.remove(lockedLayerId)
    }
    syncScrollLock()
```

Where `checkForScrollLock()` would be:
```js
const syncScrollLock = () => {
  if (layers.size > 0) {
    doScrollLock()
  } else {
    liftScrollLock()
  }
}
```

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    return () => {
+      layers.delete(layerId)
+      if (layers.size === 0) {
+        liftScrollLock()
+      }
```
augustobmoura, line 20: I would also wrap this on the `syncScrollLock` function I mentioned on the previous comment

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (!isBrowser()) {
+    return
+  }
+
+  if (!scrollLock) {
+    scrollLock = {
+      prevBodyOverflow: window.getComputedStyle(document.body).overflow,
+    }
+    document.body.style.overflow = 'hidden'
+  }
```
augustobmoura, line 41: I know that this was the previous behavior, but it doesn't feel right, `getComputedStyle()` would get the value from stylesheets which is not the true value for `style="overflow: ..."`, setting it back to `style.overflow` feels like a mistake, what do you think?

For example we could have this structure:
```html
<div class="foo">
  ...
</div>
<style>
.foo {
  overflow: hidden;
}
</style>
```
After scroll locking we would get `overflow: hidden` from the stylesheet and when unlocking we would set it back to the style attribute. The structure would become:
```html
<div class="foo" style="overflow: hidden">
  ...
</div>
<style>
.foo {
  overflow: hidden;
}
</style>
```
While this is rare problem it could cause race conditions when classes are being added or removed from `body`, or any dynamic stylesheets for that matter. I would suggest just using the prop itself, instead of `getComputedStyles`:
```suggestion
  if (!scrollLock) {
    scrollLock = {
      prevBodyOverflow: document.body.style.overflow,
    }
    document.body.style.overflow = 'hidden'
  }
```

packages/picasso/src/utils/use-body-scroll-lock.ts
augustobmoura: We could hide the implementation behind our own `useId` hook, but I think is better to wait for Picasso moving on supporting [React 18 and beyond](https://beta.reactjs.org/reference/react/useId)

packages/picasso/src/utils/use-body-scroll-lock.ts
ascrazy: >Shouldn't we also remove the layer in the case of !isLocked, imagine that the variable would flip mid-execution...

this is totally covered by how useEffect works. any change to isLocked will result in useEffect being unmounted and mounted again. During unmount it will run cleanup callback (2) - which removes component instance from layers and syncs scroll lock, then on next mount with new isLocked=false component instance isn't added to layers thanks to condition (1). 

image

ascrazy:
Is this something with my PR or some integration tests are unstable? Seeing some tests randomly failing, seems unrelated to the changes I made...

TomasSlama:
@ascrazy sadly the tree component is from time to time flaky :(

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,59 @@
...
+      layers.delete(layerId)
+      syncBodyScrollLock()
+    }
+  }, [layerId, isLocked])
+}
+
+const generateLayerId = (() => {
+  let count = 0
+
+  return () => ++count
```
OleksandrNechai, line 26:
```suggestion
  return () => {
     count = count + 1
     return count
  }
```

packages/picasso/src/utils/use-body-scroll-lock.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    return () => {
+      layers.delete(layerId)
+      syncBodyScrollLock()
+    }
+  }, [layerId, isLocked])
+}
+
+const generateLayerId = (() => {
```
OleksandrNechai, line 23: What if we take this function out of the hook? Maybe we will not need useMemo?

ascrazy: We need a new id per instance. i don't think we can take it out of hook

packages/picasso/src/utils/use-body-scroll-lock.ts
augustobmoura: You are right, I got confused with the global state stuff, the cleanup callback should do the work! Thanks @ascrazy! :+1:

augustobmoura (approved):
Looks good to me! Thanks for the contribution @ascrazy! :+1:

PR #3483, opened 2023-04-04
feat: add nested lists support

augustobmoura:
[FX-3836]

### Description

We are adding support for nested lists styling from [Figma](https://www.figma.com/file/q2nvjiyO2CLqBv4DeJnU3i/Product-Library-Documentation?node-id=8-18). We are also adjusting the spacing to be consistent with said Figma styles

### How to test

- Check Happo and the new examples on Storybook

### Screenshots

No screenshots

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

- [NA] codemod is created and showcased in the changeset
- [NA] test alpha package of Picasso in StaffPortal

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-3836]: https://toptal-core.atlassian.net/browse/FX-3836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TomasSlama (commented):
we are missing changeset for provider and it looks weird that different types of nested lists starts on different position 🤔, is it expected?
image

TomasSlama:
also in RichTextEditor for ordered list I used different list types for nested ones, but I forgot to check with Matt, if the decision is that we should always have `numeric` type, we will need to update this also in RichTextEditor. Also we can have follow up to update RichTextEditor to also use native `disc` and `circle`, we are using svg there

augustobmoura:
> we are missing changeset for provider and it looks weird that different types of nested lists starts on different position thinking, is it expected? image

I'm still talking with Matt about it, but in a summary, unordered lists have a box surrounding them, while ordered lists don't have it
image

Matt is making some adjustments to it, when he is finished I will update it to follow the new design

augustobmoura:
> also in RichTextEditor for ordered list I used different list types for nested ones, but I forgot to check with Matt, if the decision is that we should always have `numeric` type, we will need to update this also in RichTextEditor. Also we can have follow up to update RichTextEditor to also use native `disc` and `circle`, we are using svg there

We need to support roman and alpha letters for numbering, but by default, it should be only numerals, I'm still talking with Matt about it though

.changeset/shiny-flowers-trade.md
```diff
@@ -0,0 +1,9 @@
+---
+'@toptal/picasso': minor
```
TomasSlama, line 2: we will probably need to update the peer dependency of the provider, as this will not work on the old provider, so this will probably be major

packages/picasso/src/List/__snapshots__/test.tsx.snap
```diff
@@ -6,109 +6,74 @@ exports[`List renders ordered list 1`] = `
...
+            class="MuiTypography-root PicassoTypography-bodyMedium PicassoListItem-content MuiTypography-body1"
           >
             list item N2
           </div>
         </div>
       </li>
-      <li>
+      <li
+        class=""
+        index="2"
```
TomasSlama, line 42: I think these two are not needed in DOM

augustobmoura: Fixed

.changeset/shiny-flowers-trade.md
augustobmoura: Good catch! :+1:  changed it to a major

augustobmoura:
@toptal-anvil ping reviewers

packages/picasso/src/List/List.tsx
```diff
@@ -23,9 +27,31 @@ const Tags = {
...
+  classes: Record<string, string>
+) => {
+  if (variant === 'unordered') {
+    const isOddLevel = level % 2 === 0
+
+    return isOddLevel ? classes.disc : classes.circle
+  }
+
+  return undefined
+}
```
dmaklygin, line 42: I cannot actually understand this function. it's called `getOrderedStyle`, 

but... inside the function: if `variant != 'unordered'` we return undefined.

maybe we should call the function: `getUnorderedStyle` ?

packages/picasso/src/List/__snapshots__/test.tsx.snap
```diff
@@ -6,109 +6,70 @@ exports[`List renders ordered list 1`] = `
     class="Picasso-root"
   >
     <ol
-      class="PicassoList-root"
+      class="PicassoList-root PicassoList-firstLevel"
       data-testid="ordered-list"
+      start="1"
```
TomasSlama, line 11: I think `start` should not be propagated to DOM

TomasSlama (approved):

packages/picasso/src/List/__snapshots__/test.tsx.snap
augustobmoura: `start` is actually a HTML attribute for the [`ol` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#attributes), it is necessary for the custom start functionality we [already have](https://picasso.toptal.net/fx-3836-list-add-styles-for-nested-lists/?path=/story/components-list--list#ordered-with-custom-start).

We can omit it when the value is 1 though, as it is already the default. I will add a patch to use it as undefined if the value is 1

packages/picasso/src/List/List.tsx
augustobmoura: Yeah, that's true, I misnamed it based on ordered

packages/picasso/src/List/__snapshots__/test.tsx.snap
TomasSlama: oh, didn't know, thanks

packages/picasso/src/List/List.tsx
```diff
@@ -23,9 +27,27 @@ const Tags = {
   ordered: 'ol',
 } as const
 
+const getDefaultType = (variant: Props['variant'], level: number) => {
+  if (variant === 'unordered') {
+    const isOddLevel = level % 2 === 0
```
dmaklygin, line 32: hm, I thought it should be even, not odd

packages/picasso/src/List/List.tsx
```diff
@@ -23,9 +27,27 @@ const Tags = {
...
+
+    return isOddLevel ? 'disc' : 'circle'
```
dmaklygin, line 34: btw, we can improve it like this: 
```suggestion
    return level % 2 ?  'circle' : 'disc'
```

packages/picasso/src/List/List.tsx
augustobmoura: That is correct, I will fix it

packages/picasso/src/List/List.tsx
augustobmoura: The first version I did was exactly like this, but I thought a bit about it, and putting on a variable makes the code more readable and self-documented. I prefer having the temporary variable, wdyt?

packages/picasso/src/ListItem/styles.ts
```diff
@@ -11,20 +14,41 @@ PicassoProvider.override(() => ({
...
   createStyles({
-    listContainer: {
-      lineHeight: '1.375em',
-      marginBottom: '0.5em',
-    },
-    ordered: {
-      minWidth: '1.25em',
-      marginRight: '0.75em',
+    content: {
+      paddingLeft: '8px',
```
dmaklygin, line 44: I think, we still should use em/rem

packages/picasso/src/ListItem/ListItem.tsx
```diff
@@ -7,80 +7,75 @@ import cx from 'classnames'
...
+  const { styleType: parentType } = useListContext()
   const {
     children,
-    icon,
     variant = 'unordered',
-    index = 1,
-    isLastElement,
-    ...rest
+    type,
+    icon = resolveIcon(type ?? parentType),
```
dmaklygin, line 47: shouldn't we memoize the value?

packages/picasso/src/List/styles.ts
```diff
@@ -1,15 +1,22 @@
...
   createStyles({
     root: {
-      listStyle: 'none',
-      padding: 0,
-      margin: 0,
-      fontSize: '1rem',
+      fontSize: typography.fontSizes.medium,
+      color: palette.common.black,
+      marginTop: '4px',
+      paddingLeft: '16px',
```
dmaklygin, line 13: I think, we should use em/rem here

dmaklygin (commented):
Good work! Left some comments

packages/picasso/src/ListItem/ListItem.tsx
augustobmoura: Not necessary in my opinion, the function is just an if condition, the overhead of having it memoized would be bigger than just calling the function directly

packages/picasso/src/List/List.tsx
dmaklygin: no objections against a temp variable :)

augustobmoura:
@toptal-anvil ping reviewers

packages/picasso/src/ListItem/styles.ts
dmaklygin: @TomasSlama I cannot recall, should we switch to px instead of em/rem?

packages/picasso/src/List/styles.ts
augustobmoura: Changed :+1:

packages/picasso/src/ListItem/styles.ts
augustobmoura: Changed :+1:

packages/picasso/src/ListItem/styles.ts
```diff
@@ -11,20 +15,44 @@ PicassoProvider.override(() => ({
...
+  const toEmUnit = (px: number) => em(px, fromPx(typography.fontSizes.medium))
+
+  return createStyles({
+    content: {
+      paddingLeft: toEmUnit(8),
     },
-    unordered: {
-      minWidth: '1rem',
+    listContainer: {
+      marginTop: toEmUnit(4),
```
dmaklygin, line 50:
```suggestion
     import { rem } from '@toptal/picasso-shared'
     
      marginTop: rem('4px'),
```

dmaklygin (approved):
Left one comment

packages/picasso/src/ListItem/styles.ts
augustobmoura: We had some problems with using `rem` in some usages. As we discussed, the current instruction is to phase out the `rem` usage for a widget-safe alternative. We might need to find a better alternative in the future. We also need to loop Matt into it

augustobmoura: As mentioned in the daily, let's move to `rem` for now and discuss about relative units some time on a meeting with the team

PR #3538, opened 2023-05-03
feat: add responsive grid spacing

sashuk:
[FX-3930]

### Description

[According to the BASE design](https://www.figma.com/file/q2nvjiyO2CLqBv4DeJnU3i/Product-Library-Documentation?type=design&node-id=4533%3A28468&t=NpdofDozEQJAvCJX-1), grid gutters depend on the screen size. This pull request adjusts gutters depending on the screen size unless spacing is explicitly specified by users.

Screenshot 2023-05-04 at 15 49 25

### How to test

- Test different screen sizes at https://picasso.toptal.net/FX-3930-make-grid-spacing-responsive/?path=/story/layout-grid--grid

### Screenshots

https://user-images.githubusercontent.com/1390758/236229973-8199a4fa-e35e-46dd-8fad-39771dcb6a1e.mp4

### Development checks

- [N/A] Ensure that deployed demo has expected results and good examples
- [x] Self reviewed
- [N/A] Covered with tests

**Breaking change**

- [N/A] codemod is created and showcased in the changeset
- [x] test alpha package of Picasso in StaffPortal (please see the https://github.com/toptal/staff-portal/pull/10339 and compare pages that have grids, for example, `talents?rate_buffer=0&limit=20&roles%5B%5D=designer&page=1&trigger=FilterUpdate`)

Screenshot 2023-05-04 at 18 32 59

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-3930]: https://toptal-core.atlassian.net/browse/FX-3930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/red-countries-thank.md
+++ b/.changeset/red-countries-thank.md
@@ -1,0 +1,5 @@
+---
+'@toptal/picasso': major
+---
+
+- grid spacing automatically adjusts to the screen size, unless explicitly specified by consumer. If you want to keep the old behavior, please explicitly set `spacing={32}` for `Grid` components.
--- a/.changeset/red-countries-thank.md
+++ b/.changeset/red-countries-thank.md
@@ -2,4 +2,6 @@
 '@toptal/picasso': major
 ---
 
+### Grid
+
 - grid spacing automatically adjusts to the screen size, unless explicitly specified by consumer. If you want to keep the old behavior, please explicitly set `spacing={32}` for `Grid` components.
--- a/packages/picasso-forms/src/Checkbox/__snapshots__/test.tsx.snap
+++ b/packages/picasso-forms/src/Checkbox/__snapshots__/test.tsx.snap
@@ -29,7 +29,7 @@ exports[`Form.Checkbox default render for checkboxes in a group 1`] = `
             name="checkbox-group"
           >
             <div
-              class="MuiGrid-root PicassoCheckboxGroup-grid MuiGrid-container PicassoGrid-container MuiGrid-direction-xs MuiGrid-align-items-xs-flex"
+              class="MuiGrid-root PicassoCheckboxGroup-grid MuiGrid-container PicassoGrid-container MuiGrid-spacing-xs MuiGrid-direction-xs MuiGrid-align-items-xs-flex"
             >
               <div
                 class="MuiGrid-root PicassoCheckboxGroup-gridItem MuiGrid-item"
--- a/packages/picasso-provider/src/Picasso/PicassoGlobalStylesProvider.tsx
+++ b/packages/picasso-provider/src/Picasso/PicassoGlobalStylesProvider.tsx
@@ -1,11 +1,13 @@
 import type { ReactNode, ForwardRefExoticComponent, RefAttributes } from 'react'
-import React, { useRef, useState, useCallback } from 'react'
+import React, { useEffect, useRef, useState, useCallback } from 'react'
 
 import type { RootContextProps } from './RootContext'
 import { RootContext } from './RootContext'
 import type { EnvironmentType, TextLabelProps } from '../types'
 import type { PicassoRootNodeProps } from './PicassoRootNode'
 import { isBrowser } from '../utils'
+import type { BreakpointKeys } from './config/breakpoints'
+import { useScreens } from './config/breakpoints'
 
 export interface PicassoGlobalStylesProviderProps extends TextLabelProps {
   children?: ReactNode
@@ -14,6 +16,14 @@ export interface PicassoGlobalStylesProviderProps extends TextLabelProps {
   >
   environment: EnvironmentType<'test' | 'temploy'>
   disableTransitions?: boolean
+}
+
+const breakpointRangesToKeys: Record<BreakpointKeys, BreakpointKeys> = {
+  xs: 'xs',
+  sm: 'sm',
+  md: 'md',
+  lg: 'lg',
+  xl: 'xl',
 }
 
 const PicassoGlobalStylesProvider = (
@@ -29,8 +39,22 @@ const PicassoGlobalStylesProvider = (
 
   const [picassoRootMounted, setPicassoRootMounted] = useState(false)
   const rootRef = useRef<HTMLDivElement | null>(null)
+
+  const screens = useScreens<BreakpointKeys>()
+  const currentBreakpointRange = screens(breakpointRangesToKeys)
+
+  useEffect(() => {
+    if (contextValue.currentBreakpointRange !== currentBreakpointRange) {
+      setContextValue({
+        ...contextValue,
+        currentBreakpointRange,
+      })
+    }
+  }, [currentBreakpointRange])
+
   const [contextValue, setContextValue] = useState<RootContextProps>({
     rootRef,
+    currentBreakpointRange,
     hasTopBar: false,
     setHasTopBar: (hasTopBar: boolean) => {
       setContextValue({
--- a/packages/picasso-provider/src/Picasso/PicassoGlobalStylesProvider.tsx
+++ b/packages/picasso-provider/src/Picasso/PicassoGlobalStylesProvider.tsx
@@ -18,7 +18,7 @@ export interface PicassoGlobalStylesProviderProps extends TextLabelProps {
   disableTransitions?: boolean
 }
 
-const breakpointRangesToKeys: Record<BreakpointKeys, BreakpointKeys> = {
+const breakpointKeyByRange: Record<BreakpointKeys, BreakpointKeys> = {
   xs: 'xs',
   sm: 'sm',
   md: 'md',
@@ -41,7 +41,7 @@ const PicassoGlobalStylesProvider = (
   const rootRef = useRef<HTMLDivElement | null>(null)
 
   const screens = useScreens<BreakpointKeys>()
-  const currentBreakpointRange = screens(breakpointRangesToKeys)
+  const currentBreakpointRange = screens(breakpointKeyByRange)
 
   useEffect(() => {
     if (contextValue.currentBreakpointRange !== currentBreakpointRange) {
--- a/packages/picasso/src/Grid/Grid.tsx
+++ b/packages/picasso/src/Grid/Grid.tsx
@@ -1,5 +1,5 @@
 import type { ReactNode, HTMLAttributes } from 'react'
-import React, { useMemo, forwardRef } from 'react'
+import React, { forwardRef } from 'react'
 import type { Theme } from '@material-ui/core/styles'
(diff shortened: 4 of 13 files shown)


packages/picasso/src/Grid/Grid.tsx
```diff
@@ -55,31 +71,46 @@ export const Grid = forwardRef<HTMLDivElement, Props>(function Grid(
...
-      alignItems={alignItems}
-      justifyContent={justifyContent}
-      wrap={wrap}
-      classes={classes}
-      className={className}
-      style={style}
-    >
-      {children}
-    </MUIGrid>
+  return useMemo(
```
sashuk, line 77: Need to memoize component to avoid rerenders on each screen change (only when the breakpoint range changes)

packages/picasso/src/Grid/Grid.tsx
```diff
@@ -55,31 +71,46 @@ export const Grid = forwardRef<HTMLDivElement, Props>(function Grid(
...
+      ...Object.values(rest),
+    ]
   )
 })
 
 Grid.defaultProps = {
   alignItems: 'flex-start',
   direction: 'row',
   justifyContent: 'flex-start',
-  spacing: 32,
```
sashuk, line 82: No more default value

packages/picasso-forms/src/Checkbox/__snapshots__/test.tsx.snap
```diff
@@ -29,7 +29,7 @@ exports[`Form.Checkbox default render for checkboxes in a group 1`] = `
             name="checkbox-group"
           >
             <div
-              class="MuiGrid-root PicassoCheckboxGroup-grid MuiGrid-container PicassoGrid-container MuiGrid-direction-xs MuiGrid-align-items-xs-flex"
+              class="MuiGrid-root PicassoCheckboxGroup-grid MuiGrid-container PicassoGrid-container MuiGrid-spacing-xs MuiGrid-direction-xs MuiGrid-align-items-xs-flex"
```
sashuk, line 32: Some snapshots were updated due to addition of `MuiGrid-spacing-xs` class

sashuk:

sashuk:

.changeset/red-countries-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+'@toptal/picasso': major
+---
+
+- grid spacing automatically adjusts to the screen size, unless explicitly specified by consumer. If you want to keep the old behavior, please explicitly set `spacing={32}` for `Grid` components.
```
TomasSlama, line 5:
```suggestion
---

### Grid

- grid spacing automatically adjusts to the screen size, unless explicitly specified by consumer. If you want to keep the old behavior, please explicitly set `spacing={32}` for `Grid` components.
```

packages/picasso/src/Grid/Grid.tsx
```diff
@@ -11,13 +11,17 @@ import type {
...
 import styles from './styles'
 
 export interface Props extends BaseProps, HTMLAttributes<HTMLElement> {
   /** Grid content containing Grid.Item */
   children?: ReactNode
-  /** Defines amount of space between Grid.Item components (in px) */
+  /** Defines amount of space between Grid.Item components (in px). If spacing is not set, then it will
+   * be automatically adjusted based on the screen size (16px for screens smaller than medium, 24px
+   * for medium screens, and 32px for screens bigger than medium)
+   */
```
TomasSlama, line 24: ❤️

packages/picasso/src/Grid/Grid.tsx
```diff
@@ -55,31 +71,47 @@ export const Grid = forwardRef<HTMLDivElement, Props>(function Grid(
...
+      ref,
+      gridSpacing,
+      direction,
+      alignItems,
+      justifyContent,
+      wrap,
+      classes,
+      className,
+      style,
+      children,
```
TomasSlama, line 105: since children is array, does it keep the same reference? I'm not so sure about this. If you have multiple children does this useMemo work?

.changeset/red-countries-thank.md
sashuk: @TomasSlama Fixed, thanks!

packages/picasso/src/Grid/Grid.tsx
sashuk: @TomasSlama Can you please elaborate? Do you mean the case when the array of children changes (for example, an element is added to the children array), will it be caught by the `useMemo()` and the redraw will be triggered?

I've run some experiments before and it seems to work fine – the example below contains two children elements, and when we add an additional one, the grid is redrawn (the `console.log('@@@ Drawing Grid')` is placed in the very heart of the MUI component code to see when it is really redrawn)

https://user-images.githubusercontent.com/1390758/236285954-0ff145d9-06de-4afd-84ec-01a906b9f35e.mp4

packages/picasso/src/Grid/story/index.jsx
```diff
@@ -28,5 +28,10 @@ page
   .addExample('Grid/story/Alignment.example.tsx', 'Alignment')
   .addExample('Grid/story/Direction.example.tsx', 'Direction')
   .addExample('Grid/story/Wrapping.example.tsx', 'Wrapping')
+  .addExample('Grid/story/ResponsiveSpacing.example.tsx', {
+    title: 'Responsive spacing',
+    description:
```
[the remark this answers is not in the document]
sashuk: @TomasSlama Not sure, as in the case of different breakpoints ranges we will be in fact testing the way `useScreens()` hook and MUI Grid work which is already tested (although testing it using `cy.viewport` should be doable, it seems like the responsiveness have never been tested in Cypres in Picasso)

packages/picasso/src/Grid/Grid.tsx
TomasSlama: my concern was if React can keep the same reference for children, so if something changes in the parent and Grid would not have to rerender, if memo would work here, but from your example, it seems it works 🤯. I thought it is not possible

packages/picasso/src/Grid/story/index.jsx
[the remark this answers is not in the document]
TomasSlama: you might be right, maybe it is overengineering :)

TomasSlama (approved):

packages/picasso/src/Grid/Grid.tsx
```diff
@@ -55,31 +71,47 @@ export const Grid = forwardRef<HTMLDivElement, Props>(function Grid(
...
+      ...Object.values(rest),
+    ]
```
OleksandrNechai, line 107: I think `useScreens` is a mistake that causes this awkwardness. If only we refactor useScreens to return screen 'xs' | 'sm'... ) and re-render only when screen changes, we would not need this. I know it is not related to this PR, therefore I will not reject, but would be nice to fix and instead of having `useMemo` here.

OleksandrNechai (approved):
Approving, but I think `useScreens` is a mistake and we should not continue using it and polluting the code.

packages/picasso/src/Grid/Grid.tsx
sashuk: @TomasSlama I'm a bit skeptical too, so reworked it to store the current breakpoint range value in the Picasso context – this way we can avoid using `useMemo` and can fetch the value via context hook that is fired only when the value truly changes. Can you please review it again?

packages/picasso/src/Grid/Grid.tsx
sashuk: @OleksandrNechai I agree, can you please check the https://github.com/toptal/picasso/pull/3538#discussion_r1185897488 comment? The idea is that we can store the current breakpoint range in Picasso context, so whoever wants to subscribe to its changes – the `useCurrentBreakpointRange()` can be used (like in https://github.com/toptal/picasso/commit/95d31674e3dc3db58d2b71811492ebdf2638f737)

sashuk: Regarding the `useScreens()` function itself – indeed, having something that just returns the current breakpoint range whenever it changes (not when the screen is resized every time) will be handy, as an option it could be stored in the context as implemented above

packages/picasso/src/Grid/Grid.tsx
TomasSlama: @sashuk nice, this looks like better solution 🙇

TomasSlama (approved):
I really enjoy your pull request description ❤️, always joy to review

sashuk:

packages/picasso-provider/src/Picasso/PicassoGlobalStylesProvider.tsx
```diff
@@ -16,6 +18,14 @@ export interface PicassoGlobalStylesProviderProps extends TextLabelProps {
   disableTransitions?: boolean
 }
 
+const breakpointRangesToKeys: Record<BreakpointKeys, BreakpointKeys> = {
```
OleksandrNechai, line 21:
```suggestion
const breakpointKeyByRange: Record<BreakpointKeys, BreakpointKeys> = {
```

OleksandrNechai (approved):
This approach looks better to me. I wonder, should not we control these kinds of things with CSS? I guess media queries would be more performant?

sashuk:
@OleksandrNechai This was one of the options, but according to the internal MUI logic when it comes to generating gutters MUI does some calculations like

```
function generateGutter(theme, breakpoint) {
  var styles = {};
  SPACINGS.forEach(function (spacing) {
    var themeSpacing = theme.spacing(spacing);

    if (themeSpacing === 0) {
      return;
    }

    styles["spacing-".concat(breakpoint, "-").concat(spacing)] = {
      margin: "-".concat(getOffset(themeSpacing, 2)),
      width: "calc(100% + ".concat(getOffset(themeSpacing), ")"),
      '& > $item': {
        padding: getOffset(themeSpacing, 2)
      }
    };
  });
  return styles;
}
```

and then applies it to styles, so in this case, I think it would be better to use MUI Grid's public API to set `spacing` so MUI does the rest of the calculation under the hood.

packages/picasso-provider/src/Picasso/PicassoGlobalStylesProvider.tsx
sashuk: @OleksandrNechai Changed, thanks!

PR #3550, opened 2023-05-09
feat: skip ticks for x and y axes

sashuk:
[FX-3957](https://toptal-core.atlassian.net/browse/FX-3957)

### Description

This pull request adds the `showEveryNthTickOnXAxis` and `showEveryNthTickOnYAxis` properties to `BarChart`, so ticks can be skipped on both axes.

### How to test

- Navigate to https://picasso.toptal.net/fx-3957-provide-an-option-to-show-or-hide-x-axis-ticks-in-picasso-barchart/?path=/story/picasso-charts-barchart--barchart and check out the `Show every Nth tick on X or Y-axis` section

### Screenshots

Screenshot 2023-05-10 at 18 31 12

### Development checks

- [x] Self reviewed
- [N/A] Covered with tests

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-3957]: https://toptal-core.atlassian.net/browse/FX-3957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso-charts/src/BarChart/BarChart.tsx
+++ b/packages/picasso-charts/src/BarChart/BarChart.tsx
@@ -37,6 +37,8 @@ export type BarChartDataItem<K extends string | number | symbol> = {
   }
 }
 
+type ShowEverytNthTickValue = 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10
+
 export interface Props<K extends string | number | symbol>
   extends BaseChartProps {
   /**
@@ -64,6 +66,10 @@ export interface Props<K extends string | number | symbol>
   showBarLabel?: boolean
   /** If set false, animation of bar will be disabled */
   isAnimationActive?: boolean
+  /** Makes X-axis show only every Nth tick. `0` hides all ticks, `1` shows all ticks (default behavior), `2` shows every 2nd tick, and so on */
+  showEveryNthTickOnXAxis?: ShowEverytNthTickValue
+  /** Makes Y-axis show only every Nth tick. `0` hides all ticks, `1` shows all ticks (default behavior), `2` shows every 2nd tick, and so on */
+  showEveryNthTickOnYAxis?: ShowEverytNthTickValue
 }
 
 const StyleOverrides = () => (
@@ -106,6 +112,8 @@ const BarChart = <K extends string>({
   showBarLabel,
   isAnimationActive,
   layout,
+  showEveryNthTickOnXAxis = 1,
+  showEveryNthTickOnYAxis = 1,
   ...rest
 }: Props<K>) => {
   const horizontal = layout === 'horizontal'
@@ -171,6 +179,7 @@ const BarChart = <K extends string>({
             axisLine={AXIS_LINE}
             minTickGap={MIN_TICK_GAP}
             tickMargin={TICK_MARGIN}
+            interval={showEveryNthTickOnXAxis - 1}
           />
           <YAxis
             {...yAxisProps}
@@ -179,6 +188,7 @@ const BarChart = <K extends string>({
             axisLine={AXIS_LINE}
             minTickGap={MIN_TICK_GAP}
             tickMargin={TICK_MARGIN}
+            interval={showEveryNthTickOnYAxis - 1}
           />
           {tooltipElement}
           {dataKeys.map(dataKey => (
--- a/.changeset/moody-cooks-walk.md
+++ b/.changeset/moody-cooks-walk.md
@@ -1,0 +1,7 @@
+---
+'@toptal/picasso-charts': minor
+---
+
+### BarChart
+
+- add `showEveryNthTickOnXAxis` and `showEveryNthTickOnYAxis` properties to `BarChart`, so X and Y axes ticks can be skipped
--- a/packages/picasso-charts/src/BarChart/story/ShowEveryNthTick.example.tsx
+++ b/packages/picasso-charts/src/BarChart/story/ShowEveryNthTick.example.tsx
@@ -1,0 +1,55 @@
+import React from 'react'
+import { BarChart } from '@toptal/picasso-charts'
+import { palette } from '@toptal/picasso/utils'
+
+const CHART_DATA = [
+  {
+    name: 'Jan',
+    value: { recovered: 2400 },
+  },
+  {
+    name: 'Feb',
+    value: { recovered: 1398 },
+  },
+  {
+    name: 'Mar',
+    value: { recovered: 9800 },
+  },
+  {
+    name: 'Apr',
+    value: { recovered: 3908 },
+  },
+  {
+    name: 'May',
+    value: { recovered: 3900 },
+  },
+  {
+    name: 'Jun',
+    value: { recovered: 4200 },
+  },
+  {
+    name: 'Jul',
+    value: { recovered: 4500 },
+  },
+]
+
+const COLORS_MAPPING: Record<string, string> = {
+  infected: palette.blue.main,
+  recovered: palette.green.dark,
+}
+
+const Example = () => (
+  <div style={{ width: 720 }}>
+    <BarChart
+      data={CHART_DATA}
+      width='100%'
+      height={300}
+      getBarColor={({ dataKey }) => COLORS_MAPPING[dataKey]}
+      getBarLabelColor={({ dataKey }) => COLORS_MAPPING[dataKey]}
+      showEveryNthTickOnXAxis={3}
+      showEveryNthTickOnYAxis={0}
+    />
+  </div>
+)
+
+export default Example
--- a/packages/picasso-charts/src/BarChart/story/index.jsx
+++ b/packages/picasso-charts/src/BarChart/story/index.jsx
@@ -43,3 +43,8 @@ page
       'You can hide label of each bar via `showBarLabel` prop being set to `false`.',
     takeScreenshot: false,
   })
+  .addExample('BarChart/story/ShowEveryNthTick.example.tsx', {
+    title: 'Show every Nth tick on X or Y-axis',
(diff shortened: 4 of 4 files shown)


OleksandrNechai (commented):
Looks good, though I have API questions ☝️

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -64,6 +64,8 @@ export interface Props<K extends string | number | symbol>
   showBarLabel?: boolean
   /** If set false, animation of bar will be disabled */
   isAnimationActive?: boolean
+  /** Makes X-axis show only every Nth tick (all ticks are shown by default) */
+  showEveryNthTick?: '1' | '2' | '3' | '4'
```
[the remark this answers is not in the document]
sashuk: @OleksandrNechai Thank you for the comment! What do you think about the following API proposal:

```
const BarChart = ({
  ...,
  ticksConfiguration?: {
    xAxis?: {
      showTicks: boolean,
      showEveryNthTick: number
    },
    yAxis?: {
      showTicks?: boolean,
      showEveryNthTick?: number
    }
  }
}) => { ... }

```

This way the tick display configuration will affect both axes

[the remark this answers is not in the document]
OleksandrNechai: 1. I think it is important to check if this approach is consistent with other components in Picasso
2. It seems we could get rid of 2 props and have just one. showEveryNthTick = 0 could mean no ticks.

sashuk:
@toptal-anvil ping reviewers

sashuk:
@toptal-anvil ping reviewers

dmaklygin (approved):
Checked on temploy. Looks good

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -64,6 +64,10 @@ export interface Props<K extends string | number | symbol>
   showBarLabel?: boolean
   /** If set false, animation of bar will be disabled */
   isAnimationActive?: boolean
+  /** Makes X-axis show only every Nth tick. `0` hides all ticks, `1` shows all ticks (default behavior) */
+  showEveryNthTickOnXAxis?: 0 | 1 | 2 | 3
```
augustobmoura, line 68: Why are we limiting it to 0-3? Is it a design constraint?

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -64,6 +64,10 @@ export interface Props<K extends string | number | symbol>
...
+  /** Makes Y-axis show only every Nth tick. `0` hides all ticks, `1` shows all ticks (default behavior) */
+  showEveryNthTickOnYAxis?: 0 | 1 | 2 | 3
```
augustobmoura, line 70: Wdyt about calling this prop the same name as recharts?

either having a single prop:
```suggestion
  interval?: {
    x: 0 | 1 | 2 | 3,
    y: 0 | 1 | 2 | 3,
  }
```

or having a separate prop for x and y:
```suggestion
  xInterval?: 0 | 1 | 2 | 3
  /** Makes Y-axis show only every Nth tick. `0` hides all ticks, `1` shows all ticks (default behavior) */
  yInterval?: 0 | 1 | 2 | 3
```

augustobmoura (approved):
I added a few small comments, but I think the current implementation is already ready to be merged

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -171,6 +177,7 @@ const BarChart = <K extends string>({
             axisLine={AXIS_LINE}
             minTickGap={MIN_TICK_GAP}
             tickMargin={TICK_MARGIN}
+            interval={showEveryNthTickOnXAxis - 1}
```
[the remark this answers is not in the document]
augustobmoura: We are allowing the number `0` for `showEveryNthTickOnXAxis`, in that case this `interval` prop will get the number `-1`, is that correct? What would `-1` mean?

packages/picasso-charts/src/BarChart/BarChart.tsx
sashuk: @augustobmoura Thank you for the suggestion! Yep, we were discussing having such values, but let's make them truly flexible – please see the last commit, it changes the type of these props to `number`, and adds tests.

packages/picasso-charts/src/BarChart/BarChart.tsx
sashuk: @augustobmoura The `interval` property seems a bit unclear, I think, so it is better to keep it verbose enough so users can see what it does right from the name itself

packages/picasso-charts/src/BarChart/BarChart.tsx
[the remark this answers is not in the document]
sashuk: @augustobmoura Correct, in the case of `-1` it does not show ticks at all which is expected

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -106,8 +113,19 @@ const BarChart = <K extends string>({
...
 }: Props<K>) => {
+  if (
+    !isNonNegativeInteger(showEveryNthTickOnXAxis) ||
+    !isNonNegativeInteger(showEveryNthTickOnYAxis)
+  ) {
+    throw new Error(
+      'Non-negative integer values are expected for "showEveryNthTick*" properties'
+    )
+  }
+
```
OleksandrNechai, line 128: Hmm, don't we except `0 | 1 | 2 | 3` instead of `number`? Then this check would not be needed.
Even if it is not possible, I don't think we should react with error and app crash. If something unexpected passed I believe we can simply not show ticks, or use default value. This would serve users and will not flood us with ERs :)

packages/picasso-charts/src/BarChart/test.tsx
```diff
@@ -1,12 +1,39 @@
...
+              [tickConfigurationProperty]: -1,
+            })
+
+          expect(test).toThrow(
+            'Non-negative integer values are expected for "showEveryNthTick*" properties'
+          )
+        })
+      }
+    )
+  })
```
OleksandrNechai, line 35: these tests are very expensive (it loads whole react and DOM). I would give a double thought if we need it.

sashuk: @OleksandrNechai We need to load React to render the component to see if the exception is thrown

packages/picasso-charts/src/BarChart/BarChart.tsx
sashuk: @OleksandrNechai This is done to give more options to end-users after https://github.com/toptal/picasso/pull/3550#discussion_r1191184471 proposal, so it can no longer be easily typed to specific integers in TypeScript

sashuk: > If something unexpected passed I believe we can simply not show ticks, or use the default value.

I think if the error is not thrown, then it needs to be handled gracefully but the warning still should be shown which complicates the behavior (ERs will be assigned to those who misused the components)

sashuk:
@toptal-anvil ping reviewers

packages/picasso-charts/src/BarChart/test.tsx
sashuk: @OleksandrNechai I think in this case it is needed, as we need to test the calling of the function that checks if the input parameters are correct, plus we need to make sure that it is called in the proper place (in the component, so it needs to be loaded). What do you think?

packages/picasso-charts/src/BarChart/BarChart.tsx
OleksandrNechai: >This is done to give more options to end-users after https://github.com/toptal/picasso/pull/3550#discussion_r1191184471 proposal

I see. Picasso's philosophy has always been `less is more. We are limiting APIs on purpose to increase consistency in our apps. Also, it is much easier to change 0 | 1 | 2 | 3 --> number than to change it back. 

>I think if the error is not thrown, then it needs to be handled gracefully 

Yes, there is an idea to fail early. It is easier to debug when we do not swallow the error in places where it happens. But we are doing a highly reusable library. To me, Postel's law feels more appropriate https://devopedia.org/postel-s-law 
>Be liberal in what you accept, and conservative in what you send.

crashing seems too much in this situation

sashuk: > We are limiting APIs on purpose to increase consistency in our apps.

Can you please specify why using enums is considered to be more consistent in this case? Picasso components API's do not really force it and I can see that `someProp?: number` is very common in Picasso APIs

> crashing seems too much in this situation

Thanks for the link! However, I do think that in this case it is still suitable, as there is a high chance that either exception will be thrown when testing (if an engineer sets the wrong property value) or it will be caught on review. This is basic data validation of the input, so silently substituting user's input will make things more obscure from consumer point of view which is not a good approach.

OleksandrNechai (approved):
We discussed my concerns, the author insists on the current solution and I am not strongly disagreeing. Hence, approve from me 😄

packages/picasso-charts/src/BarChart/BarChart.tsx
sashuk: Clarifying the Nth ticks display question in https://toptal-core.slack.com/archives/GG3F4AS4T/p1683882228257659

sashuk: Decision https://toptal-core.slack.com/archives/GG3F4AS4T/p1683884552949139

PR #3646, opened 2023-06-22
feat: integrate default value

sashuk:
[FX-4014]

### Description

This pull request restores the ability to provide initial value to the Rich Text Editor in the form of AST as it used to work before.

### How to test

- navigate to https://picasso.toptal.net/fx-4014-integrate-default-value-2/?path=/story/forms-richtexteditor--richtexteditor and https://picasso.toptal.net/fx-4014-integrate-default-value-2/?path=/story/picasso-forms-form--form and play around with Rich Text Editor with default values
- try copy-pasting rich text content from https://docs.google.com/document/d/1uiVxvvkfGfCb7SaW_4tfKLl4sxpSwyaPra-4SPQR6iA/edit?usp=sharing

### Screenshots

Screenshot 2023-06-23 at 13 37 20

### Development checks

- [N/A] Add changeset according to [guidelines](https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md) (if needed)
- [N/A] Make sure that additions and changes on design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core
- [x] Self reviewed
- [x] Covered with tests

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4014]: https://toptal-core.atlassian.net/browse/FX-4014?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -7,6 +7,7 @@ import { useForm } from 'react-final-form'
 import type { FieldProps } from '../FieldWrapper'
 import InputField from '../InputField'
 import FieldLabel from '../FieldLabel'
+import { useEnforceHighlightAutofill } from './hooks'
```
sashuk, line 10: Hook that enforces the highlighting in Rich Text Editor fields after the highlighting is gone after the first `onChange` firing

packages/picasso-forms/src/RichTextEditor/RichTextEditor.tsx
```diff
@@ -31,17 +35,26 @@ export const RichTextEditor = (props: Props) => {
...
     (newVal: string) => {
+      registerChangeOrFocus()
       setValue(newVal)
       onChange?.(newVal)
     },
-    [onChange, setValue]
+    [onChange, setValue, registerChangeOrFocus]
   )
+
+  const handleOnFocus = useCallback(() => {
```
sashuk, line 45: Focus needs to update the highlight state, so we track it too (the highlight should be gone when user focuses into the field)

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
```diff
@@ -171,6 +179,9 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
           // remount Toolbar when disabled
           key={`${disabled || !isFocused}`}
         />
+        {defaultValue ? (
```
sashuk, line 182: Initial `onChange` firing is needed only when default value is provided

packages/picasso/src/LexicalEditor/plugins/TriggerInitialOnChangePlugin.ts
```diff
@@ -0,0 +1,31 @@
...
+    tags: Set<string>
+  ) => void
+}) => {
+  const [editor] = useLexicalComposerContext()
+
+  useIsomorphicLayoutEffect(() => {
+    if (onChange) {
+      return editor.registerUpdateListener(
+        ({ editorState, prevEditorState, tags }) => {
+          if (prevEditorState.isEmpty()) {
```
sashuk, line 20: The `onChange` is fired only if editor is in pristine state (so, after the initial render)

packages/picasso/src/LexicalEditor/utils/setEditorValue.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const setEditorValue = (editor: LexicalEditorType, value: ASTType) => {
+  const domValue = getDomValue(value)
+
+  editor.update(() => {
+    const root = $getRoot()
+    const lexicalValueNodes = $generateNodesFromDOM(editor, domValue)
+
+    lexicalValueNodes.forEach(node => {
+      if ($isElementNode(node) || $isDecoratorNode(node)) {
```
sashuk, line 22: Covering all cases of base Lexical nodes (https://lexical.dev/docs/concepts/nodes#base-nodes)

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
TomasSlama, line 1: I think it might be worth adding some documentation to this file as in the future it will be hard for others to understand what problem we are trying to solve here.

packages/picasso/src/LexicalEditor/plugins/TriggerInitialOnChangePlugin.ts
```diff
@@ -0,0 +1,31 @@
...
+}: {
+  onChange: (
+    editorState: EditorState,
+    editor: LexicalEditor,
+    tags: Set<string>
+  ) => void
+}) => {
+  const [editor] = useLexicalComposerContext()
+
+  useIsomorphicLayoutEffect(() => {
```
TomasSlama, line 16: why do we need to use the Layout effect? 🤔

packages/picasso/src/LexicalEditor/utils/getDomValue.ts
```diff
@@ -0,0 +1,10 @@
+import type { HastNode } from 'hast-util-to-dom/lib'
```
TomasSlama, line 1: can we just use `toDom` from this new [library ](https://github.com/syntax-tree/hast-util-to-dom#use) that we added?

TomasSlama (approved):
Left a few questions. I tested it in the temploy and it works pretty well. Good job! 💪

packages/picasso/src/LexicalEditor/utils/getDomValue.ts
sashuk: @TomasSlama Unfortunately, `$generateNodesFromDOM()` expects the Document

Screenshot 2023-06-27 at 13 16 13

packages/picasso/src/LexicalEditor/plugins/TriggerInitialOnChangePlugin.ts
sashuk: @TomasSlama After checking the original code in https://github.com/facebook/lexical/commit/70e13adbd49f4c638e905f0e7552e140928c8ab6#diff-aaf0860861200a139a415f81f8eebf3ea393bcb99f1bf82832e6e25887d0509dR22 it seems like there are no explanation for having this hook. So it seems like it can be removed, since the change is triggered only once so it should be safe, thanks!

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
sashuk: @TomasSlama Added, thank you!

sashuk:
@toptal-anvil ping reviewers

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
```diff
@@ -0,0 +1,44 @@
...
+    return timesChangeOrFocusTriggered < 2
+  }, [highlightAutofill, timesChangeOrFocusTriggered])
+
+  const registerChangeOrFocus = useCallback(() => {
+    if (!highlightAutofill) {
+      return
+    }
+
+    if (timesChangeOrFocusTriggered < TRACKED_EVENTS_LIMIT) {
+      setTimesChangeOrFocusTriggered(timesChangeOrFocusTriggered + 1)
```
dmaklygin, line 36:
```suggestion
      setTimesChangeOrFocusTriggered(times => times + 1)
```

to rely on the current `times`

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
```diff
@@ -0,0 +1,44 @@
...
+ * This hook keeps the highlighted state for Rich Text Editor form fields after
+ * first `onChange` callback which is triggered by editor itself when initial value is
+ * provided. Starting from second `onChange` or first `onFocus` invocation (which are
+ * triggered by user) the field is no longer highlighted.
+ *
+ * Editor triggers the very first onChange callback by itself only when initial value
+ * is provided in order to propagate the value to wrapping form.
+ */
+
+const TRACKED_EVENTS_LIMIT = 3
```
dmaklygin, line 15: did not get, why it's 3, not a `1`

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
```diff
@@ -130,6 +133,11 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
 
   const editorConfig: InitialConfigType = useMemo(
     () => ({
+      editorState: (editor: LexicalEditorType) => {
+        if (defaultValue) {
+          setEditorValue(editor, defaultValue)
```
dmaklygin, line 138: could you please take a look at: https://github.com/facebook/lexical/blob/2bcf11bec944cd3bd69b7e7b40903e593b716a06/packages/lexical-react/src/LexicalComposer.tsx#L119

maybe we could reuse something from this initial editor state function?

sashuk: @dmaklygin I think we can go step-by-step here, as the https://github.com/facebook/lexical/blob/2bcf11bec944cd3bd69b7e7b40903e593b716a06/packages/lexical-react/src/LexicalComposer.tsx#L119 actually takes input in the form of `EditorState` (https://github.com/facebook/lexical/blob/main/packages/lexical/src/LexicalEditorState.ts#L97) that includes both editor state and initial value (which is usually passed around as JSON).

In this particular case, we take the root element of `defaultValue` which is passed as AST, get its children, and process each one depending on their type

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
sashuk: @dmaklygin Changed it to `2` and added a comment, thanks!

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
sashuk: @dmaklygin Can you please elaborate on why it is needed if we have a `timesChangeOrFocusTriggered` as a dependency of `useCallback()`, so it is always fresh?

packages/picasso-forms/src/RichTextEditor/hooks/use-enforce-highlight-autofill.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const registerChangeOrFocus = useCallback(() => {
+    if (!highlightAutofill) {
+      return
+    }
+
+    if (timesChangeOrFocusTriggered < TRACKED_EVENTS_LIMIT) {
+      setTimesChangeOrFocusTriggered(timesChangeOrFocusTriggered + 1)
+    }
+  }, [timesChangeOrFocusTriggered])
```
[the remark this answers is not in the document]
sashuk: @dmaklygin It does not change throughout the hook lifecycle, but it should be still mentioned in dependencies. Added, thanks!

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
dmaklygin, line 138: shouldn't we add defaultValue as a dependency of `useMemo`?

packages/picasso/src/LexicalEditor/utils/getDomValue.ts
dmaklygin: it actually allows to return a Document: https://github.com/syntax-tree/hast-util-to-dom#fields
according to a documentation

dmaklygin: However, I still think we may reuse it. `DocumentFragment` also fits the requirements

packages/picasso/src/LexicalEditor/utils/setEditorValue.ts
```diff
@@ -0,0 +1,36 @@
...
+} from 'lexical'
+import { $generateNodesFromDOM } from '@lexical/html'
+
+import type { ASTType } from '../../RichText'
+import { getDomValue } from './getDomValue'
+
+export const setEditorValue = (editor: LexicalEditorType, value: ASTType) => {
+  const domValue = getDomValue(value)
+
+  editor.update(() => {
```
dmaklygin, line 17: seems, we don't need to wrap it again, since it's already covered here: https://github.com/facebook/lexical/blob/main/packages/lexical-react/src/LexicalComposer.tsx#L152

packages/picasso/src/LexicalEditor/utils/setEditorValue.ts
```diff
@@ -0,0 +1,36 @@
...
+        root.append(node)
+      } else if ($isLineBreakNode(node)) {
+        const paragraphNode = $createParagraphNode()
+
+        root.append(paragraphNode)
+      } else {
+        const paragraphNode = $createParagraphNode()
+
+        paragraphNode.append(node)
+        root.append(paragraphNode)
```
dmaklygin, line 32: I'm not sure, but whether these rest conditions are necessary? it won't work if we just use `root.append(node)`?

dmaklygin (approved):
Good work. Checked.
left only minor comments

packages/picasso/src/LexicalEditor/utils/getDomValue.ts
sashuk: @dmaklygin @TomasSlama I've tried case with `DocumentFragment`, this option still requires some type casting and manual creation of `Document` so the fragment nodes can be appended to it. Do you think it should still be refactored? We can get rid of one dependency but introduce more complexity in this case

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
sashuk: @dmaklygin Done, thanks!

packages/picasso/src/LexicalEditor/utils/setEditorValue.ts
sashuk: @dmaklygin Thank you for catching it, fixed!

packages/picasso/src/LexicalEditor/utils/setEditorValue.ts
sashuk: @dmaklygin They are necessary, otherwise Lexical throws an error

Screenshot 2023-06-28 at 22 17 01

So we wrap everything which is not `Element` or `Decorator` node

packages/picasso/src/LexicalEditor/utils/getDomValue.ts
TomasSlama: I don't have a strong opinion against it. I would vote for less complexity ;-)

PR #3653, opened 2023-06-26
feat(RichTextEditor): add Emoji plugin

TomasSlama:
[FX-4012]

### Description

- Create context to enable/disable optional features
- Add emoji plugin

### How to test

- Play with emojis [in the temploy](https://picasso.toptal.net/FX-4012-emoji)

### Screenshots

| image | image |

### Development checks

- [x] Self reviewed
- [x] Covered with tests

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4012]: https://toptal-core.atlassian.net/browse/FX-4012?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso/src/LexicalEditor/LexicalEditor.tsx
+++ b/packages/picasso/src/LexicalEditor/LexicalEditor.tsx
@@ -8,12 +8,12 @@ import { RichTextPlugin } from '@lexical/react/LexicalRichTextPlugin'
 import { AutoFocusPlugin } from '@lexical/react/LexicalAutoFocusPlugin'
 import { ContentEditable } from '@lexical/react/LexicalContentEditable'
 import LexicalErrorBoundary from '@lexical/react/LexicalErrorBoundary'
-import { HeadingNode } from '@lexical/rich-text'
 import { $generateHtmlFromNodes } from '@lexical/html'
-import { ListItemNode, ListNode } from '@lexical/list'
 import { $isRootTextContentEmpty } from '@lexical/text'
 import type { LexicalEditor as LexicalEditorType } from 'lexical'
 import { OnChangePlugin } from '@lexical/react/LexicalOnChangePlugin'
+import { ListItemNode, ListNode } from '@lexical/list'
+import { HeadingNode } from '@lexical/rich-text'
 
 import { TriggerInitialOnChangePlugin } from './plugins'
 import { createLexicalTheme, setEditorValue } from './utils'
@@ -22,12 +22,19 @@ import Container from '../Container'
 import Typography from '../Typography'
 import { useTypographyClasses, useOnFocus } from './hooks'
 import styles from './styles'
-import type { ChangeHandler, TextLengthChangeHandler } from './types'
+import type {
+  ChangeHandler,
+  EditorPlugin,
+  TextLengthChangeHandler,
+  CustomEmojiGroup,
+} from './types'
 import ToolbarPlugin from '../LexicalEditorToolbarPlugin'
 import LexicalTextLengthPlugin from '../LexicalTextLengthPlugin'
 import LexicalListPlugin from '../LexicalListPlugin'
 import LexicalHeadingsReplacementPlugin from '../LexicalHeadingsReplacementPlugin'
 import type { ASTType } from '../RichText'
+import { CustomEmojiNode } from '../LexicalEmojiPlugin/nodes/CustomEmojiNode'
+import LexicalEmojiPlugin from '../LexicalEmojiPlugin'
 
 const useStyles = makeStyles<Theme>(styles, {
   name: 'LexicalEditor',
@@ -68,8 +75,7 @@ export type Props = BaseProps & {
   onTextLengthChange: TextLengthChangeHandler
   /** The placeholder attribute specifies a short hint that describes the expected value of a text editor. */
   placeholder?: string
-  /** List of plugins to enable on the editor */
-  //   plugins?: EditorPlugin[]
+
   testIds?: {
     editor?: string
     // headerSelect?: string
@@ -78,7 +84,9 @@ export type Props = BaseProps & {
     // unorderedListButton?: string
     // orderedListButton?: string
   }
-  //   customEmojis?: CustomEmojiGroup[]
+  customEmojis?: CustomEmojiGroup[]
+  /** List of plugins to enable on the editor */
+  plugins?: EditorPlugin[]
 }
 
 const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
@@ -86,7 +94,7 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
   ref
 ) {
   const {
-    // plugins,
+    plugins = [],
     autoFocus = false,
     defaultValue,
     disabled = false,
@@ -108,7 +116,7 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
     // @todo don't know what to do with NAME prop
     // name,
     // highlight,
-    // customEmojis,
+    customEmojis,
   } = props
 
   const classes = useStyles()
@@ -128,7 +136,7 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
         typographyClassNames,
         classes,
       }),
-    [typographyClassNames, classes.paragraph]
+    [typographyClassNames, classes]
   )
 
   const editorConfig: InitialConfigType = useMemo(
@@ -143,7 +151,7 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
         throw error
       },
       namespace: 'editor',
-      nodes: [ListNode, ListItemNode, HeadingNode],
+      nodes: [CustomEmojiNode, ListNode, ListItemNode, HeadingNode],
       editable: !disabled,
     }),
     [defaultValue, theme, disabled]
@@ -178,6 +186,8 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
           toolbarRef={toolbarRef}
           // remount Toolbar when disabled
           key={`${disabled || !isFocused}`}
+          customEmojis={customEmojis}
+          plugins={plugins}
         />
         {defaultValue ? (
           <TriggerInitialOnChangePlugin onChange={handleChange} />
@@ -188,6 +198,7 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
         <LexicalHeadingsReplacementPlugin />
         <LexicalTextLengthPlugin onTextLengthChange={onTextLengthChange} />
         <LexicalListPlugin />
+        <LexicalEmojiPlugin />
 
         <div className={classes.editorContainer} id={id} ref={ref}>
           <RichTextPlugin
--- a/packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
+++ b/packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
@@ -1,0 +1,141 @@
+import React from 'react'
+import { $applyNodeReplacement, DecoratorNode } from 'lexical'
+import type {
+  DOMConversionMap,
+  DOMConversionOutput,
+  DOMExportOutput,
+  EditorConfig,
+  LexicalNode,
+  NodeKey,
+  SerializedElementNode,
+  Spread,
+} from 'lexical'
+
+export interface CustomEmojiPayload {
+  src: string
+  id: string
+}
+
+type SerializedCustomEmojiNode = Spread<
+  {
+    src: string
+    id: string
+  },
+  SerializedElementNode
+>
+
+const convertImageElement = (domNode: Node): null | DOMConversionOutput => {
+  if (domNode instanceof HTMLImageElement) {
+    const src = domNode.getAttribute('src')
+    const id = domNode.getAttribute('data-emoji-name')
+
+    if (src && id) {
+      return {
+        node: $createCustomEmojiNode({
+          src,
+          id,
+        }),
+      }
+    }
+
+    return null
+  }
+
+  return null
+}
+
+export class CustomEmojiNode extends DecoratorNode<JSX.Element> {
+  src: string
+  id: string
+
+  static getType() {
+    return 'custom-emoji'
+  }
+
+  static clone(node: CustomEmojiNode): CustomEmojiNode {
+    return new CustomEmojiNode(node.src, node.id)
+  }
+
+  constructor(src: string, id: string, key?: NodeKey) {
+    super(key)
+    this.src = src
+    this.id = id
+  }
+
+  static importJSON(
+    serializedNode: SerializedCustomEmojiNode
+  ): CustomEmojiNode {
+    const { src, id } = serializedNode
+
+    const node = $createCustomEmojiNode({ src, id })
+
+    return node
+  }
+
+  createDOM(config: EditorConfig): HTMLElement {
+    const span = document.createElement('span')
+
+    const theme = config.theme
+    const className = theme.customEmoji
+
+    if (className !== undefined) {
+      span.className = className
+    }
+
+    return span
+  }
+
+  updateDOM() {
+    return false
+  }
+
+  exportDOM(): DOMExportOutput {
+    const element = document.createElement('img')
+
+    element.setAttribute('src', this.src)
+    element.setAttribute('data-src', this.src)
+    element.setAttribute('data-emoji-name', this.id)
+
+    return { element }
+  }
+
+  static importDOM(): DOMConversionMap | null {
+    return {
+      img: () => ({
+        conversion: convertImageElement,
+        priority: 0,
+      }),
+    }
+  }
+
+  exportJSON(): SerializedCustomEmojiNode {
+    return {
+      version: 1,
+      type: 'custom-emoji',
+      src: this.src,
+      id: this.id,
+      children: [],
+      direction: this.__direction,
+      format: '',
+      indent: this.__indent,
+    }
+  }
+
+  decorate() {
+    return <img src={this.src} data-src={this.src} data-emoji-name={this.id} />
+  }
+}
+
+export const $isCustomEmojiNode = (
+  node: LexicalNode | null | undefined
+): node is CustomEmojiNode => {
+  return node instanceof CustomEmojiNode
+}
+
+export const $createCustomEmojiNode = ({
+  src,
+  id,
+}: CustomEmojiPayload): CustomEmojiNode => {
+  // return new CustomEmojiNode(src, id)
+  return $applyNodeReplacement(new CustomEmojiNode(src, id))
+}
--- a/packages/picasso/src/LexicalEditor/LexicalEditor.test.tsx
+++ b/packages/picasso/src/LexicalEditor/LexicalEditor.test.tsx
@@ -9,6 +9,7 @@ import LexicalTextLengthPlugin from '../LexicalTextLengthPlugin'
 import LexicalHeadingsReplacementPlugin from '../LexicalHeadingsReplacementPlugin'
 import ToolbarPlugin from '../LexicalEditorToolbarPlugin'
 import LexicalListPlugin from '../LexicalListPlugin'
+import type { CustomEmojiGroup } from './types'
 
 jest.mock('../LexicalEditorToolbarPlugin', () => ({
   __esModule: true,
@@ -17,6 +18,10 @@ jest.mock('../LexicalEditorToolbarPlugin', () => ({
 jest.mock('../LexicalListPlugin', () => ({
   __esModule: true,
   default: jest.fn(() => <div>LexicalListPlugin</div>),
+}))
+jest.mock('../LexicalEmojiPlugin', () => ({
+  __esModule: true,
+  default: jest.fn(() => <div>LexicalEmojiPlugin</div>),
 }))
 
 jest.mock('@lexical/react/LexicalComposerContext', () => ({
@@ -136,6 +141,8 @@ describe('LexicalEditor', () => {
       expect(mockedToolbarPlugin).toHaveBeenCalledWith(
         {
           disabled: true,
+          customEmojis: undefined,
+          plugins: [],
           toolbarRef: {
             current: null,
           },
@@ -152,6 +159,30 @@ describe('LexicalEditor', () => {
       expect(mockedToolbarPlugin).toHaveBeenCalledWith(
         {
           disabled: true,
+          customEmojis: undefined,
+          plugins: [],
+          toolbarRef: {
+            current: null,
+          },
+        },
+        {}
+      )
+    })
+  })
+
+  describe('when customEmojis and plugins prop is passed', () => {
+    it('renders ToolbarPlugin with correct props', () => {
+      renderLexicalEditor({
+        disabled: true,
+        customEmojis: ['foo' as unknown as CustomEmojiGroup],
+        plugins: ['link'],
+      })
+
+      expect(mockedToolbarPlugin).toHaveBeenCalledWith(
+        {
+          disabled: true,
+          customEmojis: ['foo'],
+          plugins: ['link'],
           toolbarRef: {
             current: null,
           },
--- a/packages/picasso/src/LexicalEditor/styles.ts
+++ b/packages/picasso/src/LexicalEditor/styles.ts
@@ -153,5 +153,12 @@ export default ({ typography }: Theme) => {
     italic: {
       fontStyle: 'italic',
     },
+    customEmoji: {
+      '& > img': {
+        verticalAlign: 'bottom',
+        width: '22px',
+        height: '22px',
+      },
+    },
   })
 }
--- a/packages/picasso/src/LexicalEditor/types.ts
+++ b/packages/picasso/src/LexicalEditor/types.ts
@@ -1,3 +1,35 @@
+export type SettingName = 'link' | 'emoji'
+
 export type ChangeHandler = (html: string) => void
 
 export type { TextLengthChangeHandler } from '../LexicalTextLengthPlugin'
+
+export type EditorPlugin = SettingName
+
+export type CustomEmoji = {
+  id: string
+  name: string
+  keywords: string[]
+  skins: [
+    {
+      src: string
+    }
+  ]
+}
+
+export type CustomEmojiGroup = {
+  id: string
+  name: string
+  emojis: CustomEmoji[]
+}
+
+export type Emoji = {
+  id: string
+  name: string
+  native?: string
+  unified?: string
+  keywords: string[]
+  shortcodes: string
+  emoticons?: string[]
+  src?: string
+}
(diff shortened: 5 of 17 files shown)


TomasSlama:
@toptal-anvil ping reviewers

packages/picasso/src/LexicalEditor/context/SettingsContext.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const plugins = JSON.parse(getByTestId('plugins').textContent as string)
+
+    waitFor(() => {
+      expect(nodes).toEqual([ListNode, ListItemNode])
+      expect(settings).toEqual({ link: false, emoji: false, image: false })
+      expect(plugins).toEqual([])
+    })
+  })
+
+  it('enables plugins when string plugins are provided', async () => {
```
sashuk, line 37: Maybe to put `when string plugins are provided` in separate `describe`?

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+    if (src && id) {
+      return {
+        node: $createCustomEmojiNode({
+          src,
+          id,
+        }),
+      }
+    }
+
+    return null
```
sashuk, line 40: Is `return null` needed here?

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+  static importJSON(
+    serializedNode: SerializedCustomEmojiNode
+  ): CustomEmojiNode {
+    const { src, id } = serializedNode
+
+    return $createCustomEmojiNode({ src, id })
+  }
+
+  createDOM(config: EditorConfig): HTMLElement {
+    const span = document.createElement('span')
```
sashuk, line 73: Not sure if it is related to this code, but it seems like an extra span is created for custom emoji

Screenshot 2023-06-27 at 13 52 34

packages/picasso/src/LexicalEditor/utils/synchronizeToolbarState.test.ts
```diff
@@ -14,6 +14,11 @@ import { ToolbarActions } from './toolbarState'
 import { synchronizeToolbarState } from './synchronizeToolbarState'
 import { getLexicalNode } from './getLexicalNode'
 
+jest.mock('../nodes/CustomEmojiNode', () => ({
+  // $createCustomEmojiNode: jest.fn(),
```
sashuk, line 18: NIT: should it be removed?

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
TomasSlama: yeah, I know. There is an issue, where after inserting of custom emoji, the cursor from the Editor disappeared. I fixed it by adding and inserting space right after.

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
TomasSlama: yeah, I think so.
image
the conversion function expects null | DomConversionOutput

And it is also done like this in [the example](https://github.com/facebook/lexical/blob/main/packages/lexical-playground/src/nodes/ImageNode.tsx#L43)

packages/picasso/src/LexicalEditor/context/SettingsContext.test.tsx
TomasSlama: ✅

packages/picasso/src/LexicalEditor/utils/synchronizeToolbarState.test.ts
TomasSlama: ✅

TomasSlama:
@toptal-anvil ping reviewers

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
```diff
@@ -125,17 +132,17 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
         typographyClassNames,
         classes,
       }),
-    [typographyClassNames, classes.paragraph]
+    [typographyClassNames, classes]
   )
 
-  const editorConfig: InitialConfigType = useMemo(
-    () => ({
+  const editorConfig: (nodes: NodesType) => InitialConfigType = useCallback(
```
dmaklygin, line 138:
```suggestion
  const editorConfig = useCallback<(nodes: NodesType) => InitialConfigType>(
```

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
```diff
@@ -125,17 +132,17 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
...
+    (nodes: NodesType) => ({
```
dmaklygin, line 139: then, we can avoid using type here

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+      }
+    })
+  }, [plugins, setOption])
+
+  return (
+    <Context.Provider value={contextValue}>
+      {children(contextValue)}
+    </Context.Provider>
+  )
+}
```
dmaklygin, line 110: I'm thinking about the reason of having this context

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
```diff
@@ -163,45 +170,59 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
...
-              </Container>
-            }
-            ErrorBoundary={LexicalErrorBoundary}
-          />
-        </div>
-      </div>
-    </LexicalComposer>
+              <LexicalListPlugin />
+
+              {pluginsToRender.map((plugin: ReactNode) => plugin)}
```
augustobmoura, line 197: Why do we need this `.map` call? Do you mean to clone the array?
Maybe just
```suggestion
              {pluginsToRender}
```
wdyt?

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+    return null
+  }
+
+  return null
+}
+
+export class CustomEmojiNode extends TextNode {
+  __src: string
+  __id: string
```
augustobmoura, line 48: I know that this was copied from the playground example but on our code conventions we don't use double underscores as naming features, wdyt just keeping these fields as `src` and `id`?

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  static clone(node: CustomEmojiNode): CustomEmojiNode {
+    return new CustomEmojiNode(node.__src, node.__id)
+  }
+
+  constructor(src: string, id: string, key?: NodeKey) {
+    super(' ', key)
+    this.__src = src
+    this.__id = id
+  }
```
augustobmoura, line 62: There is a trick with typescript to auto-set fields from the constructor if you defined the visibility them so:
```ts
class CustomEmojiNode {
  src: string
  id: string
  
  constructor(src: string, id: string) {
    this.src = src
    this.id = id
  }
}
```

Can be written as:
```ts
class CustomEmojiNode {
  constructor(public src: string, public id: string) {
  }
}
```

Wdyt?
```suggestion
  constructor(public src: string, public id: string, key?: NodeKey) {
    super(' ', key)
  }
```

augustobmoura (commented):
I left a few comments, but didn't tested them yet

packages/picasso/src/LexicalEditor/utils/registerLexicalEvents.ts
```diff
@@ -18,14 +30,41 @@ export const registerLexicalEvents = ({
...
+      },
+      COMMAND_PRIORITY_LOW
+    ),
+    editor.registerCommand(
+      INSERT_CUSTOM_EMOJI_COMMAND,
+      (customEmojiPayload: CustomEmojiPayload) => {
+        editor.update(() => {
+          $insertNodes([$createCustomEmojiNode(customEmojiPayload)])
+
+          editor.dispatchCommand(CONTROLLED_TEXT_INSERTION_COMMAND, ' ')
```
dmaklygin, line 61: let's maybe think why do we need to dispatch extra command here?

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
TomasSlama: you are right, we probably don't need it

packages/picasso/src/LexicalEditor/utils/registerLexicalEvents.ts
dmaklygin: `insertNodes` supports a second argument. may it affect?

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
TomasSlama: I don't have a strong opinion against it, I was just thinking to keep the same conventions as the class that we extend

augustobmoura: I would prefer to keep it consistent with the code on Picasso itself rather than consistent with lexical's implementation. Otherwise, we would have multiple styles on our codebase

packages/picasso/src/LexicalEditor/utils/registerLexicalEvents.ts
TomasSlama: This one is tricky, I spend multiple hours trying to solve it without this dispatch, but the thing is, that if we add custom emoji as a first character in the editor, we are losing focus of the editor. I couldn't come up with a proper solution, so I found this "hack" to add white space after the icon and it returns the focus after the space.

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+export const SettingsContext = ({
+  children,
+  plugins,
+}: ProviderType): JSX.Element => {
+  const [settings, setSettings] = useState(DEFAULT_SETTINGS)
+  const [nodes, setNodes] = useState(DEFAULT_NODES)
+  const [pluginsToRender, setPluginsToRender] = useState<ReactNode[]>([])
+
+  const setOption = useCallback((name: SettingName, value: boolean) => {
```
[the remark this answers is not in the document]
TomasSlama: yeah, it just sets the option to true or false. If false, it also removes the nodes.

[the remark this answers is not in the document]
augustobmoura: Maybe we can think of a better name, wdyt? I saw that `SettingName` are the name of the plugins, so maybe `setPluginActive`, or maybe I missed how could this method could be extended in the future

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: @dmaklygin, @augustobmoura This context will be needed for ReactNode custom plugins.

Inside ImagePlugin, we will need to call `setOption('image', true)` in use effect. It will be propagated to the context and proper Lexical nodes will be registered and the editor will need to be remounted.

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+        return !pluginNodes.some(pluginNode => {
+          return node === pluginNode
+        })
+      })
+    })
+  }, [])
+
+  const contextValue = useMemo(
+    () => ({ setOption, settings, nodes, pluginsToRender }),
+    [setOption, settings, nodes, pluginsToRender]
```
dmaklygin, line 92: seems, `setOption` is not used outside of this context. should we keep it?

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  useEffect(() => {
+    plugins.forEach(plugin => {
+      if (typeof plugin === 'string') {
+        setOption(plugin as SettingName, true)
+      } else {
+        setPluginsToRender(prevPlugins => [...prevPlugins, plugin])
+      }
+    })
+  }, [plugins, setOption])
```
dmaklygin, line 103: I'm afraid, this useEffect will be called each time `plugins` array (instance of the array) is changed, even if the items are the same.

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: It will be used in plugins that will be ReactNode

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: my bad, you are right, I will fix this tomorrow. I will use `JSON.stringify(plugins)`

packages/picasso/src/LexicalEditor/nodes/CustomEmojiNode.tsx
TomasSlama: I will apply fixes tomorrow

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
dmaklygin: why do we need a context for `pluginsToRender` ?

TomasSlama: as plugins will be mix of strings and ReactNodes, we need to split them and ReactNodes we want to just render, strings we need to initialize.

TomasSlama: as in this context we do logic around plugins, I thought it is the proper place to put it.

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
augustobmoura: I was thinking about the implementation, and we might avoid using a context when implementing ReactNode plugins, we can just render them as children directly, without setting on the context.

So instead of something like:
```ts
<SettingsContext plugins={plugins}>
	 {(pluginsToRender) => {
	   return (
	     <>
	       // ...
	       {pluginsToRender}
	     </>
	   )
	 }}
</SettingsContext>
```

We could just have
```ts
<SettingsContext>
	 {plugins}
</SettingsContext>
```

We can think on more complex cases on the ticket for the image plugin

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
augustobmoura: You can also set it to `[...plugins, setOption]` instead of `JSON.stringify`

augustobmoura: Another options is to memoize `plugins` in some place, and think of it as immutable, when we need a change we clone it and pass it forward

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
dmaklygin: Could you also explain the purpose of `settings`?

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
```diff
@@ -0,0 +1,112 @@
...
+  setOption: (() => {}) as SetOptionType,
+  settings: DEFAULT_SETTINGS,
+  nodes: DEFAULT_NODES,
+  pluginsToRender: [] as ReactNode[],
+})
+
+const NODES_MAP = {
+  emoji: [CustomEmojiNode],
+  link: null,
+  image: null,
```
dmaklygin, line 53: so, as soon as we implement `ImageUpload`, we should describe `ImageUploadNode` here in the map?

TomasSlama: yes 👍

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: main purpose of `settings` is to show/hide buttons in the Toolbar. But also as it is an object with primitives, I stringify it and use it as key for `<LexicalEditor key={JSON.stringify(settings)} />` as every time it changes it will remount the whole Editor with correct nodes. My thought process was that it will be more efficient to stringify settings than whole plugins

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
dmaklygin: should we support RTE with `mutable` plugins right now?

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: @augustobmoura interesting idea, I never thought about rendering anything else in context than children or children as render prop, but why not.

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
dmaklygin: I'm afraid of high coupling between plugins and core component... The code component knows many things about external plugins

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: yeah, but it would become harder when we pass ReactNode, right? With an array of primitives, it would make more sense to me 🤔

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: I'm not sure I follow. We don't have any external plugins, we only support our plugins, which we would be exporting.

packages/picasso/src/LexicalEditor/context/SettingsContext.tsx
TomasSlama: @dmaklygin I think it will be the most common use case.
```ts
<RichTextEditor plugins={['link']} />
```

TomasSlama: I'm not sure how we could force users to pass immutable array

packages/picasso/src/LexicalEditor/utils/registerLexicalEvents.ts
dmaklygin: hm, I tried to reproduce it locally, I commented the [line](https://github.com/toptal/picasso/pull/3653/files#diff-5b5ca055179bf6b7af1e34a61d150e4deb6f470b28c9e964bdbcc7764b95dc02R61) and focus still active

packages/picasso/src/LexicalEditor/LexicalEditor.tsx
```diff
@@ -163,45 +170,59 @@ const LexicalEditor = forwardRef<HTMLDivElement, Props>(function LexicalEditor(
...
-              <ContentEditable
-                className={classes.contentEditable}
-                data-testid={testIds?.editor}
+    <SettingsContext plugins={plugins}>
+      {({ nodes, settings, pluginsToRender }) => {
+        return (
+          <LexicalComposer
+            key={JSON.stringify(settings)}
+            initialConfig={editorConfig(nodes)}
+          >
```
dmaklygin, line 179: Maybe we could create another wrapper over the `LexicalComposer`?
Because, right now, `LexicalComposer` uses only variables from `settingsContext`.
I mean, there is kinda big responsibility on LexicalEditor right now: render SettingsContext, LexicalComposer, extra function, rendering toolbar with plugins + custom plugins.
Ideally, it should be like that:

```
<LexicalWrapper plugins={plugins}>
  <div onFocus={handleFocus} onBlur={handleBlur} tabIndex={-1}>
    ...
  </div>
</LexicalWrapper>
```

TomasSlama:
@toptal-anvil ping reviewers

packages/picasso/src/LexicalEditor/utils/synchronizeToolbarState.test.ts
```diff
@@ -14,6 +14,10 @@ import { ToolbarActions } from './toolbarState'
 import { synchronizeToolbarState } from './synchronizeToolbarState'
 import { getLexicalNode } from './getLexicalNode'
 
+// jest.mock('../nodes/CustomEmojiNode', () => ({
+//   default: jest.fn(),
+// }))
+
```
augustobmoura, line 20: Maybe we should remove this commented code?

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  static clone(node: CustomEmojiNode): CustomEmojiNode {
+    return new CustomEmojiNode(node.__src, node.__id)
+  }
+
+  constructor(src: string, id: string, key?: NodeKey) {
+    super(key)
+    this.__src = src
+    this.__id = id
+  }
```
augustobmoura, line 63: On the previous implementation, we agreed on using Picasso code conventions over Lexical's here, or did something change? We can also use the constructor trick to make the code simpler
```suggestion
  static getType() {
    return 'custom-emoji'
  }

  static clone(node: CustomEmojiNode): CustomEmojiNode {
    return new CustomEmojiNode(node.__src, node.__id)
  }

  constructor(public src: string, public id: string, key?: NodeKey) {
    super(key)
  }
```

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,153 @@
...
+  return node instanceof CustomEmojiNode
+}
+
+export const $createCustomEmojiNode = ({
+  src,
+  id,
+}: CustomEmojiPayload): CustomEmojiNode => {
+  // return new CustomEmojiNode(src, id)
+  return $applyNodeReplacement(new CustomEmojiNode(src, id))
+}
```
augustobmoura, line 153: I would strech Picasso code conventions to these methods as well. In fact it was a bit confusing at first glance, because of the `$` sign I thought these functions were from Lexical, when they are implemented on Picasso itself. I don't see why we would need to prefix them with dollar signs, as it only makes sense when internally used by Lexical, wdyt?
```suggestion
export const isCustomEmojiNode = (
  node: LexicalNode | null | undefined
): node is CustomEmojiNode => {
  return node instanceof CustomEmojiNode
}

export const createCustomEmojiNode = ({
  src,
  id,
}: CustomEmojiPayload): CustomEmojiNode => {
  // return new CustomEmojiNode(src, id)
  return $applyNodeReplacement(new CustomEmojiNode(src, id))
}
```

augustobmoura (commented):
Left some small comments, but it seems like its working correctly. One note though, should undo and redo be working on this? Or is this a different feature for a different ticket?

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
TomasSlama: @augustobmoura it is best practice to prefix functions with `$` when they can be used only in `editor.update` or `editor.read`

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
TomasSlama: you are right, I forgot to apply this suggestion 🙇

TomasSlama: it seems that typescript understands it, but webpack has problem with it.
image
So I at least removed the prefix

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,151 @@
...
+    const className = theme.customEmoji
+
+    if (className !== undefined) {
+      span.className = className
+    }
+
+    return span
+  }
+
+  // update() {}
```
OleksandrNechai, line 88:
```suggestion
```

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  // isInline() {
+  //   return true
+  // }
+
+  // canIndent() {
+  //   return false
+  // }
```
OleksandrNechai, line 132:
```suggestion
```

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
TomasSlama: done

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
TomasSlama: done

TomasSlama:
@toptal-anvil ping reviewers

dmaklygin (approved):

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
augustobmoura: That's weird, I don't know why would webpack break with that syntax, maybe some limitation with babel-typescript, but I'm not sure. Well, it is at least better now

packages/picasso/src/LexicalEmojiPlugin/nodes/CustomEmojiNode.tsx
augustobmoura: Well, this is a Lexical coding convention, I'm personally not a fan of [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation), and external contributors could find using the dollar sign confusing as well. If we decide on using it only for Lexical updates we need to document this somewhere, and probably have a discussion with the team, otherwise functions definitions will be inconsistent.

Let's leave this how it is and discuss it with the team tomorrow on the daily, is not a priority

augustobmoura (approved):

PR #3736, opened 2023-07-20
test: page in various screens

sashuk:
[FX-4090]

### Description

This pull request adds tests that make sure that the dropdown menu (with hamburger button) works and looks as expected on all screens.

**⚠️ Additional testing was performed**

The `disablePortal` was added to the `packages/picasso/src/PageHamburger/PageHamburger.tsx` in order to observe the menu when testing in Cypress (fix discovered by @mkrl)

### How to test

- Compare https://picasso.toptal.net/fx-4090-add-all-breakpoint-ranges-as-targets-in-pagetopbar-test/?path=/story/layout-page--page and https://picasso.toptal.net/?path=/story/layout-page--page on various screen sizes
- (Client Portal testing) Log in as Primary Contact for any Client (for example, for https://staff-portal.toptal.net/clients/295868) on staging, compare https://portal.toptal.net/ and https://client-portal-pr-6960.toptal.rocks/ on various screens
- (Staff Portal testing) Please see the https://staff-portal-pr-10698.toptal.rocks (https://github.com/toptal/staff-portal/pull/10698), but the change should not affect portal as it it uses custom menu
- (Talent Portal testing) The Portal has not migrated to new breakpoints, so the following can be done to test it locally
  - log in as active Talent (for example, as https://staff-portal.toptal.net/talents/3747845) at staging
  - check out `master` branch, set up default configuration
  - go to `node_modules/@toptal/picasso/PageHamburger/PageHamburger.js` file and add `disablePortal: true`
  
Screenshot 2023-07-24 at 21 09 38

  - run portal locally and check it on various screen sizes
 
### Development checks

- [N/A] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core
- [N/A] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).
- [x] Self reviewed
- [x] Covered with tests
- [x] Remove testing pull requests and branches before merge (https://github.com/toptal/staff-portal/pull/10698 and https://github.com/toptal/client-portal/pull/6960)

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4090]: https://toptal-core.atlassian.net/browse/FX-4090?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/component/Page.spec.tsx
+++ b/cypress/component/Page.spec.tsx
@@ -1,13 +1,51 @@
 import React from 'react'
 import type { PageSidebarProps } from '@toptal/picasso'
 import { Container, Menu, Page, Typography } from '@toptal/picasso'
+import type { BreakpointKeys } from '@toptal/picasso-provider/Picasso/config'
+import { PicassoBreakpoints } from '@toptal/picasso-provider/Picasso/config'
 
 const component = 'Page'
 const containerHeight = '30rem'
 
-const responsiveHappoTargets = [400, 600, 800, 1279, 1280, 1800].reduce<
-  Record<string, any>
->((acc, width) => {
+/**
+ * Produces an array of checkpoints – screen size values, needed for covering
+ * all existing breakpoints in tests (checkpoints are different from breakpoints
+ * by certain offset).
+ *
+ * @returns {number[]} Array of checkpoints
+ */
+const getCheckpoints = () => {
+  const offset = 5
+  const breakpoints = PicassoBreakpoints.breakpoints.values
+  const breakpointKeys = Object.keys(breakpoints)
+
+  return [
+    ...breakpointKeys
+      .filter(breakpointKey => breakpoints[breakpointKey as BreakpointKeys] > 0)
+      .map(
+        breakpointKey => breakpoints[breakpointKey as BreakpointKeys] - offset
+      ),
+    breakpoints[breakpointKeys.slice(-1) as unknown as BreakpointKeys] + offset,
+  ]
+}
+
+const checkpoints = getCheckpoints()
+
+// Sidebar menu has custom breakpoint at 1280px that changes its behavior, so 1280px
+// acts as a divider for "small" and "wide" page checkpoints
+const smallScreenCheckpoints = [
+  ...checkpoints.filter(width => width < 1280),
+  1279,
+]
+const wideScreenCheckpoints = [
+  1280,
+  ...checkpoints.filter(width => width >= 1280),
+]
+
+const responsiveHappoTargets = [
+  ...smallScreenCheckpoints,
+  ...wideScreenCheckpoints,
+].reduce<Record<string, any>>((acc, width) => {
   const name = `chrome-desktop-width-${width}`
 
   acc[name] = {
@@ -189,7 +227,7 @@ describe('Page', () => {
   })
 
   describe('for screen sizes smaller than 1280px', () => {
-    Cypress._.each([400, 600, 800, 1279], width => {
+    Cypress._.each(smallScreenCheckpoints, width => {
       describe(`when page is rendered on a ${width} screen width`, () => {
         it('renders hamburger menu and hides sidebar', () => {
           cy.viewport(width, 1000)
@@ -227,7 +265,7 @@ describe('Page', () => {
   })
 
   describe('for screen sizes equal or bigger than 1280px', () => {
-    Cypress._.each([1280, 1800], width => {
+    Cypress._.each(wideScreenCheckpoints, width => {
       describe(`when page is rendered on a ${width} screen width`, () => {
         it('does not show hamburger menu button and renders sidebar', () => {
           cy.viewport(width, 1000)
--- a/cypress/component/Page.spec.tsx
+++ b/cypress/component/Page.spec.tsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import type { PageSidebarProps } from '@toptal/picasso'
 import { Container, Menu, Page, Typography } from '@toptal/picasso'
-import { RemoteBrowserTarget } from 'happo.io'
 
 const component = 'Page'
 const containerHeight = '30rem'
@@ -9,10 +8,13 @@ const containerHeight = '30rem'
 const responsiveHappoTargets = [400, 600, 800, 1279, 1280, 1800].reduce<
   Record<string, any>
 >((acc, width) => {
-  acc[`chrome-desktop-width-${width}`] = new RemoteBrowserTarget('chrome', {
+  const name = `chrome-desktop-width-${width}`
+
+  acc[name] = {
+    name,
+    browser: 'chrome',
     viewport: `${width}x1024`,
-    applyPseudoClasses: true,
-  })
+  }
 
   return acc
 }, {})
--- a/cypress/component/Page.spec.tsx
+++ b/cypress/component/Page.spec.tsx
@@ -107,6 +107,10 @@ const Example = ({ sidebarProps }: ExampleProps) => (
         title='Default example'
         testIds={{ hamburger: 'hamburger-button' }}
       />
+      <Page.Banner>
+        We are now in the process of reviewing your profile. After your profile
+        has been checked, we will reach to you via email about next steps.
+      </Page.Banner>
       <Page.Content>
         <Sidebar {...sidebarProps} />
         <Page.Article>
@@ -180,6 +184,7 @@ describe('Page', () => {
           cy.get('body').happoScreenshot({
             component,
             variant: `page-menu-screen-smaller-than-1280/${width}-initial`,
+            targets: [`chrome-desktop-width-${width}`],
           })
 
           cy.getByTestId('hamburger-button').should('be.visible')
@@ -190,6 +195,7 @@ describe('Page', () => {
           cy.get('body').happoScreenshot({
             component,
             variant: `page-menu-screen-smaller-than-1280/${width}-opened-menu`,
+            targets: [`chrome-desktop-width-${width}`],
           })
 
           cy.getByTestId('hamburger-button').realClick()
@@ -199,6 +205,7 @@ describe('Page', () => {
           cy.get('body').happoScreenshot({
             component,
             variant: `page-menu-screen-smaller-than-1280/${width}-closed-menu`,
+            targets: [`chrome-desktop-width-${width}`],
           })
         })
       })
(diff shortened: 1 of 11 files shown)


sashuk:

sashuk:
@toptal-anvil ping reviewers

TomasSlama (changes requested):
it does not create expected images 
image

sashuk:
@TomasSlama Thank you for noticing, addressed!

Screenshot 2023-07-25 at 17 16 55

TomasSlama (commented):
Looks good, I think your new screenshot revealed a style bug
 
image
we are missing bottom spacing in the header.

Also I think that the one `diff` is not expected
image

sashuk:
@TomasSlama

> Looks good, I think your new screenshot revealed a style bug

Will be fixed in https://github.com/toptal/picasso/pull/3743, thanks!

> Also I think that the one diff is not expected

The test is flaky, I've put additional conditions so we take screenshots when the menu is actually opened

cypress/component/Page.spec.tsx
```diff
@@ -5,9 +5,24 @@ import { Container, Menu, Page, Typography } from '@toptal/picasso'
 const component = 'Page'
 const containerHeight = '30rem'
 
+const responsiveHappoTargets = [400, 600, 800, 1279, 1280, 1800].reduce<
```
mkrl, line 8: By the way, are there any guidelines as why those targets are selected? I was thinking we could get the list of desired breakpoints from `breakpoints.ts`, subtract N pixels and use that. There also seems to be more ranges that we support. Are there any components that use breakpoints we don't support?

mkrl (approved):
Awesome job. I've looked through the new screenshots and it looks solid. I have left one question (not critical)

cypress/component/Page.spec.tsx
sashuk: @mkasperowicz Great idea, thanks! The selection of resolutions where we want to check component (_checkpoints_) was automated based on the `breakpoints.ts` and the comment about special breakpoint of `1280px` was added

cypress/component/Page.spec.tsx
```diff
@@ -1,13 +1,51 @@
...
 const component = 'Page'
 const containerHeight = '30rem'
 
+const breakpoints = PicassoBreakpoints.breakpoints.values
+const breakpointKeys = Object.keys(breakpoints)
+const breakpointCheckPoints = [
+  ...breakpointKeys
+    .filter(breakpointKey => breakpoints[breakpointKey as BreakpointKeys] > 0)
+    .map(breakpointKey => breakpoints[breakpointKey as BreakpointKeys] - 5),
+  breakpoints[breakpointKeys.slice(-1) as unknown as BreakpointKeys] + 5,
```
augustobmoura, line 16: What is the purpose of this code? I'm a bit confused by the constants, maybe we could add some comments or name the magic numbers?

cypress/component/Page.spec.tsx
```diff
@@ -1,13 +1,66 @@
...
+  const breakpointKeys = Object.keys(breakpoints)
+
+  return [
+    ...breakpointKeys
+      .filter(breakpointKey => breakpoints[breakpointKey as BreakpointKeys] > 0)
+      .map(
+        breakpointKey => breakpoints[breakpointKey as BreakpointKeys] - offset
+      ),
+    breakpoints[breakpointKeys.slice(-1) as unknown as BreakpointKeys] + offset,
+  ]
```
augustobmoura, line 29: I think I understood this code, but maybe we could make it more clear if we simplified and add some comments to it, wdyt?
```suggestion
  const breakpointValues = Object.values(breakpoints).sort()

  return [
    ...breakpointValues
      .slice(1) // First breakpoint (0 to x) doesn't generate a checkpoint, as it is comprehends anything lower than the second
      .map(breakpoint => breakpoint - offset), // We subtract an offset so the width would be inside the range
    breakpointValues.slice(-1)[0] + offset,
```

cypress/component/Page.spec.tsx
```diff
@@ -1,13 +1,66 @@
...
 
+/**
+ * Produces an array of checkpoints – screen size values, needed for covering
+ * all existing breakpoints in tests (checkpoints are different from breakpoints
+ * by certain offset).
+ *
+ * @returns {number[]} Array of checkpoints
+ */
+const getCheckpoints = () => {
+  const offset = 5
```
augustobmoura, line 18: I would be even safer and use `1` as the offset, as having it with `5` would miss off-by-one errors. If possible I would even go as far as having a decimal offset, something like 0.01, or similar
```suggestion
  const offset = 1
```

cypress/component/Page.spec.tsx
augustobmoura, line 1: I would split the logic for the checkpoints on a different file, maybe even inside Picasso provider itself, as it is useful in multiple places. I could use this function on Storybook Happo settings, for example

packages/picasso/src/PageHamburger/PageHamburger.tsx
```diff
@@ -36,6 +36,8 @@ const PageHamburger = ({ id, 'data-testid': dataTestId }: Props) => {
         content: classes.responsiveWrapperContent,
         popper: classes.popper,
       }}
+      offset={{ top: 0.4 }}
```
augustobmoura, line 39: I would name this magic number so people in the future know it is about, maybe `LINE_POSITIONING_OFFSET` or something that makes better sense

packages/picasso/src/PageTopBar/__snapshots__/test.tsx.snap
```diff
@@ -75,24 +93,6 @@ exports[`Page.TopBar render with custom logo 1`] = `
...
-          style="opacity: 0; transform: scale(0.75, 0.5625); visibility: hidden;"
-        >
-          <div>
-            <div
-              id="hamburger"
-            />
-          </div>
-        </div>
-      </div>
-    </div>
```
augustobmoura, line 95: Does moving the hamburger outside of the portal make any noticeable difference?

packages/picasso/src/PageHamburger/PageHamburger.tsx
```diff
@@ -36,6 +36,8 @@ const PageHamburger = ({ id, 'data-testid': dataTestId }: Props) => {
...
+      disablePortal
```
augustobmoura, line 40: I would also add some explanation on why we are disabling the portal here, in specific explain the bug we hit, and when we expect it to be fixed, maybe a link to a cypress issue or a JIRA ticket

augustobmoura (commented):
I added some comments, mostly on documenting the workarounds and non-obvious choices

cypress/component/Page.spec.tsx
augustobmoura: We could also use named functions instead of line comments, to convey the motivation behind the choices, that are not obvious at first glance

packages/picasso/src/PageTopBar/__snapshots__/test.tsx.snap
sashuk: @augustobmoura Yep, as it renders it in a different place

cypress/component/Page.spec.tsx
sashuk: @augustobmoura I think checkpoints belong to testing only, so Picasso should not be aware of them, what do you think?

augustobmoura: Yeah, I was not sure about having it on picasso provider, but I think we could put it inside a util file at least, so we can use it on other stories and on happo config

cypress/component/Page.spec.tsx
sashuk: @augustobmoura Comment were added, thanks!

cypress/component/Page.spec.tsx
sashuk: @augustobmoura Changed to `1`, thanks!

packages/picasso/src/PageHamburger/PageHamburger.tsx
sashuk: @augustobmoura It was completely removed, thanks!

packages/picasso/src/PageHamburger/PageHamburger.tsx
sashuk: @augustobmoura The comment was added, but the problem itself goes down to the React itself, so even if we create ticket we probable never going to address it

augustobmoura: Did it work outside of Cypress? It worked when ran/deployed manually right? Wouldn't that already narrow the bug to either Cypress or Happo screenshots?

sashuk: @augustobmoura It breaks only in Cypress, so the problem is related to Cypress. The problem comes from `createPortal()` and @mkrl researched it even more and it is something wrong with React itself. 

Adding `disablePortal` does not affect consumers, so it seems to be a better option as the research will take significant amount of resources.

sashuk:

packages/picasso/src/PageHamburger/PageHamburger.tsx
augustobmoura: Yeah, I don't think we should spend too much time on cypress-specific problems, but maybe we could create an issue upstream so the project maintainers can know about it or help with debugging the problem. They can even check Picasso for problems

packages/picasso/src/test-utils/get-checkpoints.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Produces an array of checkpoints – screen size values, needed for covering
+ * all existing breakpoints in tests (checkpoints are different from breakpoints
+ * by certain offset).
+ *
+ * @returns {number[]} Array of checkpoints
+ */
+export const getCheckpoints = () => {
+  const offset = 1
+  const breakpoints = PicassoBreakpoints.breakpoints.values
+  const breakpointKeys = Object.keys(breakpoints)
```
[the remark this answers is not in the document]
sashuk: @augustobmoura Fixed, thanks!

augustobmoura (approved):
Nice work @sashuk :+1:

PR #3785, opened 2023-08-08
fix: make CircularButton responsive

mkrl:
[FX-4179](https://toptal-core.atlassian.net/browse/FX-4179)

### Description

According to BASE design, icons acting as buttons should accommodate for bigger icons.

### How to test

- Temploy
- Check the `Button` story
- There's a new `Responsive` story for the `CircularButton`
- Resize the screen to see how the icon changes

### Screenshots

| image | image |

### Development checks

- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


TomasSlama (commented):
should we also change the stories of the Page where we use the buttons in the header? 🤔

mkrl:
@toptal-anvil ping reviewers

packages/picasso/src/ButtonCircular/styles.ts
```diff
@@ -1,14 +1,20 @@
...
     root: {
       borderRadius: '50%',
       padding: 0,
       minWidth: 'initial',
-      width: '1.5em',
+      width: '2.5em',
+      height: '2.5em',
+      [screens('xl')]: {
+        width: '1.5em',
+        height: '1.5em',
```
TomasSlama, line 16: I don't remember what we agreed on, but wasn't the decision that we will provide `size` property that would be in charge of size? 🤔 So people could use even large buttons on bigger screens? 🤔

packages/picasso/src/Page/story/Default.example.tsx
```diff
@@ -67,11 +71,21 @@ const RightContent = () => (
...
 const ActionItems = () => {
+  // BASE design recommends using 24px icons on screens
+  // below the xl breakpoint for better accessibility
+  const screens = useScreens()
+  const Icon = screens(
+    {
+      xl: Heartbeat16,
+    },
+    Heartbeat24
+  ) as FC<Props>
```
TomasSlama, line 82: I'm thinking if we should have examples that would work on SSR by default, or leave it on users that are using SSR 🤔

packages/picasso/src/ButtonCircular/styles.ts
mkrl: As far as I remember, this was for icons that are parts of text buttons. Specifically for this (standalone icon that acts as a button), we agreed on keeping it automatic.

sashuk (commented):
Still think about things discussed yesterday and @mkrl comment that having responsive circular buttons and making users manually select the icon (`someIcon16` or `someIcon24`) based on screen size feels weird.

Taking into account https://github.com/toptal/picasso/pull/3788 and the fact that all the icons in Picasso now have both variants (16px and 24px), what do you think about allowing users provide `string` as `icon` parameter in a follow-up ticket?
```
<Button.Circular icon='ach'/>
```

This was we can make development experience better and select proper icon based on the screen size inside of the `<Button.Circular/>`, what do you think?

mkrl:
@sashuk I like it, but we need to carefully think of the implementation details. We can't have it import all the icons directly since it will no longer be possible to tree shake them out of the bundle. Lazy loading is an option, however with dynamic imports icons will still end up in the bundle, even not loaded on the client. Since we have hundreds of icons, this can very severely impact the total bundle size.

sashuk:
@mkrl Yep, I'll bring this up on Daily to check if we can have a task / spike for that, thanks!

packages/picasso/src/Page/story/Default.example.tsx
dmaklygin: I also vote for a single solution. Currently we allow to adjust icon size:
1. using useScreens
2. using media query

So, media queries are more preferable, because they are native and more efficient.

mkrl:
I've updated this PR to take `responsive` value from Picasso Provider into account. This PR is now based on https://github.com/toptal/picasso/pull/3823 to make use of the responsive icon set, the base will be changed to master once it is merged.

.changeset/eleven-elephants-whisper.md
```diff
@@ -0,0 +1,7 @@
+---
+'@toptal/picasso-provider': minor
+---
+
+### PicassoProvider
+
+- `responsive` prop is now accessible in root context and in `useAppConfig`
```
sashuk, line 7: Can you please specify if `RootContext` should contain the `responsive` property? Now it does not contain it https://github.com/toptal/picasso/blob/fx-4179-update-guidelines-and-stories-for-icon-size-usage/packages/picasso-provider/src/Picasso/RootContext.ts#L20, but `useAppConfig` can not as it was removed in https://github.com/toptal/picasso/pull/3785/commits/c8228c8ff1bdf99baae3b93ae0e4b491050e64b9#diff-e1a933defe5c6a0cba18020f525910de802ddb4b89f56f9c0ce494aaeb632545L17

packages/picasso/src/ButtonCircular/ButtonCircular.tsx
```diff
@@ -71,6 +74,7 @@ export const ButtonCircular: OverridableComponent<Props> = forwardRef<
       [classes.focused]: focused,
       [classes.hovered]: hovered,
       [classes.disabled]: disabled,
+      [classes.responsive]: responsive,
```
sashuk, line 77: Should `ButtonCircular` get the responsive value from the root context?

.changeset/eleven-elephants-whisper.md
mkrl: My bad, we have decided to not take context value into account since the behavior will strictly be "opt-in" for now, I'll update the description. Enabling it by default breaks too many things.

packages/picasso/src/ButtonCircular/ButtonCircular.tsx
mkrl: No longer a requirement, see the comment above. Sorry for the confusion, I've updated the description.

.changeset/eleven-elephants-whisper.md
sashuk: @mkrl Got it, thanks! In this case this changeset should be removed?

sashuk (approved):
Looks and works good! The minor comment in https://github.com/toptal/picasso/pull/3785#discussion_r1308649579 left

PR #3801, opened 2023-08-11
docs: add decision on spacing in picasso

sashuk:
[FX-4226]

### Description

This pull request adds an RFC for introducing new BASE design spacings described in https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing.

The overall plan is presented in https://toptal-core.atlassian.net/wiki/spaces/FE/pages/3389194785/Spacing.

### How to test

- Please review the https://github.com/toptal/picasso/blob/FX-4426-expose-base-spacing/docs/decisions/18-spacings.md

### Development checks

- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4226]: https://toptal-core.atlassian.net/browse/FX-4226?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/decisions/18-spacings.md
+++ b/docs/decisions/18-spacings.md
@@ -53,7 +53,7 @@ module.exports = {
 
 ## Proposal
 
-Picasso has to export spacing, compatible with [BASE design](https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing). The legacy way of specifying spacing (numbers and string constants) has to be gradually deprecated to encourage usage of BASE spacings. The [How to use spacings](https://picasso.toptal.net/?path=/story/tutorials-how-to-use-spacings--how-to-use-spacings) tutorial is updated to contain guidelines on using a new spacing system for a particular example.
+Picasso has to export spacing, compatible with [BASE design](https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing). The legacy way of specifying spacing (numbers and string constants) has to be gradually deprecated to encourage usage of BASE spacings. The [How to use spacings](https://picasso.toptal.net/?path=/story/tutorials-how-to-use-spacings--how-to-use-spacings) tutorial has to be updated to contain up-to-date guidelines on using a new spacing system.
 
 ### Technical implementation
 
@@ -78,7 +78,7 @@ export type SpacingType =
   | PicassoSpacing
 ```
 
-Exported `spacing` object is reused by components with spacing properties (`top`, `right`, `bottom`, `left`, `gap`, and `padded` properties in Container component).
+Exported `spacing` object is reused by components with spacing properties (`top`, `right`, `bottom`, `left`, `gap`, and `padded` properties in Container component and `offset` property in Dropdown component).
 
 ```tsx
 import { spacing } from '@toptal/picasso/utils'
--- a/docs/decisions/18-spacings.md
+++ b/docs/decisions/18-spacings.md
@@ -1,0 +1,116 @@
+# RFC Template
+
+## Problem
+
+BASE design defines [rules](https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing) on how to use spacing between page elements in frontend applications. Picasso has to align with design rules and has to provide tools for Picasso consumers to use BASE-compatible spacing in applications.
+
+## Current situation
+
+Picasso provides a specialised `SpacingType` type that can be either a numeric value or one of the pre-defined string constants (`xsmall`, `small`, etc.) that are resolved to a number (in both cases a numeric value corresponds to a number of `rem` units of spacing).
+
+As an example, the `Container` component has a `top` property that has `SpacingType` type. Based on the statistics of `top` property usage (searching for its occurrences in [this](https://github.com/search?q=org%3Atoptal+%22+top%3D%7B%22&type=code&p=1) GitHub search), Picasso consumers frequently use both numeric values and string constants.
+
+All of the `SpacingType` string constants resolve to numeric values that are present in [BASE design spacings](https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing#Base-Increment). Some of the BASE design spacings do not have a counterpart between `SpacingType` string constants. There are 5 `SpacingType` string constants and 12 BASE design spacings – for example, there is no string constant for `spacing-12` spacing. 
+
+## Existing approaches
+
+- [Chakra UI](https://chakra-ui.com/docs/styled-system/theme#spacing) exports spacings in a form of key-value pairs, where `key` is the pixel number value and `value` is the `rem` units string value
+
+```ts
+const spacing = {
+  space: {
+    px: '1px',
+    0.5: '0.125rem',
+    1: '0.25rem',
+    ...
+  }
+}
+```
+
+- [Caliber](https://github.com/toptal/caliber/blob/9a0b91110f1c82e07d30f684bb42b49e0e34f918/tailwind.preset.design-tokens.js#L2) defines spacings as key-value pairs and exports spacings in a form of CSS classes generated with TailwindCSS
+
+```js
+// Design tokens preset
+module.exports = {
+  theme: {
+    spacing: {
+      ...
+      4: '16px',
+      ...
+    }
+  }
+}
+```
+
+```css
+...
+/* padding: 16px */
+.p-4
+/* padding-top: 16px */
+.pt-4
+...
+```
+
+## Proposal
+
+### Goal
+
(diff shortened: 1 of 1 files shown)


sashuk:
@toptal-anvil ping reviewers

docs/decisions/18-spacings.md
```diff
@@ -0,0 +1,116 @@
...
+  12: 3,
+}
+```
+
+Exported `spacing` object with increments is reused by components with spacing properties (`top`, `right`, `bottom` and `left` properties in Container and `offset` property Dropdown)
+
+```tsx
+import { spacing } from '@toptal/picasso/utils'
+...
+<Container top={spacing[6]}/>
```
dmaklygin, line 79: this approach is also not the best one, I think. 
Developers must know or sort out spacing for their needs.

augustobmoura: We already allow arbitary spacing with numbers, `spacing[6]` is just a shortcut to the spacing design tokens

sashuk:
> from @whymancreates: @sashuk and @dmaklygin I think right now, my main challenge is that I am blindsided to how a space-10 on desktop should adjust to what space-N on mobile. If FX team sees this as a benefit for engineers then lets work on it and have something similar to the audi system. thoughts??

Let's discuss this topic again from the actual needs of our users and to evaluate the option of responsive spacings.

From my point of view responsive spacing is something that have not been requested by anybody and does not really solve any problem now. This way, if we introduce it, consumers (in our case Designers and Engineers) will encounter the following problems or extra effort spent:
- the need to understand the whole concept of tokens that changes the value depending on the screen width (popular UI libraries do not support this concept)
- Figma needs to somehow support tokens that have different values depending on the screen width, so it is comfortable to worn on designs
- need to implement additional logic on Picasso side

Taking into account all the issues above, it seems like it is not a good time for responsive spacing tokens as they are not needed. @whymancreates @dmaklygin can you please share your opinion? Maybe there are other obstacles or opportunities that I've missed?

whymancreates:
I agree for me it would be premature to jump into responsive spacing and tokens. In figma there is a way to automatically set these however, our users are far from being comfortable with this change aswell. I think we should take it in steps and first steps for us would make sense to provide a set of fixed values with proper namings (space-N) that can be used freely.

docs/decisions/18-spacings.md
```diff
@@ -0,0 +1,116 @@
...
+
+```ts
+export const spacing = {
+  0: `0rem`,
+  1: `0.25rem`,
+  ...
+  10: `2.5rem`,
+  12: `3rem`,
+}
+```
```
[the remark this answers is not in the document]
sashuk: @augustobmoura Thank you for the proposal!

After discussing with @whymancreates the plan is the following – gradually enforce usage of spacings that are mentioned in https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing, so the union type `SpacingType` will eventually will lose the `number` and `'small'`, `'medium'`, etc.

I'd propose to simplify the proposed structure and go with the following (list of spacings was shortened to show the idea)

```
type PicassoSpacing = 1.5 | 2.5
export const picassoSpacings: Record<number, PicassoSpacing> = {
  6: 1.5,
  10: 2.5
}
...
<Container top={{ xs: picassoSpacings[6] }}>Some text</Container>
```

This way, when `SpacingType` becomes `PicassoSpacing` only (no `number` and `'small'`, `'medium'`, etc.), the spacing will be properly typed like in the example below

Screenshot 2023-08-16 at 22 46 25

I am totally for the idea of properly typing the `export const spacing`, but propose to keep it as simple as possible, what do you think?

sashuk:
@toptal-anvil ping reviewers

sashuk:
@toptal-anvil ping reviewers

docs/decisions/18-spacings.md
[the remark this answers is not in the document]
sashuk: The commit with addressed change – https://github.com/toptal/picasso/pull/3801/commits/5d6303c2b650fe11f633b8fb42d048a565b23abf

sashuk:
@dmaklygin Thank you for the sync! The following was done as a follow-up:
- both pull request have better titles
- the deprecation plan for spacing in Picasso was described (I propose to go in two steps 3 weeks apart from each other)
- string constants **do** map to BASE spacings ✅ 
- custom usages were checked, there are around 10 non-BASE values of `Container.top` property in Toptal codebase

docs/decisions/18-spacings.md
```diff
@@ -0,0 +1,165 @@
...
+/* padding: 16px */
+.p-4
+/* padding-top: 16px */
+.pt-4
+...
+```
+
+## Proposal
+
+Picasso has to export spacing, aligned with [BASE design](https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing). The legacy way of specifying spacing (numbers and string constants) has to be gradually deprecated to encourage usage of BASE spacings. The [How to use spacings](https://picasso.toptal.net/?path=/story/tutorials-how-to-use-spacings--how-to-use-spacings) tutorial has to be updated to contain up-to-date guidelines on using a new spacing system.
```
OleksandrNechai, line 56: Picasso exports spacing, aligned with [BASE design](https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing)
Deprecate numbers and string constants as spacing specifiers

OleksandrNechai: Updating guidelines is not a part of proposal :) It does not make sense to do 2 first items and not update the guidelines.

docs/decisions/18-spacings.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Technical implementation
```
OleksandrNechai, line 58: Can implementation be not technical?

docs/decisions/18-spacings.md
```diff
@@ -0,0 +1,165 @@
...
+
+Picasso has to provide a new `spacing` object that includes proposed spacings in a form of key-value pairs (where the key is the index of the increment and the value is a corresponding value in `rem` units)
```
OleksandrNechai, line 60:
```suggestion
Picasso provides a new `spacing` object that includes proposed spacings in a form of key-value pairs (where the key is the index of the increment and the value is a corresponding value in `rem` units)
```

sashuk: @OleksandrNechai Fixed, thanks!

docs/decisions/18-spacings.md
sashuk: @OleksandrNechai Fixed, thanks!

docs/decisions/18-spacings.md
sashuk: @OleksandrNechai Fixed, thanks!

Regarding the update of guidelines – I think it should be kept to have a full picture what needs to be done as part of the proposal.

dmaklygin (approved):

docs/decisions/18-spacings.md
```diff
@@ -0,0 +1,165 @@
...
+
+```tsx
+import { spacing } from '@toptal/picasso/utils'
+...
+<Container top={spacing[6]}/>
+```
+
+### Deprecation of existing approaches
+
+After BASE spacing is introduced, the number and string constants as spacing value have to be deprecated. The deprecation plan has two steps:
```
OleksandrNechai, line 91: It looks more like the outline what needs to be done to migrate to this proposals. Less like a deprecation plan. Overall it is strange to have a deprecation plan in one RFC with API decision. What if one agrees with API and disagrees with the plan? 😄

docs/decisions/18-spacings.md
OleksandrNechai: 😄 okay, we also need to create tickets, merge PRs, notify teams and do many other things. Why guidelines are mentioned and everything else not?

OleksandrNechai (approved):

docs/decisions/18-spacings.md
sashuk: @OleksandrNechai Guidelines are really easy to miss, I was surprised that we even have them. So, even when the goal is reached after all the tickets created, there is still big chance that guidelines will be skipped.

docs/decisions/18-spacings.md
sashuk: @OleksandrNechai Due to yesterday discussions I think it is essential to have it here, as if we do not agree on the plan, then the API proposal will be affected as well. 

Let's say, we decide on migrating spacing in one run – in this case RFC will look differently, what do you think? 🙂

PR #3835, opened 2023-08-30
feat: [FX-4259] add new spacing values to picasso

dmaklygin:
[FX-4259]

### Description

Add new spacing values to Picasso, described in BASE. 

RFC: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md#technical-details

- [x] BASE spacings are exported according to RFC
- [x] BASE spacings are exported via theme, so they can be used in custom styles

### How to test

- Use temploy to play with SPACING variables

Example:
```
import { SPACING_4, SPACING_6 } from '@toptal/picasso/utils'

const Example = () => (
  <div>
    <Container bottom={SPACING_6}>Some text</Container>
    <Container left={SPACING_4}>Some more text with a small margin</Container>
  </div>
)
```

Example of new `spacings` variable coming from the theme:

```
export default ({ palette, sizes: { borderRadius }, spacings: { SPACING_6 } }: Theme) => {
  ...

}
```

### Development checks

- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4259]: https://toptal-core.atlassian.net/browse/FX-4259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/lovely-socks-cheat.md
+++ b/.changeset/lovely-socks-cheat.md
@@ -4,4 +4,4 @@
 '@toptal/picasso-shared': minor
 ---
 
-- add new Spacing values to picasso's theme and to `picasso-shared`. See RFC: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md#technical-details
+- add new spacing values to picasso's theme and to `picasso-shared`. See RFC: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md
--- a/.changeset/lovely-socks-cheat.md
+++ b/.changeset/lovely-socks-cheat.md
@@ -1,0 +1,7 @@
+---
+'@toptal/picasso-provider': minor
+'@toptal/picasso': minor
+'@toptal/picasso-shared': minor
+---
+
+- add new Spacing values to picasso's theme and to `picasso-shared`. See RFC: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md#technical-details
--- a/packages/picasso-provider/src/Picasso/PicassoProvider.tsx
+++ b/packages/picasso-provider/src/Picasso/PicassoProvider.tsx
@@ -12,9 +12,10 @@ import {
   typography,
   sizes,
   shadows,
+  spacings,
 } from './config'
 
-const picasso = {
+const picasso: ThemeOptions = {
   palette,
   layout,
   transitions,
@@ -37,6 +38,7 @@ const picasso = {
       notched: false,
     },
   },
+  spacings,
 }
 
 class Provider {
--- a/packages/picasso-provider/src/Picasso/config/spacings.ts
+++ b/packages/picasso-provider/src/Picasso/config/spacings.ts
@@ -1,45 +1,9 @@
+// ============================
+// TYPES
+// ============================
+
 // BASE-aligned spacing values in "rem" units
 export type PicassoSpacingValues = 0 | 0.25 | 0.5 | 0.75 | 1 | 1.5 | 2 | 2.5 | 3
-
-export class PicassoSpacing {
-  value: number
-
-  constructor(value: PicassoSpacingValues) {
-    this.value = value
-  }
-
-  valueOf() {
-    console.log('this.value: ', this.value)
-
-    return this.value
-  }
-
-  toString() {
-    return this.value.toString()
-  }
-}
-
-const SPACING_0 = new PicassoSpacing(0)
-const SPACING_1 = new PicassoSpacing(0.25)
-const SPACING_2 = new PicassoSpacing(0.5)
-const SPACING_3 = new PicassoSpacing(0.75)
-const SPACING_4 = new PicassoSpacing(1)
-const SPACING_5 = new PicassoSpacing(1.5)
-const SPACING_6 = new PicassoSpacing(2)
-const SPACING_7 = new PicassoSpacing(2.5)
-const SPACING_8 = new PicassoSpacing(3)
-
-const spacings = {
-  SPACING_0,
-  SPACING_1,
-  SPACING_2,
-  SPACING_3,
-  SPACING_4,
-  SPACING_5,
-  SPACING_6,
-  SPACING_7,
-  SPACING_8,
-}
 
 export type Sizes =
   | 'xxsmall'
@@ -49,6 +13,15 @@ export type Sizes =
   | 'large'
   | 'xlarge'
 
+export type SizeType<T extends Sizes> = T
+
+/** @deprecated **/
+export type DeprecatedSpacingType =
+  | number
+  | SizeType<'xsmall' | 'small' | 'medium' | 'large' | 'xlarge'>
+
+export type SpacingType = PicassoSpacing | DeprecatedSpacingType
+
 export enum SpacingEnum {
   xsmall = 0.5,
   small = 1,
@@ -57,11 +30,44 @@ export enum SpacingEnum {
   xlarge = 2.5,
 }
 
-export type SizeType<T extends Sizes> = T
+// ============================
+// CLASS DEFINITION
+// ============================
 
-export type SpacingType =
-  | number
-  | SizeType<'xsmall' | 'small' | 'medium' | 'large' | 'xlarge'>
-  | PicassoSpacing
+export class PicassoSpacing {
+  private value: PicassoSpacingValues
+
+  private constructor(value: PicassoSpacingValues) {
+    this.value = value
+  }
+
+  static create(value: PicassoSpacingValues): PicassoSpacing {
+    return new PicassoSpacing(value)
+  }
+
+  valueOf(): PicassoSpacingValues {
+    return this.value
+  }
+
+  toString(): string {
+    return this.value.toString()
+  }
+}
+
+// ============================
+// SPACING CONSTANTS
+// ============================
+
+const SPACING_VALUES: PicassoSpacingValues[] = [
+  0, 0.25, 0.5, 0.75, 1, 1.5, 2, 2.5, 3,
+]
+
+const spacings = Object.freeze(
+  SPACING_VALUES.reduce((acc, value, index) => {
+    acc[`SPACING_${index}`] = PicassoSpacing.create(value)
+
+    return acc
+  }, {} as Record<string, PicassoSpacing>)
+)
 
 export default spacings
--- a/packages/picasso-provider/src/Picasso/config/spacings.ts
+++ b/packages/picasso-provider/src/Picasso/config/spacings.ts
@@ -1,9 +1,5 @@
-// ============================
-// TYPES
-// ============================
-
 // BASE-aligned spacing values in "rem" units
-export type PicassoSpacingValues = 0 | 0.25 | 0.5 | 0.75 | 1 | 1.5 | 2 | 2.5 | 3
+type PicassoSpacingValues = 0 | 0.25 | 0.5 | 0.75 | 1 | 1.5 | 2 | 2.5 | 3
 
 export type Sizes =
   | 'xxsmall'
@@ -16,7 +12,7 @@ export type Sizes =
 export type SizeType<T extends Sizes> = T
 
 /** @deprecated **/
-export type DeprecatedSpacingType =
+type DeprecatedSpacingType =
   | number
   | SizeType<'xsmall' | 'small' | 'medium' | 'large' | 'xlarge'>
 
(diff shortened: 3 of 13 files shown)


dmaklygin:

packages/shared/src/types.ts
```diff
@@ -62,23 +75,6 @@ export type EnvironmentType<T extends Environments = BaseEnvironments> =
...
-  xsmall = 0.5,
-  small = 1,
-  medium = 1.5,
-  large = 2,
-  xlarge = 2.5,
-}
-
-export const spacingToRem = (spacing: SpacingType) =>
-  typeof spacing === 'number' ? `${spacing}rem` : `${SpacingEnum[spacing]}rem`
-
```
dmaklygin, line 81: as picasso-shared depends on picasso-provider, we should move definitions of spacings types and helprs to provider and then re-export them from picasso-provider package.

packages/picasso-provider/src/Picasso/utils/spacings.ts
```diff
@@ -0,0 +1,18 @@
...
+  return typeof spacing === 'number' || spacing instanceof PicassoSpacing
+}
+
+const spacingToRem = (spacing: SpacingType): string => {
+  return isNumericSpacing(spacing)
+    ? `${spacing}rem`
+    : `${SpacingEnum[spacing as SizeType<any>]}rem`
+}
+
+export { isNumericSpacing, spacingToRem }
```
dmaklygin, line 18: Decided to move those functions from `types.ts` file, which is definitely not a proper place for utils

dmaklygin:
@toptal-anvil ping reviewers

packages/picasso-provider/src/Picasso/utils/spacings.ts
sashuk, line 18: Do you think there could be a unit tests for those helpers?

packages/picasso-provider/src/Picasso/PicassoProvider.tsx
```diff
@@ -12,9 +12,10 @@ import {
   typography,
   sizes,
   shadows,
+  spacings,
```
sashuk, line 15: Just in case, we have discussed that `spacings` word sounds weird – however, it is actually grammatically correct https://www.wordhippo.com/what-is/the-plural-of/spacing.html / https://simple.wiktionary.org/wiki/spacings (so, we have different _types_ of spacing)

.changeset/lovely-socks-cheat.md
```diff
@@ -0,0 +1,7 @@
+---
+'@toptal/picasso-provider': minor
+'@toptal/picasso': minor
+'@toptal/picasso-shared': minor
+---
+
+- add new Spacing values to picasso's theme and to `picasso-shared`. See RFC: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md#technical-details
```
sashuk, line 7: NIT: 

```suggestion
- add new spacing values to picasso's theme and to `picasso-shared`. See RFC: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md#technical-details
```

packages/picasso-provider/src/Picasso/config/spacings.ts
```diff
@@ -0,0 +1,73 @@
...
+// SPACING CONSTANTS
+// ============================
+
+const SPACING_VALUES: PicassoSpacingValues[] = [
+  0, 0.25, 0.5, 0.75, 1, 1.5, 2, 2.5, 3,
+]
+
+const spacings = Object.freeze(
+  SPACING_VALUES.reduce((acc, value, index) => {
+    acc[`SPACING_${index}`] = PicassoSpacing.create(value)
```
sashuk, line 67: Can you please elaborate why `new PicassoSpacing()` is not used here?

sashuk: Using `index` does not work for all constants according to https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing – for example, `SPACING_8` is `2rem` value, but the index will be `7` (so the `SPACING_7` constant will be generated)

Screenshot 2023-08-30 at 17 32 40

packages/picasso-provider/src/Picasso/utils/spacings.ts
```diff
@@ -0,0 +1,18 @@
+import type { SizeType, SpacingType } from '../config/spacings'
+import { SpacingEnum, PicassoSpacing } from '../config/spacings'
+
+const isNumericSpacing = (spacing?: SpacingType): spacing is number => {
+  if (!spacing) {
+    return false
+  }
+
+  return typeof spacing === 'number' || spacing instanceof PicassoSpacing
```
augustobmoura, line 9: When `spacing instanceof PicassoSpacing`, spacing wouldn't be `number`, did you mean `!(spacing instanceof PicassoSpacing)`?

packages/picasso/src/Container/Container.tsx
```diff
@@ -101,12 +101,12 @@ export const Container = documentable(
...
-        ...(typeof top === 'number' && { marginTop: spacingToRem(top) }),
-        ...(typeof bottom === 'number' && {
+        ...(isNumericSpacing(top) && { marginTop: spacingToRem(top) }),
+        ...(isNumericSpacing(bottom) && {
           marginBottom: spacingToRem(bottom),
         }),
-        ...(typeof left === 'number' && { marginLeft: spacingToRem(left) }),
-        ...(typeof right === 'number' && { marginRight: spacingToRem(right) }),
+        ...(isNumericSpacing(left) && { marginLeft: spacingToRem(left) }),
+        ...(isNumericSpacing(right) && { marginRight: spacingToRem(right) }),
```
augustobmoura, line 109: These `isNumericSpacing` checks are unnecessary, as `spacingToRem` already treats all spacing cases

packages/picasso/src/utils/index.ts
```diff
@@ -66,3 +66,27 @@ export {
...
+  SPACING_0,
+  SPACING_1,
+  SPACING_2,
+  SPACING_3,
+  SPACING_4,
+  SPACING_5,
+  SPACING_6,
+  SPACING_7,
+  SPACING_8,
+}
```
augustobmoura, line 92: You could simplify it to:
```suggestion
export const {
  SPACING_0,
  SPACING_1,
  SPACING_2,
  SPACING_3,
  SPACING_4,
  SPACING_5,
  SPACING_6,
  SPACING_7,
  SPACING_8,
} = spacings
```

augustobmoura (changes requested):
The implementation should work, but I still think we are leaking too much information for users as of now. The intention of using classes is to prevent users from depending on the values directly, so we should keep the design of the API as [poka-yoke](https://en.wikipedia.org/wiki/Poka-yoke) as possible

packages/picasso-provider/src/Picasso/PicassoProvider.tsx
dmaklygin: It's not weird, there were thoughts about incorrect notation.

packages/picasso-provider/src/Picasso/config/spacings.ts
dmaklygin: static method is useful if we want to add additional logic or checks when creating an instance in the future

packages/picasso-provider/src/Picasso/config/spacings.ts
```diff
@@ -0,0 +1,73 @@
...
+  medium = 1.5,
+  large = 2,
+  xlarge = 2.5,
+}
+
+// ============================
+// CLASS DEFINITION
+// ============================
+
+export class PicassoSpacing {
```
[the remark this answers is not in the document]
dmaklygin: then we wouldn't check the exact type of the value, no?

packages/picasso-provider/src/Picasso/config/spacings.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const spacings = Object.freeze(
+  SPACING_VALUES.reduce((acc, value, index) => {
+    acc[`SPACING_${index}`] = PicassoSpacing.create(value)
+
+    return acc
+  }, {} as Record<string, PicassoSpacing>)
+)
+
+export default spacings
```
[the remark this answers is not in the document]
dmaklygin: In a technical details: https://github.com/toptal/picasso/blob/master/docs/decisions/18-spacings.md#technical-details
we should export them as constants. Since, your suggestion exports an array of elements, not exact constants, user cannot import a single constant using:

```
import { SPACING_0 } from '@toptal/picasso/utils'
```

I would keep my solution

packages/picasso-provider/src/Picasso/utils/spacings.ts
dmaklygin: nope, we expect that `spacing` is a number, either is a primitive number or an instance of `PicassoSpacing`

packages/picasso/src/Container/Container.tsx
dmaklygin: yeah, for this particular case it's necessary. We shouldn't render them if `spacing` is a string

packages/picasso-provider/src/Picasso/config/spacings.ts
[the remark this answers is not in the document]
augustobmoura: If we want to test if a value is of `PicassoSpacing` type, we export a
type-guard from the module and still keep the class constructor inaccessible
outside the module:
```ts
export const isPicassoSpacing = (target: unknown): target is PicassoSpacing =>
  target instanceof PicassoSpacingImpl
```

packages/picasso-provider/src/Picasso/config/spacings.ts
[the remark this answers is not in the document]
augustobmoura: That's true, I missed an `export` keyword on the suggestion, what I wanted to write was:
```suggestion
const spacings = [
  new PicassoSpacingImpl(0),
  new PicassoSpacingImpl(0.25),
  // ...
  new PicassoSpacingImpl(3),
]

export const [
  SPACING_0,
  SPACING_1,
  // ...
  SPACING_12,
] = spacings

export default spacings
```

packages/picasso-provider/src/Picasso/utils/spacings.ts
augustobmoura: `PicassoSpacing` is not a primitive `number`, I don't see why we "lie" to Typescript here and tell the class is a `number`, do we really need this method? :thinking:

packages/picasso/src/Container/Container.tsx
augustobmoura: Why though? There's nothing on the right-hand side of the `&&` operand that
would not accept any spacing, meaning the code below is the same as the current
one:
```ts
        ...(left && { marginLeft: spacingToRem(left) }),
        ...(right && { marginRight: spacingToRem(right) }),
```

augustobmoura: I see, then I would just change the signature of `isNumericSpacing` to:
```ts
const isNumericSpacing = (a: SpacingType): boolean =>
```

Instead of a type guard asserting that the type is a number:
```ts
const isNumericSpacing = (a: SpacingType): a is number =>
```

packages/picasso-provider/src/Picasso/utils/spacings.ts
```diff
@@ -0,0 +1,18 @@
+import type { SizeType, SpacingType } from '../config/spacings'
+import { SpacingEnum, PicassoSpacing } from '../config/spacings'
+
+const isNumericSpacing = (spacing?: SpacingType): spacing is number => {
+  if (!spacing) {
```
augustobmoura, line 5: This will fail when spacing is `0`, which is not what we want for this case, we still need to provide a margin/padding of `0` in that case, here we need to specifically test for nullity:
```suggestion
  if (spacing != null) {
```

packages/picasso-provider/src/Picasso/utils/spacings.ts
dmaklygin: yes!

dmaklygin:
@toptal-anvil ping reviewers

augustobmoura (approved):
LGTM!

PR #3926, opened 2023-10-20
fix(picasso-charts): fix BarChart for cases with 0 values

ievgrafov:
[SDE-1274]

### Description

Hi guys! We in Service Delivery are using `BarChart` to render spend chart and add label to all columns which are related to periods with breaks.

In some cases it may happen that we don't have spend data for given period but still want to show breaks icon. Currently it looks incorrect (see the original ticket). The reason for this is inability of BarChart to determine correct boundaries for value axis. In this PR I introduce a property that allows controlling boundaries of the axis manually.

Please let me know if the change is incorrect or there's a better way to achieve it. My goal here is to fix the bug so I'm pretty flexible with choosing the way to do it.

Update:
following comments to the PR I changed the implementation. Now instead of adding a prop to customize boundaries we just set default ones in case of too few ticks (0 or 1).

### How to test

- [Temploy](https://picasso.toptal.net/sde-1274-allow-passing-max-value-to-bar-chart)
- check example [here](https://picasso.toptal.net/sde-1274-allow-passing-max-value-to-bar-chart/?path=/story/picasso-charts-barchart--barchart#adjust-value-axis-boundaries)

### Screenshots

| image | image |

### Development checks

- [x] Self reviewed

**Breaking change**

- [n/a] codemod is created and showcased in the changeset
- [n/a] test alpha package of Picasso in StaffPortal

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[SDE-1274]: https://toptal-core.atlassian.net/browse/SDE-1274?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/component/BarChart.spec.tsx
+++ b/cypress/component/BarChart.spec.tsx
@@ -127,6 +127,34 @@ describe('BarChart', () => {
     })
   })
 
+  it('renders custom chart with customized indicators when values are 0s', () => {
+    cy.mount(
+      <TestBarChart
+        tooltip={false}
+        data={chartDataCustomTooltipZeroValue}
+        renderBarIndicators={({ dataKey }: any) => {
+          const indicator = INDICATORS[dataKey]
+
+          if (indicator) {
+            return (
+              <BarChartIndicator
+                label={indicator.label}
+                color={indicator.color}
+              />
+            )
+          }
+
+          return <></>
+        }}
+      />
+    )
+
+    cy.get('body').happoScreenshot({
+      component,
+      variant: 'custom-chart/custom-indicators-zero-value',
+    })
+  })
+
   it('hides label of each bar via passed `showBarLabel` prop being set to `false`', () => {
     cy.mount(<TestBarChart showBarLabel={false} tooltip={false} />)
 
@@ -187,5 +215,16 @@ const chartDataCustomTooltip = [
   {
     name: 'Los-Angeles',
     value: { infected: 2780, recovered: 3908 },
+  },
+]
+
+const chartDataCustomTooltipZeroValue = [
+  {
+    name: 'Berlin',
+    value: { infected: 0 },
+  },
+  {
+    name: 'Milan',
+    value: { infected: 0 },
   },
 ]
--- a/cypress/component/BarChart.spec.tsx
+++ b/cypress/component/BarChart.spec.tsx
@@ -127,34 +127,6 @@ describe('BarChart', () => {
     })
   })
 
-  it('renders custom chart with customized indicators when values are 0s', () => {
-    cy.mount(
-      <TestBarChart
-        tooltip={false}
-        data={chartDataCustomTooltipZeroValue}
-        renderBarIndicators={({ dataKey }: any) => {
-          const indicator = INDICATORS[dataKey]
-
-          if (indicator) {
-            return (
-              <BarChartIndicator
-                label={indicator.label}
-                color={indicator.color}
-              />
-            )
-          }
-
-          return <></>
-        }}
-      />
-    )
-
-    cy.get('body').happoScreenshot({
-      component,
-      variant: 'custom-chart/custom-indicators-zero-value',
-    })
-  })
-
   it('hides label of each bar via passed `showBarLabel` prop being set to `false`', () => {
     cy.mount(<TestBarChart showBarLabel={false} tooltip={false} />)
 
@@ -172,6 +144,36 @@ describe('BarChart', () => {
     cy.get('body').happoScreenshot({
       component,
       variant: 'vertical',
+    })
+  })
+
+  describe('when values are 0s', () => {
+    it('renders custom chart with customized indicators correctly', () => {
+      cy.mount(
+        <TestBarChart
+          tooltip={false}
+          data={chartDataCustomTooltipZeroValue}
+          renderBarIndicators={({ dataKey }: any) => {
+            const indicator = INDICATORS[dataKey]
+
+            if (indicator) {
+              return (
+                <BarChartIndicator
+                  label={indicator.label}
+                  color={indicator.color}
+                />
+              )
+            }
+
+            return <></>
+          }}
+        />
+      )
+
+      cy.get('body').happoScreenshot({
+        component,
+        variant: 'custom-chart/custom-indicators-zero-value',
+      })
     })
   })
 })
--- a/packages/picasso-charts/src/BarChart/BarChart.tsx
+++ b/packages/picasso-charts/src/BarChart/BarChart.tsx
@@ -161,14 +161,10 @@ const BarChart = <T extends string>({
     interval: 0,
     tick: { width: TICK_WIDTH },
   }
-  const valueDomain = [
-    BOTTOM_DOMAIN,
-    ticks.length >= 2 ? ticks[ticks.length - 1] : BOTTOM_DOMAIN + 10,
-  ]
   const valueAxisProps = {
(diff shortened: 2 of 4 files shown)


packages/picasso-charts/src/BarChart/BarChart.tsx
TomasSlama, line 1: I don't like adding new property for this, I looked into recharts how this issue should be handled. It seems for this case we should use `domain` property of `YAxis`.

This could be the change on line 180
```
const defaultDomain = [ticks[0], ticks[ticks.length - 1]]
  const valueAxisProps = {
    width: Y_AXIS_WIDTH,
    ticks: ticks,
    domain: ticks.length === 0 ? [0, 'dataMax + 10'] : defaultDomain,
    tickFormatter: valueAxisTickFormatter,
  }
```

WDYT? we would not need to add new property, we would just fix it for everyone

packages/picasso-charts/src/BarChart/story/CustomValueAxisBoundaries.example.tsx
```diff
@@ -0,0 +1,51 @@
...
+  },
+  {
+    name: 'Google',
+    value: { 'engineers hired': 0 },
+  },
+]
+
+const INDICATORS: any = {
+  Google: { color: palette.blue.light, label: 'E' },
+  Amazon: { color: palette.purple.main, label: 'E' },
```
TomasSlama, line 18:
```suggestion
  Apple: { color: palette.purple.main, label: 'E' },
```

packages/picasso-charts/src/BarChart/story/index.jsx
```diff
@@ -75,3 +75,9 @@ page
       'You can format the ticks on the value axis by providing the format function in the `valueAxisTickFormatter` property.',
     takeScreenshot: false,
   })
+  .addExample('BarChart/story/CustomValueAxisBoundaries.example.tsx', {
+    title: 'Adjust value axis boundaries',
+    description:
+      'You can adjust min and max value on value axis. The example shows usage of this property to render empty bars in custom scale.',
+    takeScreenshot: false,
```
TomasSlama, line 82: If we decide to not add new propety we will not need this example, but I would treat it as a bug of BarChart and we should cover this case with visual test in cypress. Look for `BarChart.spec.ts`

TomasSlama (changes requested):

packages/picasso-charts/src/BarChart/BarChart.tsx
ievgrafov: I saw an answer on SO that suggested to fix it with domain, yeah. But since we don't have something existing for this in `BarChart` props, I thought that would be better to add more flexible solution while we're anyway touching this part. But if you think it would better to just fix it here without introducing new prop, sure, I'll update the PR now, it's even simpler :+1:

TomasSlama: 👍, I would agree to add flexibility, but it looks like a bug to me, I don't think that the current implementation is usable anywhere 🤔, so we would just increase complexity for no reason

ievgrafov: @TomasSlama I made the changes you requested, could you please take another look?

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -161,10 +161,14 @@ const BarChart = <T extends string>({
     interval: 0,
     tick: { width: TICK_WIDTH },
   }
+  const valueDomain = [
+    BOTTOM_DOMAIN,
+    ticks.length >= 2 ? ticks[ticks.length - 1] : BOTTOM_DOMAIN + 10,
```
TomasSlama, line 166:
```suggestion
    ticks.length >= 2 ? ticks[ticks.length - 1] : 'dataMax + 10',
```
wdyt?

ievgrafov: typechecks fail with it https://github.com/toptal/picasso/actions/runs/6616543696/job/17971041119?pr=3926, so I changed to calculating same value here and passing it as number

ievgrafov: hmm, looking at [screenshot](https://happo.io/a/675/p/848/compare/7d934fa29545d40b99d4600596d5a32b38fa6902/ab47d3c95c88c2444ad40bf767bae2f37bf75930?t=added) in happo I'd say it doesn't fix the issue :thinking: . This one is for this commit https://github.com/toptal/picasso/pull/3926/commits/ab47d3c95c88c2444ad40bf767bae2f37bf75930 . I also checked code of `findTopDomain` and it returns `Math.max(calculatedMax, 10)` so I believe it should already handle such problem :confused: I'll try to investigate further, looks weird so far

packages/picasso-charts/src/BarChart/utils/find-top-domain/find-top-domain.ts
```diff
@@ -37,11 +37,12 @@ const findTopDomain = (data: DataItem[], stackedBars?: string[][]) => {
   const maxValue = getDataMaxValue(
     stackedBars ? aggregateStackedBarsWithData(data, stackedBars) : data
   )
+
   const base10 = Math.floor(Math.log10(maxValue))
   const roundedMaxValue = Math.pow(10, base10)
   const topDomain = roundedMaxValue * Math.ceil(maxValue / roundedMaxValue)
 
-  return Math.max(topDomain, 10)
+  return Math.max(topDomain || 0, 10)
```
ievgrafov, line 45: `topDomain` can be NaN because of line 43: when `roundedMaxValue` is 0, `maxValue / roundedMaxValue` is NaN. And this NaN then goes all the way down to value axis domain :man_shrugging: So I guess best is to catch it here

ievgrafov:
@toptal-anvil ping reviewers

packages/picasso-charts/src/BarChart/utils/find-top-domain/find-top-domain.ts
TomasSlama: nice!, great solution

TomasSlama (approved):

TomasSlama:
@toptal-anvil ping reviewers

cypress/component/BarChart.spec.tsx
```diff
@@ -127,6 +127,34 @@ describe('BarChart', () => {
     })
   })
 
+  it('renders custom chart with customized indicators when values are 0s', () => {
```
sashuk, line 130: NIT: Can you please move condition to a separate `describe` block (https://toptal-core.atlassian.net/wiki/spaces/ENG/pages/1120830304/Frontend+Testing+-+Unit+testing#Describe)?

sashuk (approved):
Left a minor comment, code looks good!

cypress/component/BarChart.spec.tsx
ievgrafov: thanks, I updated it now

ievgrafov:
@toptal-anvil ping reviewers

sashuk (approved):

ievgrafov:
@sashuk @TomasSlama folks, what should I do to get this one merged? There's failing check in happo but it looks unrelated to the PR itself

PR #4074, opened 2024-01-01
[CI-3043] Switch to actions/create-jira-deployment

jcobarreto:
[CI-3043]

### Description
The CI team is currently working on replacing the `external-proxy` with Google Identity Aware Proxy (IAP). More info [here](https://cloud.google.com/iap/docs/concepts-overview).

IAP is a Google service that allows control access to web applications running on GCP. It provides a security layer for applications by ensuring that only authenticated and authorized users/service accounts can access them. To programmatically access a resource protected by the IAP, we need to generate a JWT token and pass it alongside the request to authenticate before accessing the protected resource.

Currently, we use the `jenkins-job-trigger-action` to trigger Jenkins jobs from a GH workflow. Since this is a public action repo and to not expose our internal infrastructure implementation, we created the new private action ([trigger-jenkins-job](https://github.com/toptal/actions/tree/main/trigger-jenkins-job)) to implement the IAP necessary steps for authentication and authorization. This new action requires some new inputs (secrets) used in the action’s steps to generate the authentication token.

The IAP is already enabled for our Jenkins servers in alternative URLs during the migration, that’s why we use the input `jenkins_url` as a secret. Its main role is to make it easier and transparent when we turn the key to enable IAP in the production’s Jenkins URLs.

In this PR, we switch to the `actions/create-jira-deployment` action, which uses the new `trigger-jenkins-job` action that implements the underlying steps to authenticate with IAP when triggering a Jenkins job from a GitHub workflow.

### Review
- [x] Get 👍 from code review
- [x] Squashed related commits together
- [x] Removed labels that block merging (wip, do-not-merge, needs-management-approval)

[CI-3043]: https://toptal-core.atlassian.net/browse/CI-3043?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,7 +29,6 @@ jobs:
       issues: write
       pull-requests: write
     timeout-minutes: 45
-
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -164,7 +163,6 @@ jobs:
     runs-on: ubuntu-latest
     permissions: write-all
     needs: [static-checks]
-
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -224,7 +222,6 @@ jobs:
       contents: write
       id-token: write
     needs: [build-docker-image]
-
     steps:
       - name: GSM Secrets
         id: secrets_manager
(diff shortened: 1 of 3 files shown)


TomasSlama (approved):
looks good

Jakub-Kwak (approved):

PR #4208, opened 2024-03-26
[CPT-2029] Add support to optional tooltip for query builder fields 

toptalwadiibasmi:
[CPT-2029]

### Description

- We want to improve the query builder so that its users can provide a tooltip to help their end-users understand what a field is for.
- To achieve this functionality, we added an optional property to override the styles of the tooltip popover

**Note:**
For the clause: **Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core**
The implementation was issued from a Toptal design; please reach out to me to receive the link if you want to.

### How to test

- [Temploy](https://picasso.toptal.net/CPT-2029-fe-add-optional-tooltip-for-a-menu-item-in-the-picasso-select)
- Open the query builder storybook
- Scroll to field description
- Open the fields selector
- Move the mouse over the first name property and notice the popover
- The other select input should not regress

### Screenshots

| - |image |

### Development checks

- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[CPT-2029]: https://toptal-core.atlassian.net/browse/CPT-2029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/base/Tooltip/src/Tooltip/Tooltip.tsx
+++ b/packages/base/Tooltip/src/Tooltip/Tooltip.tsx
@@ -70,6 +70,8 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
   tooltipRef?: React.Ref<HTMLDivElement>
   /** A node, or a function that returns node. The container will have the portal children appended to it. */
   container?: ContainerValue
+  /** Allows customizing the tooltip popper style */
+  popperClassName?: string
 }
 
 const useStyles = makeStyles<Theme>(styles, { name: 'PicassoTooltip' })
@@ -81,6 +83,7 @@ export const Tooltip = forwardRef<unknown, Props>((props, ref) => {
     placement,
     interactive,
     className,
+    popperClassName,
     style,
     open,
     onOpen,
@@ -168,6 +171,7 @@ export const Tooltip = forwardRef<unknown, Props>((props, ref) => {
           [classes.compact]: compact,
           [classes.noMaxWidth]: maxWidth === 'none',
         }),
+        popper: popperClassName,
       }}
       className={className}
       style={style}
--- a/packages/base/Tooltip/src/Tooltip/Tooltip.tsx
+++ b/packages/base/Tooltip/src/Tooltip/Tooltip.tsx
@@ -53,7 +53,7 @@ const getOffset = (
   const isVertical =
     placement.startsWith('top') || placement.startsWith('bottom')
 
-  return !isVertical ? result.reverse().join(',') : result.join(',')
+  return (isVertical ? result : result.reverse()).join(',')
 }
 
 export interface Props extends BaseProps, HTMLAttributes<HTMLDivElement> {
--- a/packages/base/Tooltip/src/Tooltip/Tooltip.tsx
+++ b/packages/base/Tooltip/src/Tooltip/Tooltip.tsx
@@ -11,7 +11,9 @@ import type { TooltipProps } from '@material-ui/core'
 import { Tooltip as MUITooltip } from '@material-ui/core'
 import cx from 'classnames'
 import type { BaseProps } from '@toptal/picasso-shared'
-import { usePicassoRoot } from '@toptal/picasso-provider'
+import { remToNumber, spacingToRem } from '@toptal/picasso-shared'
+import type { PicassoSpacing } from '@toptal/picasso-provider'
+import { SPACING_0, usePicassoRoot } from '@toptal/picasso-provider'
 import { Typography } from '@toptal/picasso-typography'
 
 import styles from './styles'
@@ -26,6 +28,11 @@ export type MaxWidthType = 'none' | 'default'
 
 export type PlacementType = TooltipProps['placement']
 
+export type OffsetType = {
+  left?: PicassoSpacing
+  top?: PicassoSpacing
+}
+
 const delayDurations: { [k in DelayType]: number } = {
   short: 200,
   long: 500,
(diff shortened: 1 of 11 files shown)


packages/picasso-query-builder/src/QueryBuilder/story/FieldDescription.example.tsx
```diff
@@ -0,0 +1,87 @@
...
+      field: 'age',
+      operator: '=',
+      valueSource: 'value',
+      value: '21',
+    },
+  ],
+  combinator: 'and',
+}
+
+const fields = [
```
rasitozcan, line 25: We do not need this many fields to demonstrate, lets keep it simpler, like 2-3 fields.

packages/picasso-query-builder/src/QueryBuilder/story/FieldDescription.example.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useState } from 'react'
+import {
+  QueryBuilder,
+  type RuleGroupTypeAny,
+} from '@toptal/picasso-query-builder'
+
+const initialQuery: RuleGroupTypeAny = {
+  rules: [
+    {
+      field: 'name',
```
rasitozcan, line 10: Those fields names do not seem to be correct, so no selected fields in the storybook.

packages/picasso-query-builder/src/FieldSelector/styles.ts
```diff
@@ -0,0 +1,12 @@
+import { createStyles } from '@material-ui/core/styles'
+
+export default () =>
+  createStyles({
+    tooltipOptionText: {
+      width: '100%',
+    },
+    tooltip: {
+      // Style to push the tooltip to the right outside the select option text
+      left: '0.5rem!important',
```
sashuk, line 10: Is there way to avoid `!important` here? Why is it needed in the first place, does it get overwritten by Popper styles?

toptalwadiibasmi: Yes, it does. There is a left style that is apparently provided as inline styles and calculated to always come above this style. This is why I had to add the importance here.

packages/picasso-query-builder/src/QueryBuilder/story/FieldDescription.example.tsx
toptalwadiibasmi: I added this many to have the autocomplete displayed and match the design, but I can do so too.

packages/picasso-query-builder/src/FieldSelector/styles.ts
sashuk: @toptalwadiibasmi My concern is that we rely on the passed classname (and we spill the internal details of tooltip) rather than on special popper settings that are supposed to solve your problem with offset https://popper.js.org/docs/v1/#modifiersoffset.

I think it will be better to go with another Tooltip property (for example, `offset` object that can take only Spacing values that are converted to `rem`) instead of `popperClassName` to hide any mentions of popper from consumers, @toptal/frontend-experience-eng what do you think?

sashuk: @toptalwadiibasmi The idea is to abstract the fact that popper is used inside of Tooltip, so the Tooltip could have an `offset` prop (that is `{ top?: Spacing, left?: Spacing, ... }`) that will pass the offset value to Popper component via popper options https://github.com/toptal/picasso/blob/1665a449e7d41f40cabb620ee10cfdcbf8976d00/packages/base/Tooltip/src/Tooltip/Tooltip.tsx#L146.

This way, Tooltip component will not have to change its API when popper.js is replaced with some other popper library in the future (as the `offset` property name is not specific to any internal details of Tooltip implementation and is only about exposed configuration of tooltip).

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -35,6 +42,27 @@ const getDelayDuration = (delay: DelayType, isTouchDevice: boolean) => {
   return isTouchDevice ? 0 : delayDurations[delay]
 }
 
+const convertRemToPixels = (rem: number) => {
```
sashuk, line 45: How about using `pxFromRem` from https://github.com/toptal/picasso/blob/81fe2e899c9a5ea756d70adf40b24c57decd0cec/packages/shared/src/styles/index.ts#L44?

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -35,6 +42,20 @@ const getDelayDuration = (delay: DelayType, isTouchDevice: boolean) => {
...
+  offset: OffsetType
+): string => {
+  const { left = SPACING_0, top = SPACING_0 } = offset
+
+  const result = [pxFromRem(spacingToRem(left)), pxFromRem(spacingToRem(top))]
+
+  const isVertical =
+    placement.startsWith('top') || placement.startsWith('bottom')
+
+  return !isVertical ? result.reverse().join(',') : result.join(',')
```
sashuk, line 56:
```suggestion
  return (isVertical ? result : result.reverse()).join(',')
```

sashuk:
@toptal-anvil ping reviewers

angelinastavniiciuc (approved):
the code looks good and the implementation too, works great on Storybook!
btw have you checked if it works as expected in Staff portal with alpha release ?

toptalwadiibasmi:
How can I redeploy the branch? I don't see the changes reflected on the temploy.

sashuk:
@toptalwadiibasmi You need to wait for https://github.com/toptal/picasso/actions/runs/8469086539/job/23207031465?pr=4208 check

Screenshot 2024-03-28 at 17 03 01

sashuk:
@toptal-anvil ping reviewers

rasitozcan (approved):

packages/picasso-query-builder/src/QueryBuilder/story/index.jsx
```diff
@@ -38,6 +38,11 @@ page
     description: 'Customized header',
   })
 
+  .addExample('QueryBuilder/story/FieldDescription.example.tsx', {
+    title: 'Field Description',
+    description: `Hover over 'specificFieldName' to display field description in a tooltip`,
```
rasitozcan, line 43: When I said `specificFieldName` i was assuming you would add the actual field name like:

```suggestion
    description: `Hover over 'First  name' field in the field selector dropdown to display field description in a tooltip`,
```

PR #4224, opened 2024-04-05
Revert "[FX-5115] Update types of compound components"

TomasSlama:
Reverts toptal/picasso#4216

### DIFF
--- a/packages/base/Breadcrumbs/src/BreadcrumbsCompound/index.ts
+++ b/packages/base/Breadcrumbs/src/BreadcrumbsCompound/index.ts
@@ -1,13 +1,6 @@
 import { Breadcrumbs } from '../Breadcrumbs'
 import { BreadcrumbsItem } from '../BreadcrumbsItem'
 
-type BreadcrumbsCompoundType = typeof Breadcrumbs & {
-  Item: typeof BreadcrumbsItem
-}
-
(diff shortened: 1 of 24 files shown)


TomasSlama:
@toptal-anvil ping reviewers

sashuk (approved):

packages/base/Breadcrumbs/src/BreadcrumbsCompound/index.ts
```diff
@@ -1,13 +1,6 @@
...
-  Item: typeof BreadcrumbsItem
-}
-
-export const BreadcrumbsCompound: BreadcrumbsCompoundType = Object.assign(
-  Breadcrumbs,
-  {
-    Item: BreadcrumbsItem,
-  }
-)
+export const BreadcrumbsCompound = Object.assign(Breadcrumbs, {
```
OleksandrNechai, line 4: So, that's important part is not it? Maybe worth a comment?

dmaklygin (approved):

packages/base/Breadcrumbs/src/BreadcrumbsCompound/index.ts
TomasSlama: @OleksandrNechai we will be working on proper fix, we need to just revert it now to break other projects.

PR #4227, opened 2024-04-05
[CPT-1313] Add customized footer to picasso qb

angelinastavniiciuc:
[CPT-1313](https://toptal-core.atlassian.net/browse/CPT-1313)

### Description

In this PR we provided footer property to QB. This is the first part of implementation as the goal is to move Clear and Run query buttons to Staff portal Query Builder usage along with possibility to have customized footer content.

### How to test

- Temploy

### Screenshots

 | After.                                  |
 | --------------------------------------- |
 | Screenshot from 2024-04-05 15-07-50 |

### Development checks

- [x] Self reviewed

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


angelinastavniiciuc:

angelinastavniiciuc:
@toptal-anvil ping reviewers

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -187,7 +193,7 @@ const QueryBuilder = ({
       <Container
         className={cx(classes.global, classes.root)}
         flex
-        padded={hideControls ? undefined : 'medium'}
+        padded={hideControls && !padding ? undefined : 'medium'}
```
sashuk, line 196: Can we achieve the same functionality by wrapping Query Build into Container with custom padding, so to leave it to the consumers if they want Query Builder to be padded? It seems like extra `padding` property for Query Builder  could be avoided

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -238,6 +244,9 @@ const QueryBuilder = ({
             />
           </Container>
         )}
+        {footer && (
```
sashuk, line 247: NIT: What do you think about leaving provision of `testId` for footer and header up to consumers? If consumers want to have `testId` in footer, they can provide footer already with `testId` in it (I can see that footer is implemented the same way as `header` property in https://github.com/toptal/picasso/pull/4227/files#diff-72d559b4abc1acb004cc2227b47c68d80a07723c779ce8679a258a6111f580c6R200, but still)

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
angelinastavniiciuc: unfortunately by wrapping QB in other Container and providing padding to it does not give the expected result :(  
btw we are planning to remove hideControls property and provide functionality from the QB usage already 
so we are going to have only padding property which will take care padding

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
angelinastavniiciuc: hmm, I think it makes sense, will adjust this part

packages/picasso-query-builder/src/types/query-builder.ts
```diff
@@ -92,5 +92,4 @@ export type TestId = {
   valueEditor?: string
   fieldSelector?: string
   validationErrors?: string
-  header?: string
```
TomasSlama, line 95: As I mentioned before, we should not remove this. It can be used in some other projects for testing. This would be a breaking change.

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -187,13 +193,11 @@ const QueryBuilder = ({
...
         flex
-        padded={hideControls ? undefined : 'medium'}
+        padded={hideControls && !padding ? undefined : 'medium'}
         direction='column'
         gap='small'
       >
-        {header && (
-          <Container data-testid={testIds?.header}>{header}</Container>
-        )}
+        {header && <Container>{header}</Container>}
```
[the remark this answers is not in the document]
angelinastavniiciuc: it was suggested by @sashuk :)) 
and I checked occurrences of this test-id and couldn't find any usage of it at the moment so it is safe to remove and make it possible to consumers to add test-id if they want to use header or footer

packages/picasso-query-builder/src/types/query-builder.ts
angelinastavniiciuc: I removed it in order to keep it the same as footer so if we are going to have test-id for header and footer in Picasso or we will provide it on usage since both properties are optional https://github.com/toptal/picasso/pull/4227#discussion_r1554947964
and there are no usage of this test-id atm so it will not break anything

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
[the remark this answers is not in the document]
sashuk: @TomasSlama Yep, the idea is that as header and footer are nodes that are provided by consumers, then why should Query Builder be responsible for setting test ids for these elements (as suggested in https://github.com/toptal/picasso/pull/4227#discussion_r1554947964) – what do you think?

sashuk:
@angelinastavniiciuc Please let me sync with @TomasSlama so we unblock the pull request as soon as possible

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
angelinastavniiciuc: this thing will break existing styles on Staff Portal QB 
we don't have any changes related to this `padded` property in Staff portal right now :thinking:

sashuk: @angelinastavniiciuc 

> unfortunately by wrapping QB in other Container and providing padding to it does not give the expected result :(

This is very interesting, can you please specify why? I've just tried it in https://picasso.toptal.net/?path=/story/layout-container--container and it worked as expected

Screenshot 2024-04-08 at 11 27 36


Screenshot 2024-04-08 at 11 27 45

angelinastavniiciuc: @sashuk yep the padding is applied, but all styles are defined in QueryBuilder styles (background-color for example) so if we are going to provide padding on Container which will wrap usage of Query builder we also should provide corresponding styles to it, the thing is this is not supposed right now on Staff portal and other projects which are using QB now 
so if I will remove padded property from Picasso Query Builder we will have style differences on Staff portal and it is not expected

sashuk: @angelinastavniiciuc 

> btw we are planning to remove hideControls property

NIT: Maybe it is time to mark `hideControls` property as deprecated one like in https://github.com/toptal/picasso/blob/4b444a8b75a7b7ebec7e127877bd5f0ae7a60a12/packages/base/DatePicker/src/DatePicker/DatePicker.tsx#L91

> Yep the padding is applied, but all styles are defined in QueryBuilder styles (background-color for example) so if we are going to provide padding on Container which will wrap usage of Query builder we also should provide corresponding styles to it, the thing is this is not supposed right now on Staff portal and other projects which are using QB now

Got it, thanks! Then, as @TomasSlama suggested, it will be better to keep the same API as in Containers and have `padded` property be a `SpacingType`

angelinastavniiciuc: > Maybe it is time to mark hideControls property as deprecated one like in
not sure if it is right time :)) as it is being used now and removal will take place when we will prepare Staff portal and other projects to removal :thinking: 
and we will have one more PR to Picasso to remove it and I believe we can either add this comment or remove it at all

sashuk:
@angelinastavniiciuc 

After discussing with @TomasSlama we have the following proposals:

1. Introduced `padding` boolean property should be named `padded`, have type `SpacingType` (you can inherit it directly from `ContainerProps['padded']` type) and default value of `SPACING_6` instead of `medium`. Reasoning – the `padding` sounds like a CSS property, so it might be confusing why it is boolean. At the same time, `padded` property with `SpacingType` will be consistent with the same `padded` property for Container component.
2. Test ids for header and footer can be brought back for now (sorry for back and forth) to avoid breaking change (we can not be 100% confident that nobody uses them as Picasso is open-source library). However, in the future it is better to remove them, so Query Builder provides only placeholders for header and footer and is not aware of test ids that header and footer need.
3. It will be great to have a follow-up ticket created to delete unclear logic in `padded={hideControls && !padded ? undefined : padded}`, as it is not obvious why `hideControls` property affects `padded` property at all. One of the approaches to do it (please see [this document](https://toptal-core.atlassian.net/wiki/spaces/FE/pages/3474391105/Breaking+change+management) for more details on dealing with breaking changes)

  - deprecate `hideControls` property (so consumer stop using it)
  - announce to Query Builder consumers that it will be deleted in 3-4 months
  - delete `hideControls` property

angelinastavniiciuc:
@sashuk Thank you for summarizing everything I got the part with test-ids and padding :+1: 
related to `hideControls` removal - we have the ticket already https://toptal-core.atlassian.net/browse/CPT-2188
my concern about deprecate or not this property is about the thing that we are still going to use this prop in order to hideControls from Picasso Query Builder while the needed controls will be implemented on consumers level, in order to not break existing functionality, does it make sense to make it deprecated at this point ?
we are going to make the change on Staff portal to have Clear query and Run query button defined there, and also announce the change to other teams, at the moment when the Controls will be moved to consumers in all places where it is required we will perform the removal :thinking:

sashuk:
> my concern about deprecate or not this property is about the thing that we are still going to use this prop in order to hideControls from Picasso Query Builder while the needed controls will be implemented on consumers level, in order to not break existing functionality, does it make sense to make it deprecated at this point ?

If you mark property as deprecated today, this will provide a warning in IDE for all consumer that will try to use it. It will not stop them from using it, but it will be a good opportunity to warn that it will be gone soon in https://toptal-core.atlassian.net/browse/CPT-2188

packages/picasso-query-builder/src/types/query-builder.ts
sashuk: Updated in https://github.com/toptal/picasso/pull/4227#issuecomment-2042412118

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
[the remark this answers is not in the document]
sashuk: Updated in https://github.com/toptal/picasso/pull/4227#issuecomment-2042412118

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
sashuk: Update in https://github.com/toptal/picasso/pull/4227#issuecomment-2042412118

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
sashuk: Updated in https://github.com/toptal/picasso/pull/4227#issuecomment-2042412118

TomasSlama: it is only mark, that it will be removed in the future, so new users will use it wisely

sashuk:
@angelinastavniiciuc Can you please also specify if there is BASE design for this change (with padding)?

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -84,12 +89,14 @@ const QueryBuilder = ({
...
   onSubmit,
   customValueEditor = ValueEditor,
+  footer,
   hideControls,
   header,
   enableDragAndDrop = false,
   resetOnFieldChange = true,
   totalCount,
   totalCountLoading,
+  padding = false,
```
TomasSlama, line 99:
```suggestion
  padded = SPACING_6
```

angelinastavniiciuc: about this part btw, I don't think that it will work as expected:
we want to use hideControls for the case when there are no controls and in this case we do not expect padding like here
Screenshot from 2024-04-08 14-52-40
but when we are going to move controls to the level of consumers and also will provide the hideControls value, but we want to have padding and for this reason I added this additional property, so if I will provide default value then it will always be applied and we will have this case which is not expected 
Screenshot from 2024-04-08 14-58-28

angelinastavniiciuc:
> @angelinastavniiciuc Can you please also specify if there is BASE design for this change (with padding)?

hmm, wdym ? We are not introducing any new design in this case it was done previously I just want to have additional property to control padding for the cases when we are going to have hideControls, but at the same time have the padding

sashuk:
@angelinastavniiciuc 

> hmm, wdym ? We are not introducing any new design in this case it was done previously

So, this is approved from Design side that Query Builder can have a padding? Just wanted to double-check that Picasso stays 100% BASE-compatible

angelinastavniiciuc:
> @angelinastavniiciuc
> 
> > hmm, wdym ? We are not introducing any new design in this case it was done previously
> 
> So, this is approved from Design side that Query Builder can have a padding? Just wanted to double-check that Picasso stays 100% BASE-compatible

yeah I think so if it was provided before I assume that it was covered, unfortunately I can not double check now since @rasitozcan is OOO and he was responsible to migrate QB to Picasso :thinking:

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
angelinastavniiciuc: the main reason was to have possibility to allow hideControls, but leave padding without introducing change to the places where it is already used with this logic, but if you consider that it should be change (with the thing in mind that it will affect other places on QB update) then we can do it, and on library update also update parts with usage :thinking: 
but my thought was to have padded set to false by default to still have opportunity to hideControls together with padding  and use it all together with hideControls where we will provide controls from consumer, but with required padding

sashuk: @angelinastavniiciuc What do you think about splitting this pull request?

I can see that the part with footer is ready and it seems like changeset and the overall pull request description are about footer, so I propose to do changes to `hideControls` - related things in a different pull request / ticket, what do you think?

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
```diff
@@ -187,7 +198,7 @@ const QueryBuilder = ({
       <Container
         className={cx(classes.global, classes.root)}
         flex
-        padded={hideControls ? undefined : 'medium'}
+        padded={hideControls && !padded ? undefined : SPACING_6}
```
TomasSlama, line 201: so if padded is not defined we set undefined otherwise specing 6? It seems not correct to me.

`padded={hideControls ? undefined : padded` and padded should be set 6 by default. If someone needs padding 0 they just need to provide spacing_0

angelinastavniiciuc: so in this case it is okay to have this change first and update usage with SPACING_0 with new version of QB ?

TomasSlama: hm, I probably don't get it. Let me write down my thoughts.

Previously we had `padded={hideControls ? undefined : 'medium'}`

So by default we need to have the same functionality.
`padded={hideControls ? undefined : padded}` but on line 103 we need to add `padded = SPACING_6`.

In Staff Portal, if we need to remove the padding we just use `<QueryBuilder padded={SPACING_0} />`

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
sashuk: @angelinastavniiciuc Regarding `hideControls`, maybe the modification should be added when you actually give consumers opportunity to provide their own controls?

I am not aware of plan on your implementation, but in case you plan to provide controls the same way as footer (so it will be ReactNode), then you could have padding set completely internally? Basically, this could be the plan for the QueryBuilder component

```
Now
- if hideControls is true, then controls are hidden, padding is zero
- if hideControls is false, then controls are not hidden, padding is SPACING_6
- hideControls property is deprecated, not deleted yet

Next major version after breaking change
- hideControls property is deleted
- padded property is added so consumers can control padding (default value is SPACING_6)
- controls: ReactNode property is added so consumers can provide their own controls
```

angelinastavniiciuc: I assume that controls will be part of the Footer, but I agree with the proposal so I'll create new ticket for it and remove the changes from this PR :))

sashuk: @angelinastavniiciuc Thank you!

angelinastavniiciuc:
@sashuk @TomasSlama new ticket created https://toptal-core.atlassian.net/browse/CPT-2201
the only thing related to hideControls I left is the updated SPACING :))

packages/picasso-query-builder/src/QueryBuilder/QueryBuilder.tsx
angelinastavniiciuc: https://github.com/toptal/picasso/pull/4227#issuecomment-2042873442

TomasSlama (approved):

sashuk (approved):
Thank you @angelinastavniiciuc!

toptalwadiibasmi (approved):
Looks great! great job.

PR #4264, opened 2024-04-26
[FX-4593] Migrate tag

ruslan-sed:
[FX-4593]

### Description

This PR migrates `Tag` component to Tailwind. It also removes `Chip` component because it is not a part of MUI Base.

### How to test

- [Temploy](https://picasso.toptal.net/fx-4593-migrate-tag)

### Development checks

- [x] Self reviewed

**Breaking change**

- [x] codemod is created and showcased in the changeset

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4593]: https://toptal-core.atlassian.net/browse/FX-4593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/base/Tag/package.json
```diff
@@ -27,7 +27,7 @@
     "@toptal/picasso-typography": "2.0.0",
     "@toptal/picasso-utils": "1.0.2",
     "ap-style-title-case": "^1.1.2",
-    "classnames": "^2.5.1"
+    "tailwind-merge": "^2.2.2"
   },
   "sideEffects": [
     "**/styles.ts",
```
TomasSlama, line 33: we are missing TailwindCSS in peer dependencies and also now we can remove the MUI from peer dependencies

packages/base/Tag/src/Indicator/Indicator.tsx
```diff
@@ -21,20 +19,21 @@ export interface Props extends BaseProps {
...
-  const classes = useStyles()
 
   return (
     <div
       role='img'
       {...restProps}
-      className={cx(classes.root, className, classes[color])}
+      className={twMerge(
+        'w-2 h-2 rounded-[50%]',
+        className,
```
TomasSlama, line 34: the order in `twMerge` matters, I believe className should be always the last one, so someone can override internal classnames

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -1,26 +1,27 @@
...
 import { toTitleCase } from '@toptal/picasso-utils'
+import { twMerge } from 'tailwind-merge'
 
-import { Chip } from '../Chip'
-import styles from './styles'
+import { variantsRootClasses } from './styles'
 
-export type DivOrAnchorProps = AnchorHTMLAttributes<HTMLAnchorElement> &
-  HTMLAttributes<HTMLDivElement>
-export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
```
TomasSlama, line 23: Was this tested in the Staff Portal?

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -81,45 +78,55 @@ export const Tag = forwardRef<HTMLDivElement, Props>(function Tag(props, ref) {
...
+
+      onDelete(event)
     }
   }
 
   return (
-    <Chip
-      {...htmlAttributes}
+    <Root
+      {...rootRest}
```
TomasSlama, line 90:
```suggestion
  aria-disabled={disabled}
 ```

File: packages/base/Tag/src/Tag/Tag.tsx [hunk 5]
```diff
@@ -33,46 +34,42 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
-      <span className={classes.innerLabel}>
-        {titleCase ? toTitleCase(children) : children}
-      </span>
-      {endAdornment}
-    </>
-  )
+  const rootRest: RootRest = { ...rest }
+
+  if (onDelete || onClick) {
+    rootRest.role = 'button'
```
TomasSlama, line 67: are you sure that we should override `role` even if the user specifies it differently? 🤔 
I would expect `role = role || ((onDelete || onClick) ? 'button' : undefined)`, but I didn't check the previous implementation

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -81,45 +78,55 @@ export const Tag = forwardRef<HTMLDivElement, Props>(function Tag(props, ref) {
 
     if (onDelete) {
       event.preventDefault()
-      onDelete()
+
+      event.stopPropagation()
```
TomasSlama, line 82: why do we need this?

packages/base/Tag/src/Tag/__snapshots__/test.tsx.snap
```diff
@@ -6,23 +6,22 @@ exports[`Tag dismissable label should render dismissable label 1`] = `
     class="Picasso-root"
   >
     <div
-      class="MuiChip-root PicassoLabel-root MuiChip-deletable"
+      class="transition-none text-lg border-[0.041625rem] border-solid rounded-[6.25rem] h-6 max-w inline-flex justify-center items-center cursor-default bg-white group align-middle text-graphite border-gray"
       data-testid="tag"
       role="button"
-      tabindex="0"
```
TomasSlama, line 12: we are missing tabindex

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
```diff
@@ -38,9 +44,15 @@ const TagCheckable = forwardRef<HTMLDivElement, Props>(function TagCheckable(
...
+      className={twMerge(
+        `hover:border-graphite-700 transition-all duration-[350ms] ease-in-out cursor-pointer`,
+        className,
+        checked ? 'hover:border-red-500 hover:text-red-500' : '',
+        hovered ? 'border-graphite-700' : '',
+        checked && hovered ? 'border-red-500 text-red-500' : ''
+      )}
       icon={icon}
-      onClick={handleClick}
+      onClick={onChange || onClick ? handleClick : undefined}
```
TomasSlama, line 55: we already have the check in the function itself by using the `?`, do we need the condition even here? 🤔

packages/base/Tag/src/TagRectangular/types.ts
```diff
@@ -1,11 +1,7 @@
-import type { HTMLAttributes, AnchorHTMLAttributes } from 'react'
 import type { BaseProps, TextLabelProps } from '@toptal/picasso-shared'
 
 import type { IndicatorProps } from '../Indicator'
 
-export type DivOrAnchorProps = AnchorHTMLAttributes<HTMLAnchorElement> &
```
TomasSlama, line 6: I was not here if this was discussed, but we are changing interface of the component so this needs to be also taken as a breaking change, we need to make sure that it will not be breaking anything in the Staff Portal, could you also please summarize for me why do we need this?

packages/base/Tagselector/src/TagSelector/__snapshots__/test.tsx.snap
```diff
@@ -19,22 +19,21 @@ exports[`TagSelector disabled render 1`] = `
         >
           <div
             aria-disabled="true"
-            class="MuiChip-root PicassoLabel-root PicassoLabel-disabled MuiChip-deletable"
+            class="transition-none text-lg border-[0.041625rem] border-solid rounded-[6.25rem] h-6 max-w inline-flex justify-center items-center cursor-default bg-white group align-middle text-gray border-gray pointer-events"
             role="button"
-            tabindex="0"
```
TomasSlama, line 24: tabindex is missing

TomasSlama (commented):
I didn't test in storybook yet, but left some comments and questions in the code 🙇

packages/base/Tag/src/Tag/__snapshots__/test.tsx.snap
ruslan-sed: We discussed this with @sashuk and removed it intentionally because it was not added intentionally by us but by MUI Chip.

packages/base/Tag/src/TagRectangular/types.ts
ruslan-sed: Sure. On Friday retro we discussed `{...rest}` props. We decided not to use this approach to have more control over component API. So users will not use what we don't provide intentionally.

packages/base/Tagselector/src/TagSelector/__snapshots__/test.tsx.snap
ruslan-sed: Answered to other similar comment

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
ruslan-sed: Just for optimization purposes, so this function will not even be attached as an event handler and called.

packages/base/Tag/src/Tag/Tag.tsx
ruslan-sed: Without this one test was failing in Tag Selector.  [Here](https://happo.io/a/675/p/848/compare/fce6f6279e52162f08152fb88bd8d4afce1fa7b6/23cfe139127c961288b513d95305bdd4aee95e6e#TagSelector-default%5C%2Fafter-deleted-item) it is. 

The issue is with `With closeOnSelect disabled` (check `TagSelector` story). When you delete tag there, the list of tags is opened. And it shouldn't.

I assume, that MUI Chip had `stopPropagation` call.

packages/base/Tag/src/Indicator/Indicator.tsx
ruslan-sed: I`m not sure about always the last one. It will not break anything if it will be the last one here. Thats for sure. But if it would be the last one in `Tag` it will break styles for `disabled` for `TagCheckable`. 

Please check that. Maybe it also needs some refactoring.

.changeset/pretty-buckets-destroy.md
```diff
@@ -0,0 +1,8 @@
+---
+'@toptal/picasso-tag': major
+---
+
+### Tag
+
+- remove MUI Chip component
```
sashuk, line 7: Maybe `remove MUI Chip component` can be deleted, as consumer do not even know how `MUI Chip` is related to Tag component? 
Instead, you could add a note that component now support only properties that are explicitly defined in Picasso (as we stop support of `...rest`)

packages/base/Tag/package.json
sashuk: Just found out that we actually have `TailwindCSS is added as a peer dependency of the component` in https://toptal-core.atlassian.net/wiki/spaces/FE/pages/3694723111/PI-3379+Development#Acceptance-criteria-for-all-migrated-components

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
TomasSlama: I'm not sure that the optimization is worth it here 🤔 We don't usually do this in other components and I believe we should prefer clean code.

packages/base/Tag/src/Tag/Tag.tsx
TomasSlama: but this is not clear right away from the code, do you think it might be worth leaving a comment here so we know in the future why we have it here?

ruslan-sed: Sure, good idea 👍

packages/base/Tag/src/Tag/__snapshots__/test.tsx.snap
ruslan-sed: I will investigate this

packages/base/Tag/src/TagRectangular/types.ts
ruslan-sed: As discussed during office hour, this was reverted.

packages/base/Tag/src/Tag/Tag.tsx
ruslan-sed: Those changes where reverted

.changeset/pretty-buckets-destroy.md
ruslan-sed: Removed MUI Chip line. 
And the `...rest` part is again there.

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -81,45 +78,55 @@ export const Tag = forwardRef<HTMLDivElement, Props>(function Tag(props, ref) {
...
+      {propIcon ? (
+        <span className='w-min h-min flex items-center mr-[-0.25rem] ml-3'>
+          {React.cloneElement(propIcon, {
+            color: disabled ? 'grey' : 'dark-grey',
+          })}
+        </span>
+      ) : undefined}
+
+      <span className='flex gap-2 px-3 overflow-hidden'>
+        <span className='text-xxs min-w-0 font-semibold overflow-hidden overflow-ellipsis whitespace-nowrap'>
```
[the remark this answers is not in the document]
ruslan-sed: Probably we can. But this will make another dependency and increase the Tag package size, right?

packages/base/Tag/src/Tag/Tag.tsx
OleksandrNechai: This comment is nice now. But imagine someone reads it in 5 years. What useful information it brings? Only confusion. Developer will think (and hate us):
What is MUI Chip? Why should I care? What is `With closeOnSelect disabled` (imagine its deleted or renamed). 
Looks like valuable part is: `Prevents options list to open, because if event propagates the component thinks someone has regularly clicked on it and opens options list`

ruslan-sed: I have few thoughts about this.

In this case we can also have a question: why don't we just fix this in `TagSelector` component? And we can just do that. But, removing `event.stopPropagation()` still can cause issues because it can be used somewhere by some of our users. 

So the case with `TagSelector` is not the most important one. 

Because of this, I think it's worth mentioning MUI Chip here as a real reason for this command. 

Maybe something like this:
```
Before it was in MUI Chip that was used by this component.
We need to support this to not break anything on the users' side.
       
Without it, TagSelector's options list is opened after
removing tag in with closeOnSelect disabled.    
```

What do you @TomasSlama, @OleksandrNechai think?

packages/base/Tag/src/Tag/Tag.tsx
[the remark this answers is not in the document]
TomasSlama: The thing is that we need to be consistent. We want to use Typography for most of the text as it follows BASE. If for some reason BASE would make some changes, we can update Typography, which will be applied across all components.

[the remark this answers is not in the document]
TomasSlama: Typography is used almost everywhere so we don't need to care about bundle size as it will be already in bundle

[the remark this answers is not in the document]
ruslan-sed: Oh, yes, it is already on the dependencies list. All right, then I will use it here.

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +35,97 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
   const {
-    as = 'div',
+    as: Root = 'div',
     className,
     disabled,
     endAdornment,
-    hovered,
     children,
-    icon,
+    icon: propIcon,
```
sashuk, line 52: Is there reason to rename the `icon` prop? To reduce the complexity of the component the code below could be moved to a separate component, what do you think?

```jsx
{propIcon ? (
  <span className='w-min h-min flex items-center mr-[-0.25rem] ml-3'>
    {React.cloneElement(propIcon, {
      color: disabled ? 'grey' : 'dark-grey',
    })}
  </span>
) : undefined}
```

ruslan-sed: Yeah, I was thinking about moving it to a separate component to reduce complexity. Will do it if after addressing another  things complexity still will be high. 

And I don't remember why I renamed it. Now it doesn't make any sense. Will undo this.

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +37,95 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
+      {...rest}
+    >
+      <TagIcon icon={icon} disabled={disabled} />
+
+      <span className='flex gap-2 px-3 overflow-hidden items-center'>
+        <Typography
+          size='xsmall'
+          color='inherit'
+          weight='semibold'
+          noWrap={true}
```
TomasSlama, line 110:
```suggestion
          noWrap
```

packages/base/Tag/src/TagIcon/TagIcon.tsx
```diff
@@ -0,0 +1,23 @@
...
+const TagIcon = ({ icon, disabled }: Props) => {
+  if (!icon) {
+    return null
+  }
+
+  return (
+    <span className='w-min h-min flex items-center mr-[-0.25rem] ml-3'>
+      {React.cloneElement(icon, {
+        color: disabled ? 'grey' : 'dark-grey',
+      })}
```
TomasSlama, line 18: is this really needed? I believe that Icon is by default `fill='currentColor'`, so if we add proper font color to the span it should be applied to the icon. Or not? 🤔

ruslan-sed: I just moved this part of the code without checking. 
You are right, it works fine without it.

ruslan-sed:
@toptal-anvil ping reviewers

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
-          : undefined
-      }
-      label={label}
-      deleteIcon={
+      onClick={onClick}
+      tabIndex={onDelete || onClick ? 0 : undefined}
+      {...rest}
+    >
+      {icon && (
+        <span className='w-min h-min flex items-center mr-[-0.25rem] ml-3'>
```
TomasSlama, line 105:
```suggestion
        <span className='w-min h-min flex items-center -mr-1 ml-3'>
```

TomasSlama: https://tailwindcss.com/docs/space#using-negative-values

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
+
+        {endAdornment}
+      </span>
+
+      {onDelete && (
         <span
           aria-label='delete icon'
           role='button'
-          className={classes.deleteIcon}
+          className='w-min h-min flex items-center cursor-pointer ml-[-0.5rem] mr-2'
```
TomasSlama, line 128:
```suggestion
          className='w-min h-min flex items-center cursor-pointer -ml-2 mr-2'
```

packages/base/Tag/src/TagRectangular/TagRectangular.tsx
```diff
@@ -28,31 +21,37 @@ export const TagRectangular = forwardRef<HTMLDivElement, Props>(
...
+        {indicator && (
+          <div className='ml-1 flex content-center flex-wrap'>
+            <Indicator color={indicator} />
+          </div>
+        )}
+
+        <Typography
+          size='xxsmall'
+          weight='semibold'
+          noWrap={true}
```
TomasSlama, line 46:
```suggestion
          noWrap
```

TomasSlama (commented):
sorry 😇, I found a few more small things. Other than that it looks good!

TomasSlama:
most of the visual tests seems OK, but this one seems like it could be related 
image
Do you expect this change?

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
-    hovered,
     children,
     icon,
     onDelete,
+    onClick,
     style,
     titleCase: propsTitleCase,
     variant = 'light-grey',
+    role,
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
sashuk, line 59: NIT: Maybe to leave a comment why we want to leave out `hover`?

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
+
+      <span className='flex gap-2 px-3 overflow-hidden items-center'>
+        <Typography
+          size='xsmall'
+          color='inherit'
+          weight='semibold'
+          as='span'
+          noWrap
+        >
+          {titleCase ? toTitleCase(children) : children}
```
sashuk, line 118: https://picasso.toptal.net/?path=/story/components-typography--typography has property `titleCase`, maybe to pass it directly to component and avoid extra `if`?

packages/base/Tag/src/TagRectangular/TagRectangular.tsx
```diff
@@ -28,31 +21,37 @@ export const TagRectangular = forwardRef<HTMLDivElement, Props>(
...
+          className={twMerge(
+            'text-white mx-1 w-max',
+            variant === 'light-grey' ? 'text-graphite-800' : ''
+          )}
+        >
+          {titleCase ? toTitleCase(children) : children}
```
sashuk, line 52: Same as in https://github.com/toptal/picasso/pull/4264/files#r1595344915

packages/base/Tag/src/TagIcon/TagIcon.tsx
ruslan-sed: Turns out, it needs to be there, because we only need two colors actually.

ruslan-sed:
@TomasSlama about this:
image

No, I dont. I think it can be related to the fact, that on the master branch these elements are not aligned to bottom for some unknown reason:
image

Since it si related to container, I did not digged into it.

packages/base/Tag/src/Tag/Tag.tsx
ruslan-sed: Oh, nice 👍

packages/base/Tag/src/TagIcon/TagIcon.tsx
TomasSlama: so if we add the `text-grey-500` or the proper equivalent to the span itself, it would not work? 🤔

ruslan-sed:
> @ruslan-sed I think it might be related to different display option `block/inline`

You are right!

ruslan-sed:

OleksandrNechai (approved):
Did not test, left few comments. Looks good to go 👍

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
+
       onDelete()
     }
   }
 
   return (
-    <Chip
-      {...htmlAttributes}
+    <Root
+      role={role || (onDelete || onClick ? 'button' : undefined)}
```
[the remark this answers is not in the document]
ruslan-sed: In this case `onDelete || onClick ? 'button' : role' `role` that user provided would be used only if there is no `onDelete` or `onClick`. In current code `role` that user provided would always have priority.

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
-        root: classes.root,
-        label: classes.label,
-        clickable: classes.clickable,
-      }}
-      className={cx(className, classes[variant], {
-        [classes.hovered]: hovered,
-        [classes.disabled]: disabled,
-      })}
+      className={twMerge(
+        `transition-none border-[0.041625rem] border-solid rounded-[6.25rem]
```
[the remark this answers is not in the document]
ruslan-sed: Shouldn't we?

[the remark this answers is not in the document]
sashuk: I knew this link will be useful someday 😃 

https://github.com/toptal/picasso/pull/1593#discussion_r498227521

packages/base/Tag/src/TagIcon/TagIcon.tsx
ruslan-sed: yes, it would 🤦‍♂️

packages/base/Tag/src/Tag/Tag.tsx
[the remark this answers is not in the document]
TomasSlama: lul 3000 PRs ago 😄

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +35,109 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
-      deleteIcon={
+      onClick={onClick}
+      tabIndex={onDelete || onClick ? 0 : undefined}
+      {...rest}
+    >
+      {icon && (
+        <span
+          className={twMerge(
+            'w-min h-min flex items-center -mr-1 ml-3',
+            disabled ? '@text-gray-200' : 'text-graphite-700'
```
TomasSlama, line 108:
```suggestion
            disabled ? 'text-gray-200' : 'text-graphite-700'
```
I guess this is just typo

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +35,109 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
+        </span>
+      )}
+
+      <span className='flex gap-2 px-3 overflow-hidden items-center'>
+        <Typography
+          size='xsmall'
+          color='inherit'
+          weight='semibold'
+          as='span'
+          titleCase={titleCase}
```
TomasSlama, line 121: ❤️

packages/base/Tag/src/TagRectangular/TagRectangular.tsx
```diff
@@ -28,31 +20,38 @@ export const TagRectangular = forwardRef<HTMLDivElement, Props>(
...
+          </div>
+        )}
+
+        <Typography
+          size='xxsmall'
+          weight='semibold'
+          noWrap
+          className={twMerge(
+            'text-white mx-1 w-max',
+            variant === 'light-grey' ? 'text-graphite-800' : ''
```
TomasSlama, line 48:
```suggestion
            variant === 'light-grey' && 'text-graphite-800'
```
NIT: according twMerge documentation, we can use simple conditions

TomasSlama (approved):
I left two very small comments, but I will approve to unblock you as it is end of the day for me. Great job 🎉 💪 !

packages/base/Tag/package.json
```diff
@@ -27,15 +27,15 @@
...
     "**/styles.js"
   ],
   "peerDependencies": {
-    "@material-ui/core": "4.12.4",
     "@toptal/picasso-provider": "*",
+    "tailwindcss": ">=3",
```
sashuk, line 38: NIT: Can be replaced with `@toptal/picasso-tailwind` (https://toptal-core.slack.com/archives/GG3F4AS4T/p1715257124599019)

packages/base/Tag/src/Tag/Tag.tsx
sashuk:
```
    // `hovered` is extracted so we don't pass it to Root with {...rest}
    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

For me it seems like the fact that we do not pass `hover` to Root is quite obvious. However, the reason _why we do not pass it_ is still unclear, so the comment would benefit from explaining it, what do you think?

packages/base/Tag/src/Tag/Tag.tsx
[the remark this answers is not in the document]
ruslan-sed: @OleksandrNechai, so are we good with `rem`?

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
```diff
@@ -38,7 +44,14 @@ const TagCheckable = forwardRef<HTMLDivElement, Props>(function TagCheckable(
 
   return (
     <Tag
-      className={cx(className, classes.checkable)}
+      className={twMerge(
+        `hover:border-graphite-700 transition-all duration-[350ms] ease-in-out cursor-pointer`,
+        className,
+        checked && 'hover:border-red-500 hover:text-red-500',
+        hovered && 'border-graphite-700',
+        checked && hovered && 'border-red-500 text-red-500'
```
augustobmoura, line 52: I think we should move `className` to last, so it is possible to override all default styles from outside
```suggestion
        checked && 'hover:border-red-500 hover:text-red-500',
        hovered && 'border-graphite-700',
        checked && hovered && 'border-red-500 text-red-500'
        className,
```

packages/base/Tag/src/TagGroup/TagGroup.tsx
```diff
@@ -22,10 +14,15 @@ export const TagGroup = forwardRef<HTMLDivElement, Props>(function TagGroup(
...
 
-  const classes = useStyles()
-
   return (
-    <div {...rest} ref={ref} className={cx(classes.root, className)}>
+    <div
+      {...rest}
+      ref={ref}
+      className={twMerge(
+        'text-lg min-w-full transition-none -mr-2 -mb-2 [&>*]:mr-2 [&>*]:mb-2',
```
augustobmoura, line 22: Is there any difference between `mr-2` and `-mr-2`? Maybe it's a typo?
```suggestion
        'text-lg min-w-full transition-none mr-2 mb-2 [&>*]:mr-2 [&>*]:mb-2',
```

packages/base/Tag/src/TagRectangular/TagRectangular.tsx
```diff
@@ -28,31 +20,38 @@ export const TagRectangular = forwardRef<HTMLDivElement, Props>(
...
+        )}
+
+        <Typography
+          size='xxsmall'
+          weight='semibold'
+          noWrap
+          className={twMerge(
+            'text-white mx-1 w-max',
+            variant === 'light-grey' && 'text-graphite-800'
+          )}
```
augustobmoura, line 49: We can use `twJoin` or `classnames` here instead. We should avoid `twMerge` as much as possible. Each occurrence of it has a performance penalty
```suggestion
          className={twJoin(
            'text-white mx-1 w-max',
            variant === 'light-grey' && 'text-graphite-800'
          )}
```

packages/base/Tag/src/index.ts
```diff
@@ -4,5 +4,4 @@ export * from './TagCompound'
 export * from './TagConnection'
 export * from './TagGroup'
 export * from './TagRectangular'
-export * from './Chip'
```
augustobmoura, line 7: Is the removal of `Chip` expected? This would be a breaking change

packages/base/Tag/src/TagGroup/TagGroup.tsx
TomasSlama: @augustobmoura yeah, with `-` prefix we have negative value `margin-right: - 0.5rem`

packages/base/Tag/src/index.ts
TomasSlama: I double checked, there is not a single usage of Chip outside of Picasso https://github.com/search?q=org%3Atoptal+Chip&type=code&p=5, we are OK to kill it with 🔥. But yeah, we should mention this in the changeset just in case.

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
ruslan-sed: Not sure about this. In a case when someone passes for example `border-gree-600` as `className`, the border color will always be `green-600`, and there would not be a difference when the `hovered` or `checked` prop will be passed. 

On the other hand, a user can handle changing these styles himself. Are we ok with this?

packages/base/Tag/src/Tag/Tag.tsx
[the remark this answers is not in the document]
OleksandrNechai: wow, how could you find this comment? :) 
I am surprised to see these numbers and rems. 
What I expected to see is some class like `border-sm rounded-1`. I am not sure, is Tag that specific that it requires special treatment? 
Regarding REMS:
image
Do we rly support the changes of borders when people change their fonts? How would it look like?

To be clear, I am just questioning, to make sure that we know what we are doing ;) @TomasSlama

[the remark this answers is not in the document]
TomasSlama: I believe we should reuse existing token for this. `border` should be just enough. It was `1px` before, I think we should keep it.

image


Questions about when to use px/em/rem are asked every year and everybody has a different opinion about it. Let's not go into that conversation right now and let's just use it as it was used before for now. We can have RFC to come up with guidelines how to use this. But this question is not only for us, but also for designers. TBH I don't think it is that important right now, as our users are not elder people who updates font size of whole browser to read better. 16px is default value and we pretty much count on it.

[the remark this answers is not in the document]
TomasSlama: So I was testing it and we should be able to just have `border rounded-full`. But it will be available only after Dmitry merges the Button component to master. 
https://github.com/toptal/picasso/pull/4294/files?file-filters%5B%5D=.js&file-filters%5B%5D=.ts&show-deleted-files=true&show-viewed-files=true#diff-023efaa7d70715542f1de4ab4f7aa51647b761fb92e7dcb2293f68d3722f57c7R38

ruslan-sed:

ruslan-sed:
@toptal-devbot run package:alpha-release

packages/base/Tag/src/Indicator/Indicator.tsx
```diff
@@ -21,20 +19,21 @@ export interface Props extends BaseProps {
...
   const { className, color, style, ...restProps } = props
-  const classes = useStyles()
 
   return (
     <div
       role='img'
       {...restProps}
-      className={cx(classes.root, className, classes[color])}
+      className={twMerge(
+        'w-2 h-2 rounded-[50%]',
```
OleksandrNechai, line 33: can tokens be used instead of `rounded-[50%]`

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +36,102 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
   const handleDelete = (event: MouseEvent) => {
     if (disabled) {
       return
     }
 
     if (onDelete) {
       event.preventDefault()
+
+      /**
+       * This comment prevents the options list from opening.
```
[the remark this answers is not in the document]
OleksandrNechai: I see it is still here, just remove this line, its useless 😉

packages/base/Tag/src/TagGroup/TagGroup.tsx
TomasSlama: @OleksandrNechai noticed that Tailwind provides `space` util classes that we can maybe us to remove the `[&>*]`. Can you please check and try it out? https://tailwindcss.com/docs/space

packages/base/Tag/src/Indicator/Indicator.tsx
ruslan-sed: I don't see any predefined class with such value

packages/base/Tag/src/Tag/Tag.tsx
```diff
@@ -35,91 +35,109 @@ export interface Props extends BaseProps, TextLabelProps, DivOrAnchorProps {
...
     style,
     titleCase: propsTitleCase,
     variant = 'light-grey',
+    role,
+    // `hovered` is extracted so we don't pass it to Root with {...rest}
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    hovered,
     ...rest
   } = props
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
sashuk, line 63: Do we need this `eslint-disable-next-line` here? Can we just pass the `titleCase` from Tag props to Typography props?

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
```diff
@@ -38,7 +44,14 @@ const TagCheckable = forwardRef<HTMLDivElement, Props>(function TagCheckable(
...
+      )}
+      hovered={hovered}
```
sashuk, line 54: As discussed in DM, I propose to actually use this prop in `Tag` (not just ignore it) to achieve hovered styling (plus you will need to remove hovered styling from `TagCheckable`) – this proposal is aligned with "_if it is possible to style component with prop rather than Tailwind classes passed in `className`, style it with prop_" (cc @toptal/frontend-experience-eng)

packages/base/Tag/src/Tag/Tag.tsx
sashuk: Related proposal – https://github.com/toptal/picasso/pull/4264#discussion_r1596601575

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
ruslan-sed: It changes how `Tag` behaves. Do we want to change it in the migration ticket?

sashuk: @ruslan-sed Can you please check the https://github.com/toptal/picasso/blob/master/packages/base/Tag/src/Tag/Tag.tsx#L98? As far as I understand `Tag` was actually responsible for styling tags for "hovered" state, so we keep the API the same and do not move styling logic in TagCheckable (https://github.com/toptal/picasso/pull/4264/files#diff-0f640c533197883fe4d02c60678947c2b840bdd883ecf50bb6f3d54c8a32a26cR51)

packages/base/Tag/src/Tag/Tag.tsx
ruslan-sed: Yeah, good catch about `eslint-disable-next-line`, but passing `titleCase` directly from prop is not the same. `useTitleCase` has some logic.

sashuk: @ruslan-sed You are right, we get the app-wide setting for `titleCase` in that hook, we should keep it 👍

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
augustobmoura: We usually try to make the classnames more strict, in that sense it would be better to not let classNames override the borders.

But I noticed that before this change, className was actually: `className={cx(className, classes.checkable)}`
I believe that changing that order might have breaking changes consequences, can you check if the previous styles for `classes.checkable` are always at last position? In specific, `classes.checkable` has some pretty complex selectors, I'm not sure if we are supporting all of them here:
```js
'&$checkable:not($disabled)': {
        cursor: 'pointer',
        '&:hover, &$hovered': {
          borderColor: palette.grey.dark,
          backgroundColor: palette.common.white,
          transition: `all ${transitions.duration.short}ms ${transitions.easing.easeInOut}`,
        },
        '&:focus': {
          backgroundColor: palette.common.white,
        },
        '&$green': {
          '&:hover, &$hovered': {
            borderColor: palette.red.main,
            color: palette.red.main,
          },
        },
      },
      ```

packages/base/Tag/src/index.ts
augustobmoura: Got it, thanks for the explanation! Just wanted to make sure this wasn't accidental

packages/base/Tag/src/Indicator/Indicator.tsx
augustobmoura: There's `rounded-full`, but it is a [slightly different behavior](https://play.tailwindcss.com/ZuXnLZlr1P). I think we should stick to `50%`, the same as before the migration

packages/base/Tag/src/TagGroup/TagGroup.tsx
ruslan-sed: It can be done this way. `TagGroup` had this kind of styling with child elements before:
image

So we need `[&>*]` here.

packages/base/Tag/src/TagCheckable/TagCheckable.tsx
sashuk: As discussed, we decided to delete this property completely as it did not affect anything (reflected in changeset as well)

sashuk (approved):
Great job, thank you! All visual changes are related to font handling, the https://toptal-core.atlassian.net/browse/FX-5297 contains everything that is left to do

PR #4343, opened 2024-06-03
[FX-5302] Migrate NumberInput to TailwindCSS

ruslan-sed:
[FX-5302]

### Description

Migrate `NumberInput` to TailwindCSS.

### How to test

- [Temploy](https://picasso.toptal.net/fx-5302-migrate-numberinput)

### Development checks

- [x] Self reviewed

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-5302]: https://toptal-core.atlassian.net/browse/FX-5302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ruslan-sed:
@toptal-anvil ping reviewers

packages/base/NumberInput/src/NumberInput/__snapshots__/test.tsx.snap
```diff
@@ -5,11 +5,11 @@ exports[`NumberInput renders 1`] = `
   class="Picasso-root"
 >
   <div
-    class="MuiInputBase-root MuiOutlinedInput-root PicassoOutlinedInput-root PicassoNumberInput-root PicassoOutlinedInput-rootAuto PicassoOutlinedInput-rootMedium MuiInputBase-adornedEnd MuiOutlinedInput-adornedEnd"
+    class="MuiInputBase-root MuiOutlinedInput-root PicassoOutlinedInput-root pr-0 cursor-text PicassoOutlinedInput-rootAuto PicassoOutlinedInput-rootMedium MuiInputBase-adornedEnd MuiOutlinedInput-adornedEnd"
   >
     <input
       aria-invalid="false"
-      class="MuiInputBase-input MuiOutlinedInput-input PicassoOutlinedInput-input PicassoNumberInput-input PicassoOutlinedInput-inputMedium MuiInputBase-inputAdornedEnd MuiOutlinedInput-inputAdornedEnd"
+      class="MuiInputBase-input MuiOutlinedInput-input PicassoOutlinedInput-input [&::-webkit-outer-spin-button]:appearance [&::-webkit-outer-spin-button]:m [&::-webkit-inner-spin-button]:appearance [&::-webkit-inner-spin-button]:m PicassoOutlinedInput-inputMedium MuiInputBase-inputAdornedEnd MuiOutlinedInput-inputAdornedEnd"
```
augustobmoura, line 12: I believe appearance got borked here, because of the missing extra space I mentioned before `[&::-webkit-inner-spin-button]:appearance` doesn't make sense to me as a classname, same with `[&::-webkit-inner-spin-button]:m `

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
```diff
@@ -94,7 +92,7 @@ export const NumberInputEndAdornment = (props: Props) => {
 
     let nextValue = normalizedValue + normalizedStep
 
-    if (nextValue <= max) {
+    if (nextValue <= Number(max)) {
```
augustobmoura, line 95: Good catch! That was probably a bug that we never caught
Do you mind writing a unit test for it?

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
```diff
@@ -125,12 +123,15 @@ export const NumberInputEndAdornment = (props: Props) => {
...
-    },
-    classes[size],
-    classes.root
+  const rootClassName = twJoin(
+    `border-l border-r border-solid border-l-gray-400 border-r-transparent
+    hover:bg-gray-400 hover:border-gray-400 [&+*]:border-t [&+*]:border-solid 
+    [&+*]:border-t-gray-400 active:[&+*]:border-t active:[&+*]:border-t-solid 
+    active:[&+*]:border-gray-500 active:bg-gray-500 active:border-t-gray-500 
+    [&:first-child]:rounded-tr-sm [&:last-child]:rounded-br-sm 
+    transition-[color,_border,_background] ease-out duration-350`,
```
augustobmoura, line 132: I was thinking on grouping the styles by selector and line, mimicking the old `styles.ts` module, wdyt?
Also, we decided on replacing `&+&` JSS selectors with `[&:not(:first-of-type)]` which is more legible (see [discussion](https://toptal-core.slack.com/archives/GG3F4AS4T/p1711033742342919))
```suggestion
    `border-l border-r border-solid border-l-gray-400 border-r-transparent
    hover:bg-gray-400 hover:border-gray-400
    [&:not(:first-of-type)]:border-t [&:not(:first-of-type)]:border-solid [&:not(:first-of-type)]:border-t-gray-400 active:[&:not(:first-of-type)]:border-t
    active:[&:not(:first-of-type)]:border-t-solid active:[&:not(:first-of-type)]:border-gray-500
    active:bg-gray-500 active:border-t-gray-500 
    [&:first-child]:rounded-tr-sm
    [&:last-child]:rounded-br-sm 
    transition-[color,_border,_background] ease-out duration-350`,
```

packages/base/NumberInput/src/NumberInput/NumberInput.tsx
```diff
@@ -102,8 +93,13 @@ export const NumberInput = forwardRef<HTMLInputElement, Props>(
     return (
       <OutlinedInput
         classes={{
-          root: classes.root,
-          input: classes.input,
+          root: 'pr-0 cursor-text',
+          input:
+            // required to remove arrows in WebKit based browsers
+            `[&::-webkit-outer-spin-button]:appearance-none [&::-webkit-outer-spin-button]:m-0 
+            [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-inner-spin-button]:m-0 ` +
```
augustobmoura, line 100: NIT: We can remove the backticks here, as there is no interpolation
```suggestion
            '[&::-webkit-outer-spin-button]:appearance-none [&::-webkit-outer-spin-button]:m-0 
            [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-inner-spin-button]:m-0 ' +
```

augustobmoura (approved):
Left a few comments, but everything is working as expected! :+1:

packages/base/NumberInput/src/NumberInput/__snapshots__/test.tsx.snap
augustobmoura: You were faster than me and already fixed the code for it, we need to update the snapshots, though :sweat_smile:

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
```diff
@@ -1,12 +1,9 @@
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 import type { RefObject } from 'react'
 import React from 'react'
-import type { Theme } from '@material-ui/core/styles'
-import { makeStyles } from '@material-ui/core/styles'
 import type { BaseProps, SizeType } from '@toptal/picasso-shared'
 import { isBrowser } from '@toptal/picasso-shared'
 import { ButtonBase } from '@material-ui/core'
```
AdrianContiu, line 6: I was wondering if this button will remain here since we should have removed all the `material-ui` references...do you know if there is a task to replace it at some point?

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
ruslan-sed: Do you mean to extract handlers as utils and then test them?

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
TomasSlama: @augustobmoura I believe `[&:not(:first-of-type)]` was removed at the end during the merging to master as it was doing some problems in Client Portal

TomasSlama (commented):
The input adornment on master has `flex: 0 0 auto`, here we have `flex: 0 0 0`. It is related to already merged InputAdornment, but I think we can fix it here maybe, wdyt?

Also the `margin-left` is not the same on InputAdornment. When position is `end` it should have `ml-auto` but it has `ml-0`

image

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
augustobmoura: Not really necessary, just mount the element in a test if min and max are being applied

packages/base/NumberInput/src/NumberInput/NumberInput.tsx
augustobmoura: This is mostly a stylistic choice, both will work the same, but it is usually agreed in the community to only use backticks when you need an interpolation (`${variable}`) or breaking lines, otherwise, we should use single quotes (or double quotes, depending on the project, for Toptal projects we decided on single quotes). We should avoid concatenating strings with the `+` sign. There was a time when concatenating strings with plus sign was the preferred way, but with the advent of es2015, template literals became the norm.

For this case in specific, using the plus sign was justified because we had an explanation comment after the plus sign, this is not possible with template literals. The backticks are unnecessary however, because we don't have variable interpolations nor multiple lines. This is why I also marked the comment as [a `NIT`](https://stackoverflow.com/questions/27810522/what-does-nit-mean-in-hacker-speak)

ruslan-sed: Thank you for the explanation 👍

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
augustobmoura: > There is one small difference between `[&+&]` and `[&:not(:first-of-type)]`. It can bee seen in this example - https://play.tailwindcss.com/E70vdfwqkX.
> 
> `[&:not(:first-of-type)]` is affecting margin of element that is not placed immediately after same type of element. It will not affect the case of `NumberInputEndAdornment` but still it is not very precise. What do you think, @augustobmoura?

I think the initial intent is probably closer to `[&+&]` than `[&:not(:first-of-type)]`. But for this case where it doesn't really matter I would go with the more explicit version of `not first-of-type` rather than the cryptic `& + &`, seems more legible
 
> @augustobmoura I believe `[&:not(:first-of-type)]` was removed at the end during the merging to master as it was doing some problems in Client Portal

I'm still seeing some usages of `[&:not(:first-of-type)]` on the branch; maybe we didn't remove all of them? If their usage is problematic, we can keep it no problem

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
ruslan-sed: It does not feel right to test this. This is not related to component functionality, it is just a TypeScript issue. I believe we should not add component tests in such cases. 

In the newest versions of TypeScript it doesn't compile in such cases. Here is a [discussion](https://github.com/microsoft/TypeScript/issues/55808). And, at the bottom of the discussion, there is a merged PR that fixes this issue. 

Maybe we need one more pair of eyes here. @TomasSlama, what do you think?

TomasSlama (approved):

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
ruslan-sed: After discussion with @TomasSlama decided to use `&+&` here as it reflects what we had before migration.

packages/base/NumberInput/src/NumberInputEndAdornment/NumberInputEndAdornment.tsx
TomasSlama: @augusto it is actually not solving any bug, it behaves the same. `1 > 'foo'` returns false the same way as `1 > Number('foo')`, so the Number here is only to satisfy Typescript. I was expecting error TBH, but now I'm also not sure what can we actually test

TomasSlama: @augusto hups, I meant @augustobmoura

augustobmoura: > It does not feel right to test this. This is not related to component functionality, it is just a TypeScript issue. I believe we should not add component tests in such cases.

I was talking about the possibility of comparing numbers to strings in runtime, as @TomasSlama caught, not the Typescript issue itself

> it is actually not solving any bug, it behaves the same. `1 > 'foo'` returns false the same way as `1 > Number('foo')`, so the Number here is only to satisfy Typescript. I was expecting error TBH, but now I'm also not sure what can we actually test

I guess you are right @TomasSlama, the code should work the same (or at least close enough) as before because of JS implicit conversions from string to number, the tests are still a good improvement, but not necessary. I think the test cases that we should test are more likely when string is not a number and check if the component is correctly rendered and ignores values that are not numbers.

BTW for the curiosity this is the official specification for the behavior of `<=`: https://tc39.es/ecma262/#sec-islessthan

Spec commented:
image

ruslan-sed: All right, @augustobmoura now I got what you meant. That makes perfect sense. Now component don't ingore `min` and `max` when it can't be converted to a number. So only adding a test is not enough. 

I created a [ticket](https://toptal-core.atlassian.net/browse/FX-5484) to cover this case, so we can handle it in the future.

PR #4344, opened 2024-06-03
[FX-4610] Migrate CheckboxGroup to TailwindCSS 

AdrianContiu:
[FX-4610]

### Description

This PR removes `material-ui` styles and replace them with Tailwind CSS classes. The dependencies will be cleaned in a later task, after the migration of the `Checkbox`.

### How to test

- [Temploy](https://picasso.toptal.net/fx-4610-migrate-checkboxgroup)
- Check the [production](https://picasso.toptal.net/?path=/story/forms-checkbox--checkbox) and compare it with the temploy.

### Screenshots

### Development checks

- ~Read [CONTRIBUTING.md](https://github.com/toptal/picasso/blob/master/CONTRIBUTING.md) and [Component API principles](https://github.com/toptal/picasso/blob/master/docs/contribution/component-api.md)~
- ~Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).~
- [x] Self reviewed
- ~Covered with tests ([visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md) included)~

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-4610]: https://toptal-core.atlassian.net/browse/FX-4610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,44 +1,35 @@
...
-
 const CheckboxGroup = (props: Props) => {
   const { horizontal, spacing, xs, sm, md, lg, xl, className, ...rest } = props
 
-  const classes = useStyles()
-  const { spacing: themeSpacing } = useTheme()
-
   const direction = horizontal ? 'row' : 'column'
-  const gridSpacing = spacing ?? horizontal ? themeSpacing(2) : 0
+  const gridSpacing = spacing ?? horizontal ? 16 : 0
```
augustobmoura, line 21: It is a good practice to name constants, and avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))
```suggestion
  const gridSpacing = spacing ?? horizontal ? HORIZONTAL_SPACING : 0
```

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,44 +1,35 @@
...
 
   return (
-    <FormGroup
+    <div
       {...rest}
-      classes={classes}
-      className={cx(classes.root, className)}
+      className={twMerge(
+        'flex flex-col flex-wrap -mr-2 -mb-2',
+        horizontal ? 'flex-row' : undefined,
```
augustobmoura, line 30: NIT: We can simplify this line further
```suggestion
        horizontal && 'flex-row',
```

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -48,7 +39,7 @@ const CheckboxGroup = (props: Props) => {
         {children.map((child, index) => (
           <Grid.Item
             key={index}
-            className='leading-none [&&]:pt-0 [&&]:pb-0'
+            className='leading-none [&&]:pt-0 [&&]:pb-0 [&&>label]:mb-[0.5em]'
```
augustobmoura, line 42: What does `&&>` means? Can it maybe be simplified to `&>`?

augustobmoura (commented):
Left a few comments and updated the branch to the feature branch, it should solve the problem with unrelated Happo changes

AdrianContiu:
> Left a few comments and updated the branch to the feature branch, it should solve the problem with unrelated Happo changes

Thanks! There are still some Happo tests in StoryBook that are failing. It seems to be a small spacing difference, but I couldn't figure out how to fix it.

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -1,44 +1,37 @@
...
 
   return (
-    <FormGroup
+    <div
       {...rest}
-      classes={classes}
-      className={cx(classes.root, className)}
+      className={twMerge(
+        'flex flex-col flex-wrap -mr-[0.5em] -mb-[0.5em]',
+        horizontal && 'flex-row',
```
ruslan-sed, line 32: I don't see classes for `horizontal`. I assume this is due to Tomas's ticket for refactoring horizontal layout. It mean you don't need it anymore.

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -48,7 +41,7 @@ const CheckboxGroup = (props: Props) => {
         {children.map((child, index) => (
           <Grid.Item
             key={index}
-            className='leading-none [&&]:pt-0 [&&]:pb-0'
+            className='leading-none pt-0 pb-0 [&>label]:mb-[0.5em]'
```
ruslan-sed, line 44: `[&>label]:mb-[0.5em]` also looks like it is out of nowhere. Assume this is why you have spacing difference.

ruslan-sed (approved):
Looks nice 👍  
Few small things need to be addressed before merge.

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
ruslan-sed: Here is the same.

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
ruslan-sed: But wait. Comparing to the old code it also appears that there was no `horizonta` thingy.

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
ruslan-sed: I think that difference is because not all descendant labels should have bottom margin. In old code only those that had specific class had it.

AdrianContiu: Yes, that was the reason for the Happo test failures. Luckily, Augusto paired up and helped me acknowledge that. Thank you as well!

packages/base/Checkbox/src/Checkbox/__snapshots__/test.tsx.snap
```diff
@@ -12,7 +12,7 @@ exports[`Checkbox Checkbox.Group renders checkbox in a grid group 1`] = `
         class="box-border flex flex-row flex-wrap items-start justify-start w-[calc(100%+16px)] -m mt-0 mb-0"
       >
         <div
-          class="box-border m-0 p-[8px] sm:basis-4/12 sm:max-w leading-none pt-0 pb-0 [&>label]:mb-[0.5em]"
+          class="box-border m-0 p-[8px] sm:basis-4/12 sm:max-w leading-none pt-0 pb-0 [&>.picasso-checkbox]:mb-[0.5em]"
```
ruslan-sed, line 15: I have a question about the solution that you used. 

1. So now you are using `&>.picasso-checkbox`. Does that mean, that `picasso-checbox` class will be there after `Checkbox` is migrated?

2. And this one is for a learning purpose. So this pattern with adding some class for a descendant element can be used in such cases? @augustobmoura

AdrianContiu: > I have a question about the solution that you used.
> 
> 1. So now you are using `&>.picasso-checkbox`. Does that mean, that `picasso-checbox` class will be there after `Checkbox` is migrated?
> 2. And this one is for a learning purpose. So this pattern with adding some class for a descendant element can be used in such cases? @augustobmoura

1. I think so... as I see it, this is a default className in `FormControlLabel`, so it most likely stays that way even after migration.

packages/base/Checkbox/src/CheckboxGroup/CheckboxGroup.tsx
AdrianContiu: This comes from [`material-ui`](https://github.com/mui/material-ui/blob/v4.x/packages/material-ui/src/FormGroup/FormGroup.js) because the `div` replaced the `FormGroup` and this was one of its styles.

AdrianContiu:
> > > Left a few comments and updated the branch to the feature branch, it should solve the problem with unrelated Happo changes
> > 
> > 
> > Thanks! There are still some Happo tests in StoryBook that are failing. It seems to be a small spacing difference, but I couldn't figure out how to fix it.
> 
> We need to fix this difference before merging to the feature branch. It seems like a change in the bottom part of `CheckboxGroup`, or related to line-height
> 
> image

Done! Thank you for your help!

packages/base/Checkbox/src/Checkbox/__snapshots__/test.tsx.snap
augustobmoura: > So now you are using &>.picasso-checkbox. Does that mean, that picasso-checbox class will be there after Checkbox is migrated?

`picasso-checkbox` is [added manually to the `Checkbox`](https://github.com/toptal/picasso/blob/cb1c05a59846cd1c6dd0fe3b1f9705a18d08232b/packages/base/Checkbox/src/Checkbox/Checkbox.tsx#L122) component, we shouldn't expect it to be present automatically.

> And this one is for a learning purpose. So this pattern with adding some class for a descendant element can be used in such cases?

Although it is possible to do it, we should avoid this type of cross-component styling. The main concern is that the connection is implicit and difficult to track; the class name would need to exist on the checkbox implicitly. A better approach would be to leave the customization to a special prop or use `className` props. There are a few occasions where this approach is better suited, but it is not a common practice within Picasso. Now thinking of it, I see a few places where we could simplify the code by using it, for keeping everything tighter unit tests are crucial, I believe.

I believe it is justified for this specific case, because we are just mimicking the [original implementation](https://github.com/toptal/picasso/blob/cb1c05a59846cd1c6dd0fe3b1f9705a18d08232b/packages/base/Checkbox/src/CheckboxGroup/styles.ts#L9)

augustobmoura (approved):
Checked Happo and storybook and everything is working as expected! Nice work! :+1:

PR #4409, opened 2024-07-15
[FX-5525] Migrate Container to TailwindCSS

AdrianContiu:
[FX-5525]

### Description

This PR removes material-ui styles and replaces them with Tailwind CSS classes.

### How to test

- [Temploy](https://picasso.toptal.net/fx-5525-container)
- Check the [production](https://picasso.toptal.net/?path=/story/layout-container--container) and compare it with the temploy.

### Screenshots

| image | image |

### Development checks

- ~Read [CONTRIBUTING.md](https://github.com/toptal/picasso/blob/master/CONTRIBUTING.md) and [Component API principles](https://github.com/toptal/picasso/blob/master/docs/contribution/component-api.md)~
- ~Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).~
- [x] Self reviewed

**Breaking change**

- ~codemod is created and showcased in the changeset
- ~test alpha package of Picasso in StaffPortal

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-5525]: https://toptal-core.atlassian.net/browse/FX-5525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AdrianContiu:
@toptal-anvil ping reviewers

packages/base/Container/src/Container/Container.tsx
```diff
@@ -118,62 +97,39 @@ export const Container: ContainerProps = documentable(
...
+            variant && variantClassesByColor[variant],
+            getSpacingClasses(spacingProps),
+
+            typeof align === 'string' && alignmentClasses.textAlign[align],
+
+            alignItems && alignmentClasses.alignItems[alignItems],
+
+            justifyContent && alignmentClasses.justifyContent[justifyContent],
+
+            bordered && 'border-DEFAULT border-solid border-gray-200',
```
TomasSlama, line 119:
```suggestion
            bordered && 'border border-solid border-gray-200',
```
we don't need to use the DEFAULT

mkrl, line 119:
```suggestion
            bordered && 'border border-solid border-gray-200',
```
I believe the default border with no value should already provide the same outcome. `border-DEFAULT` is not a standard Tailwind directive

packages/base/Container/src/Container/Container.tsx
```diff
@@ -118,62 +97,39 @@ export const Container: ContainerProps = documentable(
...
+            rounded && 'rounded-md',
+            inline && 'inline-block',
+            flex ? (inline ? 'inline-flex' : 'flex') : '',
```
TomasSlama, line 122: these two lines are related and both take care of `display` CSS value. Let's move it to fn `getDisplayValue`. I believe we should try to not overuse tailwind-merge when not necessary

packages/base/Container/src/Container/utils/constants.ts
```diff
@@ -0,0 +1,546 @@
...
+export const SPACING_CLASSES = {
+  0: {
+    gap: {
+      default: 'gap-0',
+      xs: 'xs:gap-0',
+      sm: 'sm:gap-0',
+      md: 'md:gap-0',
+      lg: 'lg:gap-0',
+      xl: 'xl:gap-0',
+    },
```
mkrl, line 11: We get a lot of repetition in this one big object. Do you think it could be a function that returns the values needed based on parameters passed?
We get a 1:1 mapping with the spacing index (for example, `0`) and all the values inside it (like `gap: { xs: 'xs:-gap-0' }`). Plus, lots of other values repeated over and over again.

I would usually prefer a big mapping object, but something this big it a nice opportunity to save on space.

You could probably even keep the `SPACING_CLASSES` object itself, just initialize it dynamically with something like this:
```ts
const SPACING_CLASSES = createSpacingClasses(12)
```

packages/base/Container/src/Container/utils/get-spacing-classes.ts
```diff
@@ -0,0 +1,133 @@
...
+  ]
+}
+
+const convertToRem = (value?: SpacingType) => {
+  if (!value || typeof value !== 'number') {
+    return
+  }
+
+  return `${value}rem`
+}
```
mkrl, line 105:
```suggestion
const convertToRem = (value?: SpacingType) => {
  if (typeof value !== 'number') {
    return
  }

  return `${value}rem`
}
```

Very minor nit, we can probably get rid of the extra assertion here.

packages/picasso-provider/src/Picasso/config/spacings.ts
```diff
@@ -36,26 +36,31 @@ export enum SpacingEnum {
...
 class PicassoSpacing {
   #value: PicassoSpacingValues
   /**
    * Corresponds to token index in https://toptal-core.atlassian.net/wiki/spaces/Base/pages/3217031216/Spacing
    */
-  baseTokenIndex: number
+  baseTokenIndex: spacingToken
 
-  private constructor(value: PicassoSpacingValues, index: number) {
+  private constructor(value: PicassoSpacingValues, index: spacingToken) {
```
mkrl, line 48: I would suggest testing the new version of picasso-provider somewhere we use the new spacings (probably client-portal)? 
I'm not 100% certain it will not break type checking if someone used invalid values before. 

On the other hand, using invalid values could be considered a user error :)

packages/base/Container/src/Container/utils/constants.ts
TomasSlama: @mkrl @AdrianContiu we can't do this. We can't have dynamic classes with TailwindCSS

packages/base/Container/src/Container/Container.tsx
```diff
@@ -118,62 +97,51 @@ export const Container: ContainerProps = documentable(
...
-
-      const baseSpacingClasses = getBaseSpacingClasses(
-        { top, left, bottom, right, gap, padded },
-        classes
-      )
+      const spacingProps = { gap, padded, top, bottom, right, left }
+
+      const getDisplayValue = (
+        type: boolean | undefined,
+        display: boolean | undefined
```
TomasSlama, line 107:
```suggestion
        isInline: boolean | undefined,
        isFlex: boolean | undefined
```

TomasSlama (approved):
I agree with Misha that we should test it in some portal. Even Staff Portal is using it.

AdrianContiu:

packages/base/Container/src/Container/utils/constants.ts
mkrl: That's true. WDYT about using the workaround with class names included in code comments?

TomasSlama: If we would have to write all of those classes to the comment anyway, I would prefer it this way as we have them at least in organized structure, so if we would have to add new one, it would be simpler and we would not forget to add them to the comment. As it is in separate file and it is simple constant lookup table, it should be even more performant 🤔

AdrianContiu:

AdrianContiu:

PR #4431, opened 2024-07-31
[FX-5523] Migrate Calendar to Tailwind

sashuk:
[FX-5523]

### Description

This pull request migrates [Calendar](https://picasso.toptal.net/?path=/story/forms-datepicker--datepicker) component to TailwindCSS.

### How to test

- [Temploy](https://picasso.toptal.net/fx-5523-calendar/?path=/story/forms-datepicker--datepicker)
- Compare with `master`, use https://chromewebstore.google.com/detail/synchronize-tab-scrolling/phceoocamipnafpgnchbfhkdlbleeafc for easier tab scrolling 
- Compare implementation with https://github.com/toptal/picasso/blob/master/packages/base/Calendar

### Development checks

- [N/A] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core
- [N/A] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md).
- [N/A] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-5523]: https://toptal-core.atlassian.net/browse/FX-5523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/base/Calendar/src/CalendarDay/CalendarDay.tsx
```diff
@@ -63,7 +61,36 @@ const CalendarDay = (dayProps: DayProps): JSX.Element => {
...
-  const classes = useStyles()
+  const stateDependentClasses = useMemo(() => {
+    const currentComponentState = {
+      isSelected,
+      isDisabled,
+      isWeekend,
+      isToday,
+      isOutside,
+      isRangeStart,
+      isRangeMiddle: isRangeMiddle || isTemporaryRangeMiddle,
```
sashuk, line 72: `isTemporaryRangeMiddle` and `isTemporaryRangeEnd` trigger the same behaviour as `isRangeMiddle` and `isRangeEnd`, so `DayState` type does not contain `isTemporaryRangeMiddle` and `isTemporaryRangeEnd`

packages/base/Calendar/src/CalendarDay/CalendarDay.tsx
```diff
@@ -74,15 +103,15 @@ const CalendarDay = (dayProps: DayProps): JSX.Element => {
...
-        [classes.grayed]: isOutside && !isDisabled,
-        [classes.disabled]: isDisabled,
-        [classes.startSelection]: isRangeStart,
-        [classes.withinSelection]: isRangeMiddle || isTemporaryRangeMiddle,
-        [classes.endSelection]: isRangeEnd || isTemporaryRangeEnd,
-      })}
+      className={twJoin(
+        'h-10 w-10 min-w-10',
+        'm-0 p-0',
+        'text-[0.75rem]',
```
sashuk, line 109: The `text-xxs` sets both font size and line height which looks different from before. Setting the desired line height explicitly does not work due to calculation of line height (there is a tiny difference in visuals in this case due to miscalculated line height)

Screenshot 2024-08-05 at 12 41 48

sashuk, line 109: The `text-xxs` sets both font size and line height which looks different from before.

Setting the desired line height explicitly (via `leading-normal` or `leading-[1.5em]`) does not work due to calculation of line height, there is still a tiny difference in visuals in this case due to miscalculated line height (please see the demo)


https://github.com/user-attachments/assets/bd7c41a4-bd83-4783-877a-27aab36b2b45

sashuk:
@toptal-anvil ping reviewers

packages/base/Calendar/src/CalendarDay/utils/get-hover-and-focus-effects-classes.ts
```diff
@@ -0,0 +1,55 @@
...
+    return ''
+  }
+
+  if (isRangeEnd) {
+    // For Tailwind to pick classes up
+    // [&]:hover:text-white [&]:hover:bg-blue-500
+    // [&]:focus:text-white [&]:focus:bg-blue-500
+    return `${applyToHoverAndFocus('text-white')} ${applyToHoverAndFocus(
+      'bg-blue-500'
+    )}`
```
augustobmoura, line 26: If we are writing the classes on comments, we are now duplicating their value, both on the string value and the comment itself.
What is the benefit of this approach over just having the classname directly on the value?

Also note that because the comment is implicitly linked to the function implementation we need to keep the two in sync, if we change the function's implementation we need to remember to also change the comment where the function is being called. I would prefer to just have the class names inside the string constants
```suggestion
    return '[&]:hover:text-white [&]:hover:bg-blue-500 [&]:focus:text-white [&]:focus:bg-blue-500'
```

augustobmoura (approved):

augustobmoura:
Left a few comments that need to be addressed before merging

.changeset/sharp-clouds-thank.md
```diff
@@ -0,0 +1,6 @@
+---
+'@toptal/picasso-calendar': major
+---
+
+- migrate component to TailwindCSS
+- update peer dependencies
```
TomasSlama, line 6:
```suggestion
---
'@toptal/picasso-calendar': major
'@toptal/picasso': patch
---

### Calendar

- migrate component to TailwindCSS
- remove peer dependency on `@material-ui/core`
- add peer dependency on `@toptal/picasso-tailwind-merge`
```

This way it will also appear on release notes for the whole Picasso package and users will not have to go into the PR to understand what peer dependencies were actually updated. WDYT?

packages/base/Calendar/src/CalendarDay/CalendarDay.tsx
TomasSlama: did you try `text-xss/[1.5em]` or it will be the same problem?

packages/base/Calendar/src/CalendarDay/CalendarDay.tsx
```diff
@@ -74,15 +103,15 @@ const CalendarDay = (dayProps: DayProps): JSX.Element => {
...
+        'flex items-center justify-center vertical-align-middle relative',
+        'user-select-none',
+        'outline-none',
+        stateDependentClasses
```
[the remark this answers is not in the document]
TomasSlama: it seems there are no overrides, looks very nice ❤️

packages/base/Calendar/src/CalendarDay/types.ts
```diff
@@ -0,0 +1,10 @@
+export type DayState = {
+  isSelected: boolean
+  isDisabled: boolean
+  isWeekend: boolean
+  isToday: boolean
+  isOutside: boolean
+  isRangeStart: boolean
+  isRangeMiddle: boolean
+  isRangeEnd: boolean
```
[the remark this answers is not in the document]
TomasSlama: if we follow clean code I believe there should be `is` prefix, as when you read the code you don't even need to check the type

packages/base/Calendar/src/CalendarDay/utils/get-hover-and-focus-effects-classes.ts
TomasSlama: good point ☝️ also for better readability it is nice to use array

```
const classes = [
  '[&]:hover:text-white',
  '[&]:hover:bg-blue-500',

  '[&]:focus:text-white',
  '[&]:focus:bg-blue-500',
]
```

as when there is the same prefix on each line your eyes will start ignoring it and you can better see the important values

packages/base/Calendar/src/CalendarDay/CalendarDay.tsx
sashuk: @TomasSlama Yep, it is the same somehow

https://github.com/user-attachments/assets/bb53af13-b0d4-460b-9540-804571170602

packages/base/Calendar/package.json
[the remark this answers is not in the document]
sashuk: @augustobmoura Deleted, thank you!

packages/base/Calendar/src/CalendarContainer/CalendarContainer.tsx
```diff
@@ -24,17 +17,17 @@ const CalendarContainer = ({
...
     <div
-      className={cx(classes.root, {
-        [classes.hasFooter]: hasFooter,
-        [classes.flexible]: isFlexible,
-      })}
+      className={twJoin(
+        'p-6 text-graphite-800 flex flex-col basis-[20.5rem] shadow-5 bg-white',
+        hasFooter ? 'rounded-t-sm rounded-b-none' : 'rounded-sm',
+        !isFlexible && 'max-w-[20.5rem]'
+      )}
```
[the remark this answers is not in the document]
sashuk: @augustobmoura Thank you!

packages/base/Calendar/src/CalendarDay/utils/get-hover-and-focus-effects-classes.ts
sashuk: @augustobmoura @TomasSlama Agree, thank you guys! The change was reverted, now we deal with arrays which is slightly less readable, but instead we do not have to keep comments in sync (as they do not exist)

.changeset/sharp-clouds-thank.md
sashuk: @TomasSlama Agree, addressed, thank you!

sashuk:
@toptal-anvil ping reviewers

packages/base/Calendar/src/CalendarDay/utils/get-hover-and-focus-effects-classes.test.ts
```diff
@@ -0,0 +1,72 @@
+import { getHoverAndFocusEffectsClasses } from './get-hover-and-focus-effects-classes'
+import type { DayState } from '../types'
+
+describe('getHoverAndFocusEffectsClasses', () => {
+  describe('when day is not a weekend', () => {
+    it('returns default classes', () => {
+      const state = {} as DayState
+
+      expect(getHoverAndFocusEffectsClasses(state).toString()).toBe(
+        '[&]:hover:text-black [&]:focus:text-black,[&]:hover:bg-blue-500/25 [&]:focus:bg-blue-500/25,[&]:hover:rounded-sm [&]:focus:rounded-sm'
```
TomasSlama, line 10: did you consider to move these strings to variables, so we can reuse the variable in both code and in the test?. This way when we update any part of the styles we also need to update the test.

I was using similar approach in [OutlinedInput](https://github.com/toptal/picasso/blob/master/packages/base/OutlinedInput/src/OutlinedInput/stylesInput.test.ts)

TomasSlama: WDYT?

sashuk: @TomasSlama Hm, if we mess up CSS in one of the variables, then tests will stay green, right?

TomasSlama: for styles itself we have visual tests, right? But we should try to write unit tests in a way that we don't need to update it with every single change or am I wrong? 🤔

TomasSlama: I will approve as it is common to write tests in this way, this more like topic for discussion

TomasSlama (approved):

PR #4472, opened 2024-08-14
[FX-5536] Migrate FormAutoSaveIndicator to TailwindCSS

ruslan-sed:
[FX-5536]

### Description

Migrate FormAutoSaveIndicator to TailwindCSS.

**Note**

Jest tests were replaced with Cypress tests because Jest test for `when saved > should show label` was failing. The reason is the following - `.toBeVisible()` was failing to recognize that the element is hidden when it has the class `invisible`. The class was there, double-checked with `screen.debug()`. 

This issue is discussed [here](https://stackoverflow.com/questions/52813527/cannot-check-expectelm-not-tobevisible-for-semantic-ui-react-component) and it looks like Jest is unable to load CSS classes from Tailwind. 

### How to test

- [Temploy](https://picasso.toptal.net/fx-5537-migrate-form-auto-save-indicator)

### Development checks

- ~Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core~
- [x] Self reviewed

**Breaking change**

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-5536]: https://toptal-core.atlassian.net/browse/FX-5536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/base/Form/src/FormAutoSaveIndicator/FormAutoSaveIndicator.tsx
```diff
@@ -60,12 +52,7 @@ const FormAutoSaveIndicator = ({
   }, [savingState, hideTimeout])
 
   return (
-    <Container
-      className={cx(classes.root, {
-        [classes.visible]: savingState === SavingState.Saved,
-      })}
-      align='right'
-    >
+    <Container className={getStyles({ savingState })} align='right'>
```
dmaklygin, line 55: Could we memoize result of `getStyles` using useMemo?

packages/base/Form/src/FormAutoSaveIndicator/styles.ts
```diff
@@ -1,13 +1,15 @@
...
-      marginTop: '0.25em',
-      marginBottom: '0.25em',
-    },
-    visible: {
-      visibility: 'visible',
-    },
-  })
+import { twJoin } from '@toptal/picasso-tailwind-merge'
+
+import { SavingState } from './FormAutoSaveIndicator'
```
dmaklygin, line 3: I wouldn't depend on `SavingState` here, first of all it creates a circular dependency between files, and also brings a complexity to this style file.
I would specify a simple argument of `getStyles` function, let say:
```
export const getStyles = ({ visible }: { visible: boolean })
```

and generate styles according to `visible` argument which is boolean

dmaklygin: Then, function call would be the following:

```
getStyles({ visible: savingState === SavingState.Saved })
```

so, we won't be needed extra function here (`getVisibilityClass`)

ruslan-sed: Circular dependency issue can be solved by moving types to `types.ts` or something like that. 

About `getStyles({ visible: savingState === SavingState.Saved })`. In other cases, as for example [here](https://github.com/toptal/picasso/blob/142a4320343605959d90a3a8c68a14ad8fedb356/packages/base/OutlinedInput/src/OutlinedInput/OutlinedInput.tsx#L153) we decided to have all logic of deciding what style should be used to style helper function. So in this case I want to be consistent. It makes the function more complex, but it simplifies the component.

Lets ask @TomasSlama about this, since it was he who came up with this approach. What do you think, @TomasSlama?

TomasSlama: We use helper functions when there are many styles and it helps with readability and reduces complexity. In this case when we have only 3 lines styles I think it is overkill as it is not complex and very easy to understand.

ruslan-sed: All right, then I'll remove those style helpers

TomasSlama (approved):

ruslan-sed:
@toptal-anvil ping reviewers

ruslan-sed:
@toptal-anvil ping reviewers

packages/base/Form/src/FormAutoSaveIndicator/test.tsx
```diff
@@ -1,32 +0,0 @@
...
-  describe('when saved', () => {
-    it('should show label', () => {
-      const { queryByText, rerender } = render(<FormAutoSaveIndicator saving />)
-
-      rerender(<FormAutoSaveIndicator />)
-
-      expect(queryByText('Saved')).toBeVisible()
-    })
-  })
-})
```
dmaklygin, line 32: I would not remove it, but just a keep one test case which takes a snapshot. 
Snapshots are still useful, they helped us to determine difference in the HTML during Picasso migration.
While screenshots help to determine visual difference, snaphosts may catch internal changes

dmaklygin (approved):

PR #4534, opened 2024-09-06
[FX-5837] Disable autocomplete in tag selector

sashuk:
[FX-5837]

### Description

This pull request does two closely related things:

- fixes bug with thin popper container appearing when there are no options and no "empty options" text in `Autocomplete`. Use the code below to observe it in https://picasso.toptal.net/

Screenshot 2024-09-09 at 10 57 05

<details>
  <summary>Reproduction code</summary>

```
import React, { useState } from 'react'
import type { AutocompleteItem } from '@toptal/picasso'
import { Autocomplete, Form } from '@toptal/picasso'
import { isSubstring } from '@toptal/picasso-utils'

const Example = () => {
  const [value, setValue] = useState('')
  const [options, setOptions] = useState<AutocompleteItem[] | null>([])

  return (
    <div>
      <Form autoComplete='off'>
        <Autocomplete
          placeholder='Start typing country...'
          value={value}
          options={[]}
          noOptionsText={null}
          onChange={newValue => {
            console.log('onChange returns just item value:', newValue)

            setOptions(filterOptions(newValue))
            setValue(newValue)
          }}
          data-testid='autocomplete'
        />
      </Form>
    </div>
  )
}

export default Example
```
</details>

- adds support for non-dropdown option submission in `TagSelector` component (it heavily relies on `Autocomplete`) via `submitOtherOptionOnEnter` prop. To achieve the behavior from the [original thread](https://toptal-core.slack.com/archives/CCC3GP6CC/p1724313176283519), consumers need to specify certain combination of other properties (dedicated story was added).

### How to test

- Check `Other option submission on Enter` story at [temploy](https://picasso.toptal.net/fx-5837-disable-autocomplete-in-tag-selector/?path=/story/forms-tagselector--tagselector)
- Client Portal tests in https://github.com/toptal/client-portal/pull/9733 are fine

### Screenshots

https://github.com/user-attachments/assets/fc676d4c-ee4d-47b9-a0d0-844473fe51da

### Development checks

- [N/A] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`)
- [x] Self reviewed

<details>
<br />

List of available commands:

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FX-5837]: https://toptal-core.atlassian.net/browse/FX-5837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/three-coats-press.md
+++ b/.changeset/three-coats-press.md
@@ -1,0 +1,7 @@
+---
+'@toptal/picasso-tagselector': minor
+---
+
+### TagSelector
+
+- feature: use `submitOtherOptionOnEnter` property to submit other option on Enter key press
--- a/cypress/component/TagSelector.spec.tsx
+++ b/cypress/component/TagSelector.spec.tsx
@@ -47,6 +47,7 @@ const TagSelectorExample = ({
     <Container padded='medium'>
       <TagSelector
         {...props}
+        getKey={item => item.code as string}
         placeholder='Start typing...'
         options={filteredOptions}
         getDisplayValue={getDisplayValue}
--- a/packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
+++ b/packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
@@ -224,7 +224,10 @@ export const Autocomplete = forwardRef<HTMLInputElement, Props>(
 
     const optionsLength = options ? options.length : 0
 
-    const optionsMenu = options && (
+    console.log('@@@ options', options)
+
+    const hideOptionsCompletely = optionsLength === 0 && !noOptionsText
+    const optionsMenu = !hideOptionsCompletely && options && (
       <SelectOptions
         data-testid={testIds?.scrollMenu}
         selectedIndex={highlightedIndex}
--- a/packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
+++ b/packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
@@ -17,25 +17,30 @@ import type { PopperOptions } from 'popper.js'
 import { Input } from '@toptal/picasso-input'
 import { Container } from '@toptal/picasso-container'
 import { Loader } from '@toptal/picasso-loader'
-import { SelectOptions } from '@toptal/picasso-select'
 import { Popper } from '@toptal/picasso-popper'
 import { InputAdornment } from '@toptal/picasso-input-adornment'
 import { unsafeErrorLog } from '@toptal/picasso-utils'
 import type { InputProps } from '@toptal/picasso-input'
-import { MenuItem } from '@toptal/picasso-menu'
 import type { BaseInputProps, Status } from '@toptal/picasso-outlined-input'
 import { useFieldsLayoutContext } from '@toptal/picasso-form'
 import { twMerge } from '@toptal/picasso-tailwind-merge'
 
-import PoweredByGoogle from './PoweredByGoogle'
-import NoOptionsMenuItem from './NoOptionsMenuItem'
-import OtherOptionMenuItem from './OtherOptionMenuItem'
 import type { Item, ChangedOptions } from './types'
 import { useAutocomplete, EMPTY_INPUT_VALUE } from './use-autocomplete'
 import { rootClassByWidth } from './styles'
+import type { OptionsMenuProps } from './OptionsMenu'
+import OptionsMenu from './OptionsMenu'
 
 export interface Props
   extends BaseProps,
+    Pick<
+      OptionsMenuProps,
+      | 'otherOptionText'
+      | 'renderOtherOption'
+      | 'noOptionsText'
+      | 'renderOption'
+      | 'poweredByGoogle'
+    >,
     Omit<
       InputHTMLAttributes<HTMLInputElement>,
       'defaultValue' | 'value' | 'onChange' | 'onSelect' | 'onKeyDown' | 'size'
@@ -57,10 +62,6 @@ export interface Props
   ) => void
   /** Placeholder for value */
   placeholder?: string
-  /** Text prefix for other option */
-  otherOptionText?: string
-  /** Callback responsible for rendering the other option given the input's value */
-  renderOtherOption?: (value: string) => ReactNode
   /** Width of the component */
   width?: 'full' | 'shrink' | 'auto'
   /** Width of the menu */
@@ -69,12 +70,10 @@ export interface Props
   loading?: boolean
   /** Allow to show the other option in the list of options */
   showOtherOption?: boolean
-  /** Label to show when no options were found */
-  noOptionsText?: string | null
   /** List of options */
   options?: Item[] | null
   /** A function that takes a display value from the option item */
-  getDisplayValue?: (item: Item | null) => string
+  getDisplayValue?: OptionsMenuProps['getDisplayValue']
   /**  Callback invoked when key is pressed */
   onKeyDown?: (
     event: KeyboardEvent<HTMLInputElement>,
@@ -94,8 +93,6 @@ export interface Props
   icon?: ReactNode
   /** Custom input component */
   inputComponent?: ComponentType<InputProps>
-  /** Callback responsible for rendering the option given the option and its index in the list of options */
-  renderOption?: (option: Item, index: number) => ReactNode
   /** Provide unique key for each option */
   getKey?: (item: Item) => string
   /** Specifies whether the autofill enabled or not, disabled by default */
@@ -111,8 +108,6 @@ export interface Props
   /** Options provided to the popper.js instance */
   popperOptions?: PopperOptions
   inputProps?: BaseInputProps
-  /** Show the "Powered By Google" label */
-  poweredByGoogle?: boolean
   testIds?: InputProps['testIds'] & {
     menuItem?: string
     scrollMenu?: string
@@ -138,13 +133,13 @@ export const Autocomplete = forwardRef<HTMLInputElement, Props>(
       endAdornment,
       status,
       getDisplayValue = getItemText,
-      getKey: customGetKey,
+      getKey,
       icon,
       inputComponent,
       loading,
       menuWidth,
       name,
-      noOptionsText = 'No options',
+      noOptionsText,
       closeOnSelect,
       onBlur,
       onChange,
@@ -181,22 +176,6 @@ export const Autocomplete = forwardRef<HTMLInputElement, Props>(
       )
     }
 
-    const getKey = (item: Item) => {
-      if (customGetKey) {
-        return customGetKey(item)
-      }
-
-      const displayValue = getDisplayValue(item)
-
-      if (!displayValue) {
-        unsafeErrorLog(
-          'Autocomplete expects you to provide key prop value with getKey or Item.value!'
-        )
-      }
-
-      return displayValue
-    }
-
     const {
       highlightedIndex,
       isOpen,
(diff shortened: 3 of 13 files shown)


sashuk:
@toptal-anvil ping reviewers

cypress/component/TagSelector.spec.tsx
```diff
@@ -47,6 +47,7 @@ const TagSelectorExample = ({
     <Container padded='medium'>
       <TagSelector
         {...props}
+        getKey={item => item.code as string}
```
sashuk, line 50: It used to work in tests, but it caused lots of errors with invalid `key`. After changes in this pull request it broke, so this line is supposed to bring test back to React standards (so it works as expected)

packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
```diff
@@ -17,28 +17,33 @@ import type { PopperOptions } from 'popper.js'
...
-import OtherOptionMenuItem from './OtherOptionMenuItem'
 import type { Item, ChangedOptions } from './types'
 import { useAutocomplete, EMPTY_INPUT_VALUE } from './use-autocomplete'
 import { rootClassByWidth } from './styles'
+import type { OptionsMenuProps } from './OptionsMenu'
+import OptionsMenu from './OptionsMenu'
 
 export interface Props
   extends BaseProps,
+    Omit<
```
sashuk, line 39: This is small attempt in refactoring and extracting at least Autocomplete options menu. In this case the Autocomplete reuses some of the component properties of `OptionsMenu` to avoid duplication.

packages/base/Autocomplete/src/Autocomplete/test.tsx
```diff
@@ -129,12 +129,16 @@ describe('Autocomplete', () => {
...
-      renderAutocomplete({
+      const { getByTestId } = renderAutocomplete({
         options: testOptions,
         value: '',
         getKey,
       })
 
+      const input = getByTestId('autocomplete')
+
+      fireEvent.click(input)
```
sashuk, line 140: The extra click is needed as component no longer renders the options list before it is even needed (sometimes it is not needed ever, so we save a bit on performance).

packages/base/Autocomplete/src/Autocomplete/test.tsx
```diff
@@ -224,9 +228,8 @@ describe('Autocomplete', () => {
 
       fireEvent.click(input)
 
-      // 5 times on first render
-      // another 5 times on second rerender after click
-      expect(renderOption).toHaveBeenCalledTimes(10)
+      // 5 times on rerender after click
+      expect(renderOption).toHaveBeenCalledTimes(5)
```
sashuk, line 232: The `renderOption` is called 2 times less due to https://github.com/toptal/picasso/pull/4534/files#r1751693628 (no initial useless render)

packages/base/Autocomplete/src/Autocomplete/use-autocomplete/use-autocomplete.ts
```diff
@@ -96,6 +95,28 @@ export interface Props {
...
 
+type BaseItemsProps = {
+  role: string
+  'aria-selected': boolean
+  selected: boolean
+  onMouseMove: () => void
+  onMouseDown: (event: React.MouseEvent) => void
+}
+
+export type GetBaseItemPropsSignature = (index: number) => BaseItemsProps
```
sashuk, line 106: The hook exports type for easier reuse and typing of component that rely on `getBaseItemProps`, `getOtherItemProps`, etc functions

packages/base/Autocomplete/src/Autocomplete/use-autocomplete/use-autocomplete.ts
```diff
@@ -96,6 +95,28 @@ export interface Props {
   ref?: Ref<HTMLInputElement>
 }
 
+type BaseItemsProps = {
```
ruslan-sed, line 98: Maybe it would be better to keep these types in `OptionsMenu` that dictates what this type should be, wdyt? Hook only need to respect what `OptionsMenu` requires.

ruslan-sed (approved):
Nicely done!
Left some minor comments.

.changeset/three-coats-press.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+### TagSelector
+
+- feature: use `submitOtherOptionOnEnter` property to submit other option on Enter key press
+
+### Autocomplete
+
+- fix: do not show empty popper container when there are no options and no empty options message
+- refactor: dropdown options menu renders in separate component
```
augustobmoura, line 13: I think we should split this changeset file, so `CHANGELOG.md` files for `Autocomplete` and `TagSelector` only receive their relevant section

packages/base/Tagselector/src/TagSelector/test.tsx
```diff
@@ -121,6 +151,34 @@ describe('TagSelector', () => {
...
+    })
+
+    const input = screen.getByTestId('input')
+
+    fireEvent.change(input, { target: { value: firstOtherOptionText } })
+    fireEvent.keyDown(input, { key: 'Enter' })
+    fireEvent.change(input, { target: { value: secondOtherOptionText } })
+    fireEvent.keyDown(input, { key: 'Enter' })
+
+    const firstTag = getByText('aaa')
```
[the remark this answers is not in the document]
sashuk: @ruslan-sed Fixed, thank you!

packages/base/Autocomplete/src/Autocomplete/use-autocomplete/use-autocomplete.ts
sashuk: @ruslan-sed I think in this case this types are part of public API of the hook (in particular the signatures of functions returned by `useAutocomplete` hook), so maybe we could keep `OptionsMenu` as a consumer of those types, what do you think?

packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
```diff
@@ -17,28 +17,33 @@ import type { PopperOptions } from 'popper.js'
...
+      OptionsMenuProps,
+      | 'getDisplayValue'
+      | 'highlightedIndex'
+      | 'shouldShowOtherOption'
+      | 'getItemProps'
+      | 'getOtherItemProps'
+    >,
```
augustobmoura, line 46: To make this stricter and future-proof. Instead of omitting other props I would only pick the ones we already support, so in the case new props are added to `OptionsMenu` we don't have to add them here:
```suggestion
    Pick<
      OptionsMenuProps,
      | 'otherOptionText'
      | 'renderOtherOption'
      | 'noOptionsText'
      | 'renderOption'
      | 'poweredByGoogle'
      | 'getDisplayValue'
    >,
```

cypress/component/TagSelector.spec.tsx
augustobmoura: Wouldn't this be a breaking change? Can we be sure that this will not break in other places, besides cypress. Either on product projects, or their cypress equivalent. Ideally, we wouldn't need to update any tests, if there is no breaking changes

packages/base/Tagselector/src/TagSelector/TagSelector.tsx
```diff
@@ -71,10 +71,12 @@ export interface Props
   otherOptionLabel?: string
   /** Callback invoked when other option selected */
   onOtherOptionSelect?: (value: string) => void
+  /** Select other option on Enter key press */
+  submitOtherOptionOnEnter?: boolean
   /** Allow to show the other option in the list of options */
   showOtherOption?: boolean
   /** Label to show when no options were found */
-  noOptionsText?: string
+  noOptionsText?: string | null
```
augustobmoura, line 79: Why do we need both `?:` and `null` here? Wouldn't the property being optional already suffice?

cypress/component/TagSelector.spec.tsx
sashuk: @augustobmoura It should not, as the test was misconfigured from the very beginning, the internal logic of applying `getKey` in TagSelctor / Autocomplete have not changed

packages/base/Autocomplete/src/Autocomplete/Autocomplete.tsx
sashuk: @augustobmoura Agree, addressed, thank you!

.changeset/three-coats-press.md
sashuk: @augustobmoura Changeset was split, thanks!

packages/base/Tagselector/src/TagSelector/TagSelector.tsx
augustobmoura: Using `null` as a semantic symbol is a bit confusing, specially when paired with `undefined`. I usually expect both `undefined` and `null` to have the same meaning. Maybe we could use `false` here as an alternative? Wdyt? It is usually what is used on other component APIs to signify that a prop should be disabled
```suggestion
  noOptionsText?: string | false
```

sashuk: @augustobmoura I think in this case this could be addressed in a separate ticket, as this is just a refactoring that tries not to break existing API, what do you think?

sashuk: @augustobmoura I would prefer to keep it consistent with `options` (that can also be `null`), otherwise we will have a breaking change, what do you think? I agree that `null` is a bit cryptic, but this seems to be a change out of scope

Screenshot 2024-09-10 at 13 00 10

packages/base/Autocomplete/src/Autocomplete/use-autocomplete/use-autocomplete.ts
ruslan-sed: Yeah, it is part of hook API, but the type is dictated by `OptionsMenu`. Anyway, it is nitpicky and subjective, so keep it as you think it is better from your perspective.

ruslan-sed: We have three functions:
1. `getBaseItemProps`
2. `getItemProps`
3. `getOtherItemProps`

The first one is used only in the second and third functions. The other two are used only to get object with props inside of `OptionsMenu`. It is determined by `OptionsMenu` what we should pass to those functions as arguments and what it should return. That is what I meant by 'dictating' type.

sashuk: > The other two are used only to get object with props inside of OptionsMenu

I hope I see what you mean – overall, I think both `useAutocomplete` and `OptionsMenu` should define their public contracts (either inputs that they expect from consumer or outputs that provide to consumer) by themselves. This means that we define types of  `getItemProps` and `getOtherItemProps` in both components and let TypeScript do all the checks (if there is a mismatch, TypeScript will throw an error). However, it is going to be a lot of duplication.

This brings me to a thought that we still should keep these types defined only once (to avoid duplication) and to resolve the problem that you spotted (kudos!) at the same time. To address this, I moved these types to `types.ts`, so they are available to all interested parties from a single place.

Another things to keep in mind is scalability, so keeping types definitions outside of `OptionsMenu` will work well if we decide to have some new way of rendering options (so, there will be another component `OptionsButtons`) – the new component will still rely on the same properties that come from `types.ts`.

sashuk:
@toptal-anvil ping reviewers

packages/base/Autocomplete/src/Autocomplete/use-autocomplete/use-autocomplete.ts
ruslan-sed: I like your solution 👍 Thank you!

cypress/component/TagSelector.spec.tsx
augustobmoura: Got it, thanks for the explanation!

packages/base/Tagselector/src/TagSelector/TagSelector.tsx
augustobmoura: I don't know, using `null` is not really a pattern in Picasso, except for a few exceptions (this being one of them), most of the code uses optional properties at best.

Can we update the documentation for these props, then? Explaining that `null` is a special case and passing it to the props will effectively disable them

sashuk: @augustobmoura Documentation was updated, thanks! But for `options: null` there is nothing to add, as it is the same as `options: []`

augustobmoura (approved):

packages/base/Tagselector/src/TagSelector/TagSelector.tsx
augustobmoura: >But for options: null there is nothing to add, as it is the same as options: []

That is an example of an unnecessary `null` sigil
We can probably use `false` now, since the argument for consistency doesn't hold up anymore (`options` is not using `null` for the same reason as `noOptionsText`)

I'm already satisfied with the documentation for `null`, so no need to refactor it to `false`. But it is something to keep in mind in the future

sashuk: @augustobmoura Agree!

PR #4910, opened 2026-03-20
feat(Paper): add borderRadius prop

pltimme:
## Summary

Adds a `borderRadius` prop to the `Paper` component so consumers can use Picasso's design-system border-radius scale without resorting to inline `style` overrides.

- Accepts `'none' | 'sm' | 'md' | 'full'` — maps to Tailwind classes `rounded-none`, `rounded-sm` (4px), `rounded-md` (8px), `rounded-full`
- Defaults to `undefined` (no border radius applied — fully backward compatible)
- Exports `PaperBorderRadius` type

## Motivation

Without this prop, consumers who need rounded Paper surfaces (e.g. stat/metric cards) fall back to `style={{ borderRadius: Npx }}` with arbitrary pixel values outside the design system's scale. This was observed in downstream tooling where a `style={{ borderRadius: 6 }}` workaround was documented as a convention — a signal that the prop belongs in the component itself.

## Usage

```tsx
// Rounded stat card
<Paper borderRadius="sm">
  <Container padded={SPACING_6}>
    <Typography variant="heading" size="xlarge">{value}</Typography>
  </Container>
</Paper>
```

## Test plan

- [ ] Existing snapshot test passes unchanged (no default border radius added)
- [ ] New parameterised tests verify each `borderRadius` value applies the correct Tailwind class
- [ ] New test verifies no radius class is present when prop is omitted

### DIFF
--- a/.changeset/paper-border-radius-prop.md
+++ b/.changeset/paper-border-radius-prop.md
@@ -1,0 +1,6 @@
+---
+'@toptal/picasso-paper': minor
+'@toptal/picasso': minor
+---
+
+Add `borderRadius` prop to `Paper` component. Accepts `'none' | 'sm' | 'md' | 'full'` and maps to Picasso's Tailwind border-radius scale. Defaults to no border radius (backward compatible).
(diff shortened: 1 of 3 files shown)


pltimme:
Dropping this — the right fix is to use the existing Tailwind class directly via className on Paper. No new prop needed.

PR #4915, opened 2026-04-27
[TAPS-2284] Fix Tabs onChange not firing when re-clicking selected tab

PiotrRashkouski:
## Summary

Restores `Tab.onClick` on `@toptal/picasso-tabs`. The handler was silently
dropped after the migration to `@mui/base/Tabs` in
`@toptal/picasso-tabs@6.0.0` (PR #4874, "Migrate Tabs to Tailwind"), because
`@mui/base/Tab` destructures `onClick` out of its own props and never
forwards it. Wiring `onClick` through `slotProps.root` restores the
expected behavior, including firing on re-clicks of the already-selected
tab — which is what consumers like `StickyTabs` rely on to scroll the
section into view.

## Why

- Picasso's `Tab` advertises `onClick` as part of its public API.
- After the MUI base migration the prop became a no-op, breaking any
  consumer that did imperative work in `Tab.onClick` (e.g. `StickyTabs`
  scrolling on re-click of the active tab).
- `Tabs.onChange` continues to fire only on actual value changes, matching
  MUI base's `useTabsRoot` semantics. Consumers that need a "fired on every
  click" hook should use `Tab.onClick`.

## Changes

- `packages/base/Tabs/src/Tab/Tab.tsx` — forward `onClick` via
  `slotProps.root` so `@mui/base/Tab` actually attaches it to the rendered
  button. Top-level `onClick` is left in place for prop typing but is no
  longer the source of truth.
- `packages/base/Tabs/src/Tabs/test.tsx` — regression tests covering
  `Tab.onClick` on first click and on re-click of the selected tab, with
  both default positional values and explicit custom values.
- `.changeset/tabs-onchange-reclick.md` — patch changeset for
  `@toptal/picasso-tabs` and `@toptal/picasso` describing the fix and the
  `Tab.onClick` vs `Tabs.onChange` distinction for consumers.

## Notes

- No public API changes — pure regression fix.
- `Tabs.tsx` is untouched; `TabsContext` shape is unchanged.
- Disabled tabs continue to ignore clicks (existing
  "doesnt fire onChange when disabled" test still passes).

## Test plan

- [ ] `yarn jest packages/base/Tabs` — all tests pass, including the new
      `Tab.onClick` re-click cases.
- [ ] `yarn tsc -b packages/base/Tabs/tsconfig.json --force` — clean
      typecheck.
- [ ] Storybook smoke: open a `Tabs` story, attach a `console.log` to
      `Tab.onClick`, verify clicking the active tab logs every time.
- [ ] Downstream check on `StickyTabs` consumer: confirm scroll-on-re-click
      behavior is restored.

Jira: TAPS-2284

### DIFF
--- a/packages/base/Tabs/src/Tab/Tab.tsx
+++ b/packages/base/Tabs/src/Tab/Tab.tsx
@@ -114,12 +114,7 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
     ...rest
   } = props
   const titleCase = useTitleCase(propsTitleCase)
-  const {
-    orientation,
-    variant,
-    value: contextValue,
-    onChange: contextOnChange,
-  } = useContext(TabsContext)
+  const { orientation, variant } = useContext(TabsContext)
   const isHorizontal = orientation === 'horizontal'
 
   const renderLabel = getLabelComponent({
@@ -164,12 +159,7 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
               'relative ',
               className
             ),
-            onClick: (event: React.MouseEvent<HTMLButtonElement>) => {
-              onClick?.(event)
-              if (ownerState.selected) {
-                contextOnChange?.(event, contextValue)
-              }
-            },
+            ...(onClick && { onClick }),
           }
         },
       }}
--- a/packages/base/Tabs/src/Tab/Tab.tsx
+++ b/packages/base/Tabs/src/Tab/Tab.tsx
@@ -114,7 +114,12 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
     ...rest
   } = props
   const titleCase = useTitleCase(propsTitleCase)
-  const { orientation, variant } = useContext(TabsContext)
+  const {
+    orientation,
+    variant,
+    value: contextValue,
+    onChange: contextOnChange,
+  } = useContext(TabsContext)
   const isHorizontal = orientation === 'horizontal'
 
   const renderLabel = getLabelComponent({
@@ -135,7 +140,6 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
       disabled={disabled}
       value={value}
       onChange={onChange}
-      onClick={onClick}
       slotProps={{
         root: ownerState => {
           return {
@@ -160,6 +164,12 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
               'relative ',
               className
             ),
+            onClick: (event: React.MouseEvent<HTMLButtonElement>) => {
+              onClick?.(event)
+              if (ownerState.selected) {
+                contextOnChange?.(event, contextValue)
+              }
+            },
           }
         },
       }}
--- a/packages/base/Tabs/src/Tabs/test.tsx
+++ b/packages/base/Tabs/src/Tabs/test.tsx
@@ -56,41 +56,6 @@ const renderTabs = (
   )
 }
 
-const ControlledTabs = ({
-  tabs,
-  initialValue,
-  onChange,
-}: {
-  tabs: TabProps[]
-  initialValue: TabsValueType
-  onChange?: (event: React.ChangeEvent<{}> | null, value: TabsValueType) => void
-}) => {
-  const [value, setValue] = React.useState<TabsValueType>(initialValue)
-
-  return (
-    <TestingPicasso>
-      <Tabs
-        value={value}
-        onChange={(event, newValue) => {
-          setValue(newValue)
-          onChange?.(event, newValue)
-        }}
-      >
-        {tabs.map((tab, index) => (
-          <Tabs.Tab
-            key={index}
-            data-testid={`tab-${index + 1}`}
-            value={tab.value}
-            label={tab.label}
-            disabled={tab.disabled}
-            onClick={tab.onClick}
-          />
-        ))}
-      </Tabs>
-    </TestingPicasso>
-  )
-}
-
 describe('Tabs', () => {
   it('renders', () => {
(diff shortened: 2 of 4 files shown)


packages/base/Tabs/src/Tabs/test.tsx
```diff
@@ -165,4 +165,37 @@ describe('Tabs', () => {
...
+      {
+        value: 0,
+        onChange,
+      }
+    )
+
+    fireEvent.click(getByTestId('tab-1'))
+    expect(onChange).toHaveBeenCalledTimes(1)
+    expect(onChange).toHaveBeenCalledWith(expect.anything(), 0)
+  })
```
OleksandrNechai: Hmm, I thought we needed to fire onClick event. There was no problem with onChange firing, is not it? The problem was that onClick did not fire on clicking on same tab.

packages/base/Tabs/src/Tab/Tab.tsx
```diff
@@ -160,6 +164,12 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
               'relative ',
               className
             ),
+            onClick: (event: React.MouseEvent<HTMLButtonElement>) => {
+              onClick?.(event)
+              if (ownerState.selected) {
```
OleksandrNechai: why when selected only? we assume that owner was not selected before? But what if it was? Second click on same tab will fire onChange?

packages/base/Tabs/src/Tab/Tab.tsx
```diff
@@ -160,6 +164,12 @@ export const Tab = forwardRef<HTMLButtonElement, Props>(function Tab(
...
+                contextOnChange?.(event, contextValue)
```
OleksandrNechai: maybe we can consider namin somethign like tabsOnChange. Context is a bit too overloaded term.

OleksandrNechai: overal could you please specify why onChange needs to be invoked when we click tab? How this onClick override breaks tabs level onChange?

packages/base/Tabs/src/Tabs/test.tsx
OleksandrNechai: Or is this to prevent regression?

OleksandrNechai (commented):
I guess this could be solution, but I still don't understand it completely, left questions.

packages/base/Tabs/src/Tabs/test.tsx
```diff
@@ -165,4 +201,116 @@ describe('Tabs', () => {
...
+      { value: 0 }
+    )
+
+    fireEvent.click(getByTestId('tab-2'))
+    expect(onClick).not.toHaveBeenCalled()
+  })
+
+  it('does not fire onChange when re-clicking the already-selected tab', () => {
+    const onChange = jest.fn()
+    const { getByTestId } = render(
```
OleksandrNechai: why sometimes we have render, sometimes we have renderTabs?

packages/base/Tabs/src/Tabs/test.tsx
```diff
@@ -165,4 +201,116 @@ describe('Tabs', () => {
...
+        initialValue={0}
+        onChange={onChange}
+      />
+    )
+
+    fireEvent.click(getByTestId('tab-1'))
+    expect(onChange).not.toHaveBeenCalled()
+  })
+
+  it('fires onChange only once on double-click of an unselected tab', () => {
```
OleksandrNechai: does not this case overlaps with `does not fire onChange when re-clicking the already-selected tab`? Best unit test is the one that focuses on one thing.

packages/base/Tabs/src/Tabs/test.tsx
```diff
@@ -165,4 +201,116 @@ describe('Tabs', () => {
...
+        initialValue={0}
+      />
+    )
+
+    fireEvent.click(getByTestId('tab-1'))
+    fireEvent.click(getByTestId('tab-1'))
+    expect(onClick).toHaveBeenCalledTimes(2)
+  })
+
+  it('fires Tab onClick twice on double-click of an unselected tab', () => {
```
OleksandrNechai: sounds like overlap with `fires Tab onClick on every click, including re-click of the selected tab`

packages/base/Tabs/src/Tabs/test.tsx
```diff
@@ -165,4 +201,116 @@ describe('Tabs', () => {
...
+        initialValue={0}
+      />
+    )
+
+    fireEvent.click(getByTestId('tab-2'))
+    fireEvent.click(getByTestId('tab-2'))
+    expect(onClick).toHaveBeenCalledTimes(2)
+  })
+
+  it('fires Tab onClick twice but does not fire onChange on double-click of selected tab', () => {
```
OleksandrNechai: I had impression that we have tested earlier

OleksandrNechai (approved):
Looks good 👍 . I left comments, I would revisit tests because impression is such that they are overlapping each other.

packages/base/Tabs/src/Tabs/test.tsx
PiotrRashkouski: switched to `renderTabs`

packages/base/Tabs/src/Tabs/test.tsx
PiotrRashkouski: Removed

packages/base/Tabs/src/Tabs/test.tsx
PiotrRashkouski: Removed

packages/base/Tabs/src/Tabs/test.tsx
PiotrRashkouski: Removed

PR #4920, opened 2026-05-04
[PF-1993] Migrate from Yarn 1 to pnpm 10

denieler:
## Summary

Migrates picasso's package manager from Yarn 1 to pnpm 10. This unblocks `@toptal/davinci-syntax@24` (which requires pnpm) and aligns picasso with the rest of Toptal's frontend tooling.

- Lockfile converted via `pnpm import` to preserve every resolved version from `yarn.lock` (verified by byte-identical reproduction and per-package sampling).
- All 90 workspace packages declared as root devDependencies (concrete versions per syncpack) so `.storybook/*` and `tailwind.config.js` resolve them — yarn 1 auto-hoisted via the `workspaces` field, pnpm doesn't.
- `tsconfig` project references regenerated for the new workspace graph.
- All CI workflows + composite actions migrated (yarn-install action removed; local `pnpm-install` composite action handles corepack + pnpm-store cache + `--frozen-lockfile`).
- `Dockerfile` switched to `corepack` + `pnpm install --frozen-lockfile`.

### Override changes vs old `resolutions`
- ✅ Restored `@storybook/core-server>ip` → `@toptal/davinci-ip@2.0.3` (mitigates [GHSA-2p57-rm9w-gvfp](https://github.com/advisories/GHSA-2p57-rm9w-gvfp); the alias had been dropped in the WIP).
- 🗑️ Dropped `js-yaml ^3.13.1` global override — it was forcing every consumer to v3 and downgrading `lerna`, `semantic-release`, `jsdom`, `cosmiconfig`, etc. that yarn.lock had on v4.

### Other fixes
- `.eslintrc.js` extends path → `require.resolve(...)` (no longer depends on linker mode).
- `bin/build.js` tsc path → `require.resolve('typescript/bin/tsc')`.
- Alpha publish action: `npx davinci-engine` → `pnpm exec davinci-engine` (deterministic resolution against the just-installed binary).

### Storybook
Intentionally **not** upgraded — staying on Storybook 6 + the existing `@storybook/react ^6.5.15` override.

## Test plan

Local validation passed:
- [x] `pnpm install --frozen-lockfile`
- [x] `pnpm lint` (0 errors)
- [x] `pnpm typecheck`
- [x] `pnpm build:package` (90 projects)
- [x] `pnpm test:unit:ci`
- [x] `pnpm build:storybook` resolution clean (manager bundle built; preview bundle OOM'd locally — bundle size; CI runners have sufficient memory)

Watch in CI:
- [x] CI workflow (lint, typecheck, unit, circularity, syncpack)
- [x] `davinci-integration-tests.yml` — Cypress canary against the built artifacts
- [x] `visual-testing.yml` — Happo storybook smoke
- [x] Docker build


bin/add-missing-workspace-deps.mjs
```diff
@@ -0,0 +1,174 @@
...
+    if (changed) await writeJson(pkgPath, pkg)
+  }
+
+  console.log(`\nTotal deps added: ${totalAdded}`)
+}
+
+main().catch(e => {
+  console.error(e)
+  process.exit(1)
+})
```
denieler: delete before merging

docs/migration-references/pnpm-migration-plan-for-claude-code.md
```diff
@@ -0,0 +1,785 @@
+# pnpm Migration Plan for Claude Code (v2)
```
denieler: delete before merging

docs/migration-references/yarn-to-pnpm-migration-tutorial.md
```diff
@@ -0,0 +1,1696 @@
+# Yarn 1 to pnpm Migration Tutorial (v2)
```
denieler: delete before merging

MIGRATION_PLAN.md
```diff
@@ -0,0 +1,247 @@
+# Yarn → pnpm migration plan
```
denieler: delete before merging

patches/webpack@5.98.0.patch
```diff
@@ -0,0 +1,73 @@
+diff --git a/lib/FileSystemInfo.js b/lib/FileSystemInfo.js
```
denieler: copied from Staff Portal pnpm migration - https://github.com/toptal/staff-portal/blob/master/patches/webpack%2B5.95.0.patch

packages/picasso-provider/src/Favicon/test.tsx
```diff
@@ -1,8 +1,22 @@
 import React from 'react'
-import { render, waitFor } from '@toptal/picasso-test-utils'
+import { render as baseRender, waitFor } from '@testing-library/react'
```
denieler, line 2: to remove circularity test-utils -> provider -> shared -> test-utils

Dockerfile
```diff
@@ -1,5 +1,10 @@
 # syntax=docker/dockerfile:1
-FROM node:18-alpine
+ARG NODE_VERSION=22.20.0
+ARG PNPM_VERSION=10.32.1
```
vedrani, line 3: We can move this to package.json packageManager to have one single point of truth for version
https://github.com/toptal/staff-portal/blob/master/package.json#L153

Dockerfile
```diff
@@ -31,106 +37,16 @@ RUN chown -R node /app
...
-COPY --chown=node:node packages/base/Tooltip/package.json ./packages/base/Tooltip/package.json
-COPY --chown=node:node packages/base/TreeView/package.json ./packages/base/TreeView/package.json
-COPY --chown=node:node packages/base/TypographyOverflow/package.json ./packages/base/TypographyOverflow/package.json
-COPY --chown=node:node packages/base/Typography/package.json ./packages/base/Typography/package.json
-COPY --chown=node:node packages/base/Slide/package.json ./packages/base/Slide/package.json
-COPY --chown=node:node packages/base/UserBadge/package.json ./packages/base/UserBadge/package.json
-COPY --chown=node:node packages/base/Utils/package.json ./packages/base/Utils/package.json
+# Prune to manifests so dependency installation stays cache-friendly.
+RUN find packages -mindepth 2 -maxdepth 2 \! -name "package.json" -print | xargs rm -rf
+RUN find packages/base -mindepth 2 -maxdepth 2 \! -name "package.json" -print | xargs rm -rf
```
vedrani, line 47: After the first RUN, packages/base/ is empty, so the subsequent pnpm install --frozen-lockfile cannot resolve any of the 70+ base workspace packages and the image build will fail.

Or it's fine because you copied them?

Dockerfile
```diff
@@ -31,106 +37,16 @@ RUN chown -R node /app
 
 USER node
 
-# Enables layer caching
-COPY --chown=node:node package.json yarn.lock ./
+# Enables layer caching: copy lockfile + workspace manifests first,
+# install, then copy the rest of the source.
+COPY --chown=node:node package.json pnpm-lock.yaml pnpm-workspace.yaml .npmrc lerna.json ./
+COPY --chown=node:node packages packages
```
vedrani, line 43: The new flow is COPY packages packages → RUN find … rm -rf → RUN pnpm install. The first COPY layer's hash depends on every file under packages/, so the install layer invalidates on every code change.

package.json
```diff
@@ -81,38 +81,52 @@
...
+      "react-test-renderer": "^19.2.0",
+      "semver-regex": "^3.1.3",
+      "trim": "^0.0.3",
+      "unset-value": "^2.0.1",
+      "webpack": "^5.0.0",
+      "yaml": "2",
+      "micromatch": "^4.0.8",
+      "nx": "21.5.1",
+      "@nx/js": "21.5.1",
+      "@types/react": "17"
```
vedrani: Why not 18?

vedrani (approved):
Looks great 💯 

Just a couple of things to check. Claude is throwing couple of concerns about docker prune and caching.

dulishkovych (approved):

package.json
denieler: @vedrani we still use react 17 in some nested packages, so I used the "minimal compatible version"

denieler: @vedrani it seems this is the blocker to migrate to types from v18:

```
The build fails because packages/base/Utils/src/utils/forward-ref.ts:33-38 uses the older React 17 forwardRef signature. React 18 types tighten forwardRef's argument to ForwardRefRenderFunction<T, PropsWithoutRef<P>>. This is a real reason to keep 17 — unless we fix the helper. Let me try a targeted cast.
```

looks legit to me, but probably easy to fix also 🤔

Dockerfile
denieler: addressed in https://github.com/toptal/picasso/pull/4939

Dockerfile
denieler: addressed in https://github.com/toptal/picasso/pull/4939

Dockerfile
denieler: addressed in https://github.com/toptal/picasso/pull/4939

PR #4939, opened 2026-05-07
[PF-1993] Improve Dockerfile pnpm install caching

denieler:
## Summary

Follow-up to #4920. Addresses three points raised in code review:

- **Drop `ARG PNPM_VERSION`.** Corepack reads the version from `package.json#packageManager` (`pnpm@10.32.1`), so the duplicate constant in the Dockerfile is unnecessary and risks drift.
- **Fix the manifest prune.** The previous `find packages -mindepth 2 -maxdepth 2 ! -name "package.json" | xargs rm -rf` matches `packages/base/<name>` directories at depth 2 and removes them along with their `package.json` files. The follow-up `find packages/base ...` pass then runs against an empty directory. Once exercised in CI, `pnpm install --frozen-lockfile` would fail for the 70+ workspace packages under `packages/base/*`.
- **Make the install layer cacheable on manifests only.** `COPY packages packages` includes every source file in the layer's hash, so editing component code invalidates `node_modules`. Switching to explicit per-manifest `COPY` commands ties the install cache key to lockfile + workspace manifests only.

## Test plan

- [x] CI green on this branch
- [x] `docker build` against this Dockerfile produces a valid image with `node_modules` populated under `packages/base/<name>` workspaces
- [x] Editing a source file under `packages/<x>/src` does not invalidate the `pnpm install` layer in the next build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,7 @@
 # syntax=docker/dockerfile:1
 ARG NODE_VERSION=22.20.0
-ARG PNPM_VERSION=10.32.1
 
 FROM node:${NODE_VERSION}-alpine
-
-ARG PNPM_VERSION
 
 ENV PATH="${PATH}:/app/node_modules/.bin" \
   # Installs Chromium (77) package.
@@ -28,7 +25,8 @@ RUN apk add --no-cache \
 # so when we will be running container from CI it would have
 # all necessary rights for npm/pnpm publish
 RUN groupmod -g 469 node && usermod -u 469 -g 469 node
-RUN corepack enable && corepack prepare pnpm@${PNPM_VERSION} --activate
+# Corepack reads the pnpm version from package.json#packageManager.
+RUN corepack enable
 
 WORKDIR /app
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -38,13 +36,103 @@ RUN chown -R node /app
...
+# explicitly so the install layer's cache key depends only on the
+# manifests + lockfile, not on application source.
+# When a workspace package is added/removed, update the list below.
 COPY --chown=node:node package.json pnpm-lock.yaml pnpm-workspace.yaml .npmrc lerna.json ./
-COPY --chown=node:node packages packages
 
-# Prune to manifests so dependency installation stays cache-friendly.
-RUN find packages -mindepth 2 -maxdepth 2 \! -name "package.json" -print | xargs rm -rf
-RUN find packages/base -mindepth 2 -maxdepth 2 \! -name "package.json" -print | xargs rm -rf
+COPY --chown=node:node packages/base-tailwind/package.json ./packages/base-tailwind/package.json
```
denieler, line 45: this is returned back from the original Dockerfile solution from master

vedrani (approved):

PR #4947, opened 2026-05-08
[PF-1994] Migrate Button to @base-ui/react + Tailwind

vedrani:
# Button migration diff

Generated: 2026-05-08 08:39:43 CEST

**Package:** `packages/base/Button`

## Files

## Imports

**Removed:**

```
packages/base/Button/src/Button/Button.tsx:import { noop } from '@toptal/picasso-utils'
packages/base/Button/src/ButtonBase/ButtonBase.tsx:import { Button as MUIButtonBase } from '@mui/base/Button'
packages/base/Button/src/ButtonBase/ButtonBase.tsx:import type { ButtonRootSlotProps } from '@mui/base/Button'
```

**Added:**

```
packages/base/Button/src/Button/Button.tsx:import { noop, withClasses } from '@toptal/picasso-utils'
packages/base/Button/src/ButtonBase/ButtonBase.tsx:import { Button as BaseUIButton } from '@base-ui/react/button'
```

## MUI v4 / JSS residue check

| Check | Count |
|---|---|
0 |

## package.json delta

```diff
@@ -32,7 +32,7 @@
     "@toptal/picasso-utils": "4.0.0",
     "@toptal/picasso-link": "4.0.0",
     "ap-style-title-case": "^1.1.2",
-    "@mui/base": "5.0.0-beta.58",
+    "@base-ui/react": "^1.4.1",
     "classnames": "^2.5.1"
   },
   "sideEffects": [
@@ -43,7 +43,7 @@
     "@toptal/picasso-tailwind-merge": "^2.0.0",
     "@toptal/picasso-tailwind": ">=2.7",
     "@toptal/picasso-provider": "*",
-    "react": ">=16.12.0 < 19.0.0"
+    "react": ">=16.12.0"
   },
   "exports": {
     ".": "./dist-package/src/index.js"
```

## Prop-surface diff

```diff
```

</details>

## Happo
Happo log: `migration-runs/2026-05-08/Button/happo.log` (0
? flagged lines).

## React 19 smoke

### DIFF
--- a/packages/base/Button/src/Button/Button.tsx
+++ b/packages/base/Button/src/Button/Button.tsx
@@ -9,7 +9,7 @@ import type {
   OverridableComponent,
   TextLabelProps,
 } from '@toptal/picasso-shared'
-import { noop, withClasses } from '@toptal/picasso-utils'
+import { noop } from '@toptal/picasso-utils'
 // we need to ensure the correct order of styles import
 // TODO: [FX-4614] To be removed when Link component is migrated to tailwind
 import { Link } from '@toptal/picasso-link'
@@ -34,23 +34,13 @@ export type VariantType =
 
 export type IconPositionType = 'left' | 'right'
 
-export type ButtonClassKey = 'root' | 'label' | 'icon'
-
-const baseClasses: Record<ButtonClassKey, string> = {
-  root: '',
-  label: '',
-  icon: '',
-}
-
 export interface Props
-  extends Omit<StandardProps, 'classes'>,
+  extends StandardProps,
     TextLabelProps,
     ButtonOrAnchorProps {
   /** Show button in the active state (left mouse button down) */
   active?: boolean
   as?: ElementType
-  /** Override or extend per-slot classes (root, label, icon) */
-  classes?: Partial<Record<ButtonClassKey, string>>
   /** Disables button */
   disabled?: boolean
   /** Content of Button component */
@@ -85,13 +75,11 @@ const getIcon = ({
   icon,
   iconPosition,
   size,
-  iconClassName,
 }: {
   children: ReactNode
   icon?: ReactElement
   iconPosition?: IconPositionType
   size: SizeType<'small' | 'medium' | 'large'>
-  iconClassName?: string
 }) => {
   if (!icon) {
     return undefined
@@ -103,7 +91,7 @@ const getIcon = ({
   })
 
   return React.cloneElement(icon, {
-    className: twMerge(iconClassNames, iconClassName, icon.props.className),
+    className: twMerge(iconClassNames, icon.props.className),
   })
 }
 
@@ -115,7 +103,6 @@ export const Button: OverridableComponent<Props> = forwardRef<
     active = false,
     as = 'button',
     children = null,
-    classes,
     disabled = false,
     focused = false,
     fullWidth = false,
@@ -132,14 +119,11 @@ export const Button: OverridableComponent<Props> = forwardRef<
 ) {
   const { icon, className, ...rest } = props
 
-  const merged = withClasses(baseClasses, classes)
-
   const iconComponent = getIcon({
     children,
     icon,
     iconPosition,
     size,
-    iconClassName: merged.icon,
   })
   const coreClassNames = createCoreClassNames({
     disabled,
@@ -161,7 +145,6 @@ export const Button: OverridableComponent<Props> = forwardRef<
     variantClassNames,
     sizeClassNames,
     fullWidth ? 'w-full' : '',
-    merged.root,
     className
   )
 
@@ -174,13 +157,10 @@ export const Button: OverridableComponent<Props> = forwardRef<
     large: ['text-button-large'],
   }
 
-  const contentClassName = twMerge(
-    cx(
-      'font-semibold whitespace-nowrap',
-      contentSizeClassNames[size],
-      loading ? 'opacity-0' : ''
-    ),
-    merged.label
+  const contentClassName = cx(
+    'font-semibold whitespace-nowrap',
+    contentSizeClassNames[size],
+    loading ? 'opacity-0' : ''
   )
 
   return (
--- a/packages/base/Button/src/Button/Button.tsx
+++ b/packages/base/Button/src/Button/Button.tsx
@@ -35,7 +35,7 @@ export type VariantType =
 export type IconPositionType = 'left' | 'right'
 
 export interface Props
-  extends StandardProps,
+  extends Omit<StandardProps, 'classes'>,
     TextLabelProps,
     ButtonOrAnchorProps {
   /** Show button in the active state (left mouse button down) */
--- a/packages/base/Button/src/Button/Button.tsx
+++ b/packages/base/Button/src/Button/Button.tsx
@@ -9,7 +9,7 @@ import type {
   OverridableComponent,
   TextLabelProps,
 } from '@toptal/picasso-shared'
-import { noop } from '@toptal/picasso-utils'
+import { noop, withClasses } from '@toptal/picasso-utils'
 // we need to ensure the correct order of styles import
 // TODO: [FX-4614] To be removed when Link component is migrated to tailwind
 import { Link } from '@toptal/picasso-link'
@@ -34,13 +34,23 @@ export type VariantType =
 
 export type IconPositionType = 'left' | 'right'
 
+export type ButtonClassKey = 'root' | 'label' | 'icon'
+
+const baseClasses: Record<ButtonClassKey, string> = {
+  root: '',
+  label: '',
+  icon: '',
+}
+
 export interface Props
-  extends StandardProps,
+  extends Omit<StandardProps, 'classes'>,
     TextLabelProps,
     ButtonOrAnchorProps {
   /** Show button in the active state (left mouse button down) */
   active?: boolean
   as?: ElementType
+  /** Override or extend per-slot classes (root, label, icon) */
+  classes?: Partial<Record<ButtonClassKey, string>>
   /** Disables button */
   disabled?: boolean
   /** Content of Button component */
@@ -75,11 +85,13 @@ const getIcon = ({
   icon,
   iconPosition,
   size,
+  iconClassName,
 }: {
   children: ReactNode
   icon?: ReactElement
   iconPosition?: IconPositionType
   size: SizeType<'small' | 'medium' | 'large'>
+  iconClassName?: string
 }) => {
   if (!icon) {
     return undefined
@@ -91,7 +103,7 @@ const getIcon = ({
   })
 
   return React.cloneElement(icon, {
-    className: twMerge(iconClassNames, icon.props.className),
+    className: twMerge(iconClassNames, iconClassName, icon.props.className),
   })
 }
 
@@ -103,6 +115,7 @@ export const Button: OverridableComponent<Props> = forwardRef<
     active = false,
     as = 'button',
     children = null,
+    classes,
     disabled = false,
     focused = false,
     fullWidth = false,
@@ -119,11 +132,14 @@ export const Button: OverridableComponent<Props> = forwardRef<
 ) {
   const { icon, className, ...rest } = props
 
+  const merged = withClasses(baseClasses, classes)
+
   const iconComponent = getIcon({
     children,
     icon,
     iconPosition,
     size,
+    iconClassName: merged.icon,
   })
   const coreClassNames = createCoreClassNames({
     disabled,
@@ -145,6 +161,7 @@ export const Button: OverridableComponent<Props> = forwardRef<
     variantClassNames,
     sizeClassNames,
     fullWidth ? 'w-full' : '',
+    merged.root,
     className
   )
 
@@ -157,10 +174,13 @@ export const Button: OverridableComponent<Props> = forwardRef<
     large: ['text-button-large'],
   }
 
-  const contentClassName = cx(
-    'font-semibold whitespace-nowrap',
-    contentSizeClassNames[size],
-    loading ? 'opacity-0' : ''
+  const contentClassName = twMerge(
+    cx(
+      'font-semibold whitespace-nowrap',
+      contentSizeClassNames[size],
+      loading ? 'opacity-0' : ''
+    ),
+    merged.label
   )
 
   return (
--- a/packages/base/Button/src/ButtonBase/ButtonBase.tsx
+++ b/packages/base/Button/src/ButtonBase/ButtonBase.tsx
@@ -19,7 +19,7 @@ import { createCoreClassNames } from './styles'
 export type IconPositionType = 'left' | 'right'
 
 export interface Props
-  extends StandardProps,
+  extends Omit<StandardProps, 'classes'>,
     TextLabelProps,
     ButtonOrAnchorProps {
   as?: ElementType
--- a/packages/base/Button/src/ButtonBase/ButtonBase.tsx
+++ b/packages/base/Button/src/ButtonBase/ButtonBase.tsx
@@ -9,8 +9,7 @@ import type {
   TextLabelProps,
 } from '@toptal/picasso-shared'
 import { useTitleCase } from '@toptal/picasso-shared'
-import { Button as MUIButtonBase } from '@mui/base/Button'
-import type { ButtonRootSlotProps } from '@mui/base/Button'
+import { Button as BaseUIButton } from '@base-ui/react/button'
 import { Loader } from '@toptal/picasso-loader'
 import { Container } from '@toptal/picasso-container'
 import { noop, toTitleCase } from '@toptal/picasso-utils'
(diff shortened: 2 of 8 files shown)


narghev (approved):

packages/base/Button/src/Button/Button.tsx
```diff
@@ -34,13 +34,23 @@ export type VariantType =
...
 export interface Props
-  extends StandardProps,
+  extends Omit<StandardProps, 'classes'>,
     TextLabelProps,
     ButtonOrAnchorProps {
   /** Show button in the active state (left mouse button down) */
   active?: boolean
   as?: ElementType
+  /** Override or extend per-slot classes (root, label, icon) */
+  classes?: Partial<Record<ButtonClassKey, string>>
```
vedrani: Maybe we should ditch StandardProps and move to something like SlottedProps

```
export interface SlottedProps<K extends string = string> {
  classes?: Partial<Record<K, string>>
}
```

and then

```
export interface ButtonProps extends SlottedProps<ButtonClassKey>, TextLabelProps, ButtonOrAnchorProps {
  // no manual classes prop — it's already typed via the generic
}
```

packages/base/Button/src/Button/Button.tsx
```diff
@@ -75,11 +85,13 @@ const getIcon = ({
   icon,
   iconPosition,
   size,
+  iconClassName,
```
vedrani: Why did you add iconClassName?

packages/base/Button/src/Button/Button.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Considering this. You're right — the `Omit<StandardProps, 'classes'>` + manual narrowing pattern repeats per-component, while `SlottedProps<K>` factors it into one generic. The hesitation is scope: this changes `packages/shared/src/types.ts` mid-migration, so it affects every Tier 0/1 component currently in flight (Backdrop, Switch, Drawer, etc.) — not just Button.

Proposal:
1. Add `SlottedProps<K extends string = string>` to `packages/shared/src/types.ts` (alongside, not replacing, `StandardProps` — components without `classes` keep using `StandardProps`).
2. Update Button.tsx to: `extends SlottedProps<ButtonClassKey>, BaseProps, TextLabelProps, ButtonOrAnchorProps` (still need `BaseProps` for `className`/`style`/`data-testid`).
3. Drop the manual `classes?: ...` override from Button's Props.
4. Document the new convention in `docs/migration/rules/api-preservation.md` so subsequent migrations adopt it.

Trade-off: lighter touch is keeping `Omit<StandardProps, 'classes'>` as-is for Button only and adopting `SlottedProps` from the next migration onward (avoids reopening the shared-types diff for an in-flight refactor). Heavier touch is doing it now and back-filling the in-flight components.

Apply the `SlottedProps` refactor in this PR? 👍 to confirm.

packages/base/Button/src/Button/Button.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Added `iconClassName` to plumb `merged.icon` (the result of `withClasses(baseClasses, classes)`) through to the `cloneElement(icon, { className: ... })` call. That's how a consumer's `classes={{ icon: 'mt-1' }}` actually reaches the rendered icon's `className` — without this parameter the `icon` slot key in `ButtonClassKey` would be declared but unreachable.

Three alternatives if you'd prefer simpler:

1. **Inline the merge**: drop the `iconClassName` param from `getIcon`, do the `twMerge(merged.icon, ...)` at the call site. Cleaner signature, but the merge logic spreads.
2. **Drop the `icon` slot entirely**: trim `ButtonClassKey` to `'root' | 'label'`. Consumers wanting to style the icon style the icon element they pass in. Smaller API.
3. **Keep as-is**: parameter stays, full slot coverage.

Per `docs/migration/components/Button.md` "Slot keys" section, the canonical set is `'root' | 'label' | 'icon'` — so option 3 matches the plan, and (1)/(2) would diverge.

Want (1) inline-merge, (2) drop the icon slot, or (3) keep as-is? 👍 to keep, or reply with the option number.

vedrani: I think there are couple of problems here:

* icon already receives classname from icon.props.className
* Button was never using classes it was passing it to ButtonBase as ...rest and I suppose MuiButtonBase was using it in some way.

Investigate and share opinion.

vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Investigated. You're right on both points:

**1. `icon.props.className` already flows through.** Pre-migration `getIcon` at the equivalent of line 93 does `twMerge(iconClassNames, icon.props.className)`, so `<Button icon={<Icon className="mt-1" />} />` already reaches the rendered icon. My added `iconClassName` parameter is redundant for the icon-styling use case — consumers can already pass className on the icon element they hand in.

**2. `classes` was dead on Button pre-migration.** Verified against `node_modules/@mui/base/Button/Button.types.d.ts` — `ButtonOwnProps` accepts only `action`, `children`, `className`, `slotProps`, `slots`, `rootElementName`. **No `classes` prop.** Pre-migration Button.tsx never extracted `classes` either; it leaked from `StandardProps` through `{...rest}` → ButtonBase → `<MUIButtonBase {...rest}>` and was dropped on the floor as an unknown attribute. Grep across `packages/**/*.{ts,tsx}` finds zero `<Button[...] classes={...}>` callsites in the monorepo.

So the `withClasses` scaffolding I added is preserving a phantom feature. The Button.md plan file put Button in the "apply withClasses" bucket because Button extends `StandardProps` — but the audit missed that the prop wasn't actually rendered.

**Proposal: rip the scaffolding out of this PR.**

Concretely:
- Revert `Omit<StandardProps, 'classes'>` → `extends StandardProps` (prop surface stays identical to pre-migration: `classes?: Classes` inherited but unused, same as before — no regression).
- Drop `ButtonClassKey`, `baseClasses`, `withClasses` import & call, `iconClassName` parameter on `getIcon`.
- Delete `docs/migration/Button-diff.json` (no narrowing, no removal).
- Keep the `@base-ui/react/button` swap in ButtonBase as-is (the actual migration work).

This also moots the `SlottedProps<K>` proposal *for this PR* — if `classes` isn't live on Button, there's nothing to type with `SlottedProps<ButtonClassKey>`. `SlottedProps` could still be a worthwhile abstraction for components that DO actively use `classes` (Modal, Container, FormLabel, Typography per the audit) — but that's a separate refactor PR, not this one.

Apply the revert? 👍 to confirm, or push back if you want classes preserved (e.g. for future Tailwind-routing use even though it's currently a no-op).

packages/base/Button/src/ButtonBase/ButtonBase.tsx
```diff
@@ -98,7 +91,8 @@ export const ButtonBase: OverridableComponent<Props> = forwardRef<
 
   const titleCase = useTitleCase(propsTitleCase)
   const finalChildren = [titleCase ? toTitleCase(children) : children]
-  const finalRootElementName = typeof as === 'string' ? as : 'a'
```
vedrani, line 101: Do we still support this case where it can be `as` or `a`?

packages/base/Button/src/Button/Button.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — applied the revert in this iteration's pending commit:

- `Button.tsx`: dropped `ButtonClassKey`, `baseClasses`, `withClasses` import & call, `iconClassName` parameter on `getIcon`. Reverted `Omit<StandardProps, 'classes'>` → `extends StandardProps`. Restored the original `cx`-based `contentClassName`.
- `Button-diff.json`: emptied `removed` / `renamed` / `narrowed` arrays (the file should be deleted, but the orchestrator's sandbox blocks `rm`/`git rm` — emptied as the closest reachable state; please delete on merge or in a follow-up).
- ButtonBase.tsx is unchanged from this PR's earlier state (the actual `@mui/base` → `@base-ui/react/button` swap, which is the migration's real work).

Gate run locally: scoped lint 0 errors, `yarn workspace @toptal/picasso-button build:package` clean, 36/36 tests pass, 20/20 snapshots match (no regenerations needed — the snapshots already converged when I previously regenerated for the `data-disabled=""` + stale `base-` literal removal in iteration 1).

This also moots the `SlottedProps<K>` thread for this PR — there's no longer a `classes` prop type to factor out. Closing that proposal as obsolete; the SlottedProps abstraction is still worth pursuing for components that DO actively render `classes` (Modal, Container, FormLabel, Typography per the audit), but in a separate refactor PR.

packages/base/Button/src/Button/Button.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Closing this thread as obsolete for this PR — see [reply on the iconClassName thread](https://github.com/toptal/picasso/pull/4947#discussion_r3207422857). After investigating the iconClassName question I found that Button's `classes` prop was vestigial pre-migration (no `classes` prop on `@mui/base/Button`, and the public `Button.tsx` never extracted/rendered it). With your 👍 on the revert, I've removed the entire `withClasses`/`ButtonClassKey` scaffolding from this PR — so there's no longer a `classes` shape to factor through `SlottedProps<K>` here.

The `SlottedProps<K>` abstraction is still a good idea for the components that DO actively render `classes` (Modal, Container, FormLabel, Typography per the audit). Recommend pursuing it as a separate refactor PR before the Tier 1 migrations land, so those components adopt it from day one. Want me to open a follow-up issue / PR proposal for that?

packages/base/Button/src/ButtonBase/ButtonBase.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Yes — all the `as` cases stay supported; the deleted line was specific to `@mui/base/Button`'s slot API and isn't needed against `@base-ui/react/button`.

**What the deleted line did**: `const finalRootElementName = typeof as === 'string' ? as : 'a'` produced a `keyof HTMLElementTagNameMap` for `@mui/base/Button`'s `rootElementName` prop. `@mui/base` needed a string tag-name hint separate from the actual rendered component (which it took from `slotProps.root.as`). The `'a'` fallback was a placeholder tag for the case where `as` was a component (e.g. `<Button as={Link}>`) — since `rootElementName` only accepts string keys.

**What the migration does instead**: `@base-ui/react/button` doesn't have `rootElementName` — it has `render: ReactElement | undefined` plus `nativeButton: boolean`. The two replacement lines:

```ts
const finalAs: ElementType = as ?? 'button'
const isNativeButton = finalAs === 'button'
```

feed directly into:

```tsx
<BaseUIButton
  nativeButton={isNativeButton}
  render={isNativeButton ? undefined : React.createElement(finalAs)}
  ...
>
```

Cases:
- `as` omitted / `as='button'` → `nativeButton: true`, `render: undefined` → Base UI renders the native `<button>`.
- `as='a'` → `nativeButton: false`, `render: <a/>` → Base UI renders `<a>` and skips the native-button warning.
- `as='span'` → `nativeButton: false`, `render: <span/>` → renders `<span>`.
- `as={Link}` (component) → `nativeButton: false`, `render: <Link/>` → renders the Link component (which itself renders an `<a>`).

The existing snapshot tests cover all four shapes (`as='a'`, `as=Link` component, `as='span'` with `href`/`to`, default `'button'`) — all 36/36 pass against the migrated code with no behavioral diff (only the one-time `data-disabled=""` + stale `base-` literal removal).

One user-facing semantic delta worth noting: pre-migration, when `as={Link}` (component), the `rootElementName` hint was always `'a'` regardless of what tag the component actually rendered. This was a no-op since `rootElementName` was just a hint. Post-migration there's no equivalent hint; the rendered tag is whatever the `render` element produces. No test breakage observed.

**Should we keep the `as` prop's current API?** Per `rules/api-preservation.md` — yes, unchanged. `as?: ElementType` accepts string tags and components, and the test snapshots verify both shapes still render correctly.

packages/base/Button/src/Button/Button.tsx
vedrani: I think there is another problem here, Button is just passing classes as part of rest to ButtonBase, and ButtonBase was doing same to MUIButtonBase. If somebody would pass `classes` they would apply to MUIButtonBase. Research implications of it and propose pragmatic solution for it.

vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Researched. You're right — the leak path is:

```
<Button classes={obj} />
  → Button.tsx: `const { icon, className, ...rest } = props` (classes NOT extracted) → falls into rest
  → <ButtonBase {...rest} />
  → ButtonBase.tsx: `const { icon, className, contentClassName, style, title, value, titleCase, ...rest } = props` (classes NOT extracted) → falls into rest
  → <BaseUIButton {...rest} /> (pre-migration: `<MUIButtonBase {...rest}>`)
  → BaseUIButton has no `classes` prop, spreads to underlying `<button>`/`<a>`
  → React DOM: object-valued non-HTML attribute
```

**React's actual behavior with `classes={obj}` on a DOM element**:
- React 18 dev: console warning `Warning: Invalid value for prop \`classes\` on <button> tag. Either remove it from the element, or pass a string or number value...`. Prop is dropped from the rendered HTML.
- React 18 prod: silently dropped, no warning.
- React 19: silently dropped (the warning was relaxed for non-DOM-valid prop names).
- React 16.x (Picasso's floor): warning fires, prop dropped.

So pre-migration consumers passing `<Button classes={...}>` already got dev warnings + zero styling effect — confirming the prop was dead. My migration (and current pending revert) preserves that behavior unchanged.

**Pragmatic solution** — strip `classes` at the boundary so it never reaches the DOM:

```ts
// Button.tsx
// eslint-disable-next-line @typescript-eslint/no-unused-vars
const { icon, className, classes: _classes, ...rest } = props
```

```ts
// ButtonBase.tsx — same: add `classes: _classes` to the destructure
```

**Effect**: `classes` no longer leaks via `{...rest}` to `BaseUIButton` → no DOM warning when consumers pass it. Public type unchanged (`classes` still in scope via `StandardProps`/`JssProps`), so existing consumer code that types it doesn't break. Behavioral delta vs pre-migration is: a dev-mode warning goes away. No functional regression.

**Alternatives considered**:

| Option | Pros | Cons |
|---|---|---|
| A. Strip in destructure (proposed above) | Minimal, no breaking change, kills the warning | The `_classes` rename + eslint-disable is a minor wart |
| B. Strip + drop from type via `Omit<StandardProps, 'classes'>` | API is honest about not using it | Breaking change for any external consumer typing `classes={...}` (zero internal callsites, but 23-repo portfolio not audited) |
| C. Leave as-is | No code change | Dev console keeps complaining; the bug vedrani flagged stays |

Recommend **Option A** as the pragmatic fix — it solves the actual symptom (DOM leak warning) without touching the public type surface.

Apply Option A? 👍 to confirm, or reply with 'B' if you want the type-level cleanup too. (Heads up: if A, I'd also strip `classes` in ButtonBase.tsx since the same leak path exists one layer down.)

vedrani: B

vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — applied Option B in this iteration's pending commit:

- `Button.tsx`: `extends StandardProps` → `extends Omit<StandardProps, 'classes'>`.
- `ButtonBase.tsx`: same — `extends Omit<StandardProps, 'classes'>` (the leak path existed one layer down too, so it needed both).
- `docs/migration/Button-diff.json`: restored with a `removed` entry for `classes` (`codemod: required`) — reason notes that the prop was vestigial pre-migration and that zero internal call-sites pass it, but the 23-repo external portfolio is unaudited.

**Effect**: TypeScript now prevents `<Button classes={...}>` at compile time, so the DOM-leak path is closed at the API surface. No runtime strip needed since Picasso is TS-typed and consumers compile against this type. JS-only callers bypassing TS would still hit React's existing dev-mode warning, but that's no worse than pre-migration behavior.

**Gate run locally**: scoped lint 0 errors, `yarn workspace @toptal/picasso-button build:package` clean, `yarn typecheck` (full repo) passes — no internal consumer broke from the type narrowing, 36/36 tests pass, 20/20 snapshots match (no regeneration needed).

Flagging for codemod tracking (PF-2024): the `classes` removal goes into the diff JSON with `codemod: required`. External consumers passing `classes={...}` to picasso Button will see a TS error and need to remove the prop — that's the breaking change. Recommend the codemod author scan the 23-repo portfolio for `<Button[^>]*classes=` shapes and emit a removal codemod alongside any other Button prop-surface diffs.

narghev (approved):

PR #4956, opened 2026-05-13
Add picasso design patterns

denieler:
### Description

Introducing the AI Claude Review check to validate newly added or modified Picasso components against Picasso base design rules and patterns. This check supposed to be non-blocking for now. It should provide useful information to the component author.

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


PICASSO_COMPONENT_DESIGN_PATTERNS.md
```diff
@@ -0,0 +1,90 @@
...
+4. **Mirror native HTML prop names.** When a prop corresponds to a native HTML attribute, use the native name verbatim (e.g., `name`, `value`, `type`, `disabled`, `placeholder`, `checked`, `readOnly`, `autoFocus`, `href`, `target`, `rel`, `alt`, `src`). Do not rename, prefix, or alias native attributes (avoid `inputName`, `fieldValue`, `isDisabled`, `linkHref`). For event handlers, keep the native name (`onChange`, `onBlur`, `onFocus`, `onClick`, …), but the callback signature may diverge from the native one when a simpler shape is more ergonomic — e.g., `onChange?: (value: string) => void` is preferred over passing the raw event when only the value is needed.
+5. **Style overrides only via `className` or `style`.** Consumers may customize a component's appearance exclusively through the `className` and `style` props. Do not expose other styling hooks such as `classes`, `styles`, `sx`, `css`, theme overrides, slot-level class props, or styled-component wrappers as part of the public API.
```
vedrani, line 11: We should put here some exception for existing state until we migrate from it and decide how to replace some of them with e.g. slots

PICASSO_COMPONENT_DESIGN_PATTERNS.md
```diff
@@ -0,0 +1,90 @@
...
+6. **Prefer `children` over content props.** For simple components, pass content through `children` rather than a dedicated prop. Use `<Button>Save</Button>`, not `<Button content="Save" />`. Reserve named content props for components with multiple, distinct content slots.
+7. **Use `rem` for all sizes.** Express dimensions, spacing, font sizes, and radii in `rem`. The only exception is `1px` (e.g., hairline borders), which may be used directly.
```
vedrani, line 13: I think there are some other exceptions

PICASSO_COMPONENT_DESIGN_PATTERNS.md
```diff
@@ -0,0 +1,90 @@
...
+      gradients: {
+        blue: string
+        darkerBlue: string
+        lightGrey: string
+        grey: string
+        darkerGrey: string
+      }
+    }
+    ```
+14. **No `is` prefix on boolean props.** Boolean props are named with the adjective only — `open`, `disabled`, `loading`, `selected`, `collapsed` — not `isOpen`, `isDisabled`, `isLoading`. The same applies to `has`/`should` prefixes.
```
vedrani, line 63: Ok, but at the same time we should add rule to keep backward compatibility of component interfaces as MUST and only human should approve breaking changes.

vedrani (commented):
Can you extend it with some exceptions for current state so that agent can keep interfaces stable. If I plan to use this I will add in migration prompt that goal is to keep backward compatibility during migration.

PICASSO_COMPONENT_DESIGN_PATTERNS.md
denieler: no exceptions, if it will fail - we will see that. We can add instructions to the orchestrator to ignore this error maybe 🤔

PICASSO_COMPONENT_DESIGN_PATTERNS.md
denieler: I checked codebase of Picasso and couldn't find any that I'd say usage of `px` makes sense there

PICASSO_COMPONENT_DESIGN_PATTERNS.md
denieler: I think it's not a "design pattern", it's more like a development practice to prevent breaking changes and keep backward compatibility if the change is really justifiable 🤔 I understand from the orchestrator point of view this addition, but if talking about this particular check on isolation - I think it doesn't make sense to include it here 🤔

PICASSO_COMPONENT_DESIGN_PATTERNS.md
vedrani: Yeah, I could add some exception section to overrule it for migration phase.

PICASSO_COMPONENT_DESIGN_PATTERNS.md
vedrani: I agree with you that it should be kept pure. But in that case we need to add additional info based on 'migration', 'review' or some other phase what is real-world practice.

E.g. 
* new-feature - error 
* migration - backward-compatibility over patterns -> warning/ignore
* review - warning

vedrani (approved):

PR #4957, opened 2026-05-13
[PF-1994] Migrate Badge to @base-ui/react + Tailwind

vedrani:
# Badge migration diff

Generated: 2026-05-14 00:10:05 CEST

**Package:** `packages/base/Badge`

## Files

## Imports

**Removed:**

```
packages/base/Badge/src/Badge/Badge.tsx:import { Badge as MuiBadge } from '@mui/base'
```

## MUI v4 / JSS residue check

| Check | Count |
|---|---|
0 |

## package.json delta

```diff
@@ -22,7 +22,6 @@
   },
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "dependencies": {
-    "@mui/base": "5.0.0-beta.58",
     "@toptal/picasso-shared": "15.0.0",
     "@toptal/picasso-utils": "4.0.0"
   },
@@ -32,7 +31,7 @@
   ],
   "peerDependencies": {
     "@toptal/picasso-tailwind-merge": "^2.0.0",
-    "react": ">=16.12.0 < 19.0.0"
+    "react": ">=16.12.0"
   },
   "exports": {
     ".": "./dist-package/src/index.js"
```

## Prop-surface diff

```diff
```

</details>

## Happo
Happo log: `migration-runs/2026-05-13/Badge/happo.log` (0
? flagged lines).

## React 19 smoke

### DIFF
--- a/packages/base/Badge/src/Badge/Badge.tsx
+++ b/packages/base/Badge/src/Badge/Badge.tsx
@@ -1,6 +1,5 @@
 import type { ReactNode } from 'react'
 import React, { Children, forwardRef } from 'react'
-import { Badge as MuiBadge } from '@mui/base'
 import type { BaseProps } from '@toptal/picasso-shared'
 import { twJoin, twMerge } from '@toptal/picasso-tailwind-merge'
 
@@ -28,7 +27,7 @@ const thresholds: Record<SizeType, number> = {
   large: 99,
 }
 
-export const Badge = forwardRef<HTMLDivElement, Props>(function Badge(
+export const Badge = forwardRef<HTMLSpanElement, Props>(function Badge(
   {
     children,
     style,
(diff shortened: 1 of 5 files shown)


packages/base/Badge/src/Badge/Badge.tsx
```diff
@@ -28,7 +27,7 @@ const thresholds: Record<SizeType, number> = {
   large: 99,
 }
 
-export const Badge = forwardRef<HTMLDivElement, Props>(function Badge(
+export const Badge = forwardRef<HTMLSpanElement, Props>(function Badge(
```
vedrani, line 30: What's the reason to use `span` and not `div`?

vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Two reasons:

1. **DOM-preserving migration.** `@mui/base` Badge defaulted its root slot to `span` (`const Root = slots.root ?? 'span'`, see `node_modules/@mui/base/Badge/Badge.js`). Pre-migration consumers received a `<span>` at runtime — the old `forwardRef<HTMLDivElement>` ref type was actually misleading. Keeping `span` keeps the rendered DOM identical post-migration.
2. **Semantics.** The root carries `inline-flex` + `align-middle` and is designed to sit inline next to text/icons (e.g. tabs, table cells, button labels). A `div` would be block by default — fine with `inline-flex`, but it changes default phrasing-vs-flow-content semantics and could surprise consumers whose surrounding markup currently allows a span but not a div (e.g. inside `<p>` or other inline-only contexts).

Happy to switch to `div` if you prefer — the visual output would be unchanged since `inline-flex` overrides the default display. But it would be a DOM-level breaking change for any consumer rendering Badge inside phrasing content. 👍 to switch, otherwise I'll leave it as `span`.

narghev (approved):

PR #4963, opened 2026-05-15
[PF-2031] Upgrade TypeScript to 5.5 and align davinci tooling

dulishkovych:
## Summary

- Bumps TypeScript from `~4.7.0` to `~5.5.0` across root + 10 publishable packages.
- Pulls lint / test / codegen infrastructure from the stable releases of [toptal/davinci#2677](https://github.com/toptal/davinci/pull/2677): `syntax@^25.0.0`, `engine@^15.0.0`, `qa@^19.1.0`, `ci@^8.0.0`, `code@^3.0.0`. (`ci` and `code` had no TS-5.5 changes, so they resolve to the existing stable majors.)
- Bumps `@babel/preset-typescript` to `^7.26.0` (required by `davinci-engine@15`).

> **Note**: This PR supersedes #4952. Squashed onto a fresh branch from master after extensive investigation into the alpha-publish failures on the previous branch (see [#4952](https://github.com/toptal/picasso/pull/4952) for full history). All review comments from #4952 should be re-raised here.

## Public type changes

- `OverridableComponent<P>` in `@toptal/picasso-shared` is rewritten as a single-signature interface `(props: P & { [key: string]: any }) => JSX.Element | null`. Declared fields of `P` remain strictly typed at JSX call sites (e.g. `<Button size={42} />` still errors), and any other prop is accepted untyped. This lets `forwardRef<R, P>(...)` assign directly without an escape hatch, even when `P` has required fields, and preserves the polymorphic `as` usage pattern. **Trade-off** versus the pre-PF-2031 shape: TypeScript no longer pulls prop types FROM the `as` target — `<Button as={Link} to={...} />` does not validate `to` against `Link`'s props. Full polymorphic-inheritance typing for the `as` prop is tracked in [FF-125](https://toptal-core.atlassian.net/browse/FF-125).
- `composeValidators` (`@toptal/picasso-forms`) is now generic over `TValue` so downstream inference propagates.

## Internal type cleanups (no public API impact)

Resolve `any` regressions surfaced by `@typescript-eslint` v8 in: `picasso-codemod`, `picasso-shared`, `Tagselector`, `Container`, `Menu`, `PromptModal`, `NumberInput`. `OverviewBlock`, `Page`, `Breadcrumbs`, `Button`, `ButtonBase`, `ButtonCircular`, `MenuItem`, `Link`, and `SidebarItem` compile cleanly without source changes under the new `OverridableComponent` shape. `ButtonAction` got a one-line internal fix (an `icon` helper returning `null` where `ReactElement | undefined` was declared) that the stricter declared-prop typing in the new shape surfaced.

Per review, opportunistic `any`→type tightening unrelated to the upgrade was reverted to keep the diff scoped (List, Select, TreeView, LineChart, BarChart, BarChartIndicators, and two codemod utils). Those `any`s remain as on master.

## Test plan

- [x] `pnpm tsc:all` — 0 errors
- [x] `pnpm lint` — 0 errors (198 warnings; the increase over master's 184 is `any` usages kept per review, which `@typescript-eslint` v8 flags at warn level)
- [x] `pnpm test:unit` — 307 suites / 1292 tests / 274 snapshots pass
- [x] Alpha publish — succeeds (verified)
- [ ] `pnpm test:integration` (Happo + Cypress) — needs CI
- [ ] Verify storybook builds clean — needs CI

## Follow-up

~~Once [toptal/davinci#2677](https://github.com/toptal/davinci/pull/2677) merges and the alpha tags promote to stable, swap the five pinned alpha versions in root `package.json` to caret stable ranges.~~ Done — pinned to the stable releases.

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[FF-125]: https://toptal-core.atlassian.net/browse/FF-125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/shared/src/types.ts
```diff
@@ -48,18 +46,22 @@ export type OmitInternalProps<T, K = ''> = Pick<
...
 }
+
+// Strict on declared props, permissive on extras. Declared `P` fields are
+// type-checked at call sites (e.g. `<Button size='wrong'>` still errors).
+// Any other prop is accepted untyped — this is what makes the polymorphic
+// `as` usage continue to work without a generic call signature, and what
+// allows `forwardRef<HTMLElement, Props>(...)` to assign directly. The
+// trade-off versus the OLD shape: TS no longer infers prop types FROM the
+// `as` target (`<Button as={Link} to={...} />` won't validate `to` against
+// Link's props). Full polymorphic-inheritance typing is tracked in FF-125.
```
denieler: if maybe we can add here more details from FF-125? Picasso is open source project, but Jira is not available to everyone, so I think it's important to keep the context

pnpm-workspace.yaml
```diff
@@ -10,3 +16,6 @@ minimumReleaseAge: 10080
 minimumReleaseAgeExclude:
   - '@toptal/*'
   - '@topkit/*'
+  - '@playwright/*'
+  - 'playwright'
+  - 'playwright-core'
```
denieler: is this necessary? I don't think we even use playwright in Picasso

packages/picasso-forms/src/Form/Form.tsx
```diff
@@ -25,10 +25,10 @@ export type Props<T = AnyObject> = FinalFormProps<T> & {
   'data-testid'?: string
 }
 
-const getValidationErrors = (
+const getValidationErrors = <T extends AnyObject>(
   validators: Validators,
-  formValues: any,
-  form: FormApi<any>
+  formValues: T,
+  form: FormApi<T>
```
denieler: I like stricter types, but do we need to do them in this PR? if not, I'd suggest to keep old ones

packages/picasso-codemod/src/v40.0.0/utils/get-node-for-size-string-constant.ts
```diff
@@ -9,7 +11,7 @@ export const SIZE_TO_SPACING_CONSTANT: Record<string, string> = {
 }
 
 export const getNodeForSizeStringConstant = (
-  node: any,
+  node: StringLiteral,
```
denieler: wasn't `any` here good enough? it's just codemod

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -74,7 +74,15 @@ export interface Props<K extends string | number | symbol>
...
   maxBarSize?: number
-  /** Function to format ticks of the value axis */
+  /**
+   * Function to format ticks of the value axis. Matches Recharts' own
+   * `TickFormatter` type, which intentionally accepts `any` because the
+   * runtime value depends on the axis kind (category / number / time) —
+   * narrowing here would force consumer callbacks declared with a
+   * concrete parameter type (e.g. `(value: number) => …`) to widen
+   * unnecessarily under `strictFunctionTypes`.
+   */
```
denieler: I don't think we should change prop description

packages/base/Container/src/Container/styles.ts
```diff
@@ -1,33 +1,26 @@
-const alignItemsVariants = [
-  'flex-start',
-  'flex-end',
-  'center',
-  'stretch',
-  'baseline',
-] as const
+export type AlignItemsType =
```
denieler, line 1: old types looked just fine to me 🤔 is it required to change them?

pnpm-workspace.yaml
dulishkovych: Applied blindly by @vedrani suggestion. he uses the same in his PRs, so seems like that will be needed for orchestrator anyway https://toptal-core.slack.com/archives/C0AGRJLURAR/p1778661869055969

packages/base/Container/src/Container/styles.ts
dulishkovych: Yeah, this one's required. The old version keeps those values as `as const` arrays but only ever reads them back as types (`typeof alignItemsVariants[number]`). typescript-eslint v8 now flags that under `no-unused-vars` ("assigned a value but only used as a type"), and it's an error, not a warning, so the build won't pass with the arrays in place. Declaring the unions directly drops the now-unused runtime arrays. The exported types (`AlignItemsType`, `JustifyContentType`, `VariantType`) and their members are unchanged, so nothing downstream is affected.

pnpm-workspace.yaml
denieler: I'd let @vedrani to add this in his PR if necessary, but keeping this PR clean

packages/base/Button/src/ButtonAction/ButtonAction.tsx
```diff
@@ -22,7 +22,7 @@ const getIcon = ({
   iconPosition?: IconPositionType
 }) => {
   if (!icon) {
-    return null
+    return undefined
```
denieler, line 25: seems not needed change

packages/base/List/src/List/utils/generateListItems.tsx
```diff
@@ -1,9 +1,12 @@
+import type { ComponentProps } from 'react'
 import React from 'react'
 
 import List from '../List'
 
+type ListItemProps = ComponentProps<typeof List.Item>
+
 /* eslint-disable react/no-array-index-key */
-const generateListItems = (total: number, listItemProps?: any) =>
+const generateListItems = (total: number, listItemProps?: ListItemProps) =>
```
denieler: should we keep them as `any` for now?

packages/base/Select/src/SelectBase/utils/fire-on-change-event/fire-on-change-event.ts
```diff
@@ -1,18 +1,36 @@
...
-  event: any
+  event: SyntheticEvent
   value: ValueType | ValueType[]
   name?: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   onChange?: (event: any) => void
 }) => {
   event.persist()
-  event.target = { value: eventValue, name }
+  ;(event as unknown as { target: SelectEventTarget }).target = {
```
denieler: strange type casting here

packages/base/TreeView/src/TreeView/use-zoom.ts
```diff
@@ -46,8 +46,8 @@ export const useZoom = <ZoomRefElement extends ZoomedElementBaseType>({
 
     const transformContainer = select(rootRef.current.firstElementChild)
 
-    zoom.on('zoom', (event: any) => {
-      transformContainer.attr('transform', event.transform)
+    zoom.on('zoom', (event: D3ZoomEvent<ZoomRefElement, unknown>) => {
```
denieler: can we keep `any` here?

packages/picasso-charts/src/BarChart/BarChart.tsx
```diff
@@ -75,6 +75,7 @@ export interface Props<K extends string | number | symbol>
   /** Maximum size for the bar */
   maxBarSize?: number
   /** Function to format ticks of the value axis */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
denieler: if we don't change anything this comment is not needed

packages/picasso-charts/src/BarChartIndicators/BarChartIndicators.tsx
```diff
@@ -3,44 +3,69 @@ import type { ReactNode } from 'react'
...
+export interface RenderIndicatorParams {
+  dataKey: string
+  index: number
+  dataItem: BarSeriesItem
+}
+
 export type Props = {
-  renderIndicator: (params: any) => ReactNode
-  formattedGraphicalItems?: any
+  renderIndicator: (params: RenderIndicatorParams) => ReactNode
```
denieler: ☝️

packages/picasso-charts/src/LineChart/LineChart.tsx
```diff
@@ -297,9 +297,11 @@ export const LineChart = ({
                 allowTooltipEscapeViewBox ? positionOverride : undefined
               }
               content={customTooltip}
-              ref={(node: any) =>
-                (tooltipRef.current = node?.wrapperNode || null)
-              }
+              ref={node => {
+                tooltipRef.current =
+                  (node as { wrapperNode?: HTMLDivElement | null } | null)
```
denieler: ?

packages/picasso-codemod/src/v40.0.0/utils/get-updated-node.ts
```diff
@@ -23,11 +25,14 @@ export const getUpdatedNode = (
   } else if (node.type === 'NumericLiteral') {
     node = getNodeForNumber(node, options)
   } else if (node.type === 'ObjectExpression') {
-    const updatedProperties = node.properties.map((property: any) => {
-      property.value = getUpdatedNode(property.value, options)
+    const updatedProperties = (node.properties as Property[]).map(property => {
```
denieler: ?

packages/base/Button/src/ButtonAction/ButtonAction.tsx
dulishkovych: This one's actually required. The new `OverridableComponent` shape makes declared props strict, so `ButtonBase`'s `icon: ReactElement | undefined` no longer accepts the `null` that `getIcon` returned. Reverting it gives `TS2322` at the `<ButtonBase icon={...} />` call. Switching the helper to return `undefined` is the minimal fix and matches the declared type.

denieler: 🤔 but this is `ButtonAction` and I don't see any changes in `ButtonBase` where we have similar function

packages/picasso-codemod/src/v18.0.0/picasso-lab/utils.ts
```diff
@@ -29,11 +42,11 @@ const warnUsersAbout = (
   }
 }
 
-const getSpecifierName = (specifier: any) => {
+const getSpecifierName = (specifier: Specifier) => {
```
denieler: I don't mind, as it's codemod, but seems not needed change

packages/picasso-codemod/src/v40.0.0/utils/get-node-for-number.ts
```diff
@@ -13,7 +15,7 @@ export const NUMERIC_VALUE_TO_SPACING_CONSTANTS = [
 ]
 
 export const getNodeForNumber = (
-  node: any,
+  node: NumericLiteral,
```
denieler: same, seems not needed

packages/picasso-forms/src/Field/Field.tsx
```diff
@@ -32,6 +33,13 @@ export type Props<
     type?: string
     label?: React.ReactNode
     status?: OutlinedInputStatus
+    // The runtime child props are a structural mix of the wrapped component's
+    // props plus final-form `input` bindings plus Field-synthesized metadata
+    // (e.g. `highlight: 'autofill'`). Static typing as TWrappedComponentProps
+    // breaks consumers that destructure the synthesized metadata; typing the
+    // synthesized metadata structurally breaks consumers that bind their
+    // signature to TWrappedComponentProps directly. The boundary is unsound
+    // by design — see PF-2031 for the upgrade context.
```
denieler: it seems we didn't change anything here, so maybe this comment is not needed?

packages/picasso-forms/src/Field/Field.tsx
```diff
@@ -131,8 +144,7 @@ const Field = <
     status,
     ...rest,
     ...input,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    onChange: (event: ChangeEvent<HTMLElement> | any) => {
```
denieler, line 135: do we need to remove this any?

packages/picasso-forms/src/Field/Field.tsx
```diff
@@ -40,14 +48,19 @@ export type Props<
...
-    return composeValidators([requiredValidator, validate])
+    return composeValidators([
+      requiredValidator as FieldValidator<TValue>,
+      validate,
+    ])
   }
 
   if (required && !validate) {
-    return requiredValidator
+    return requiredValidator as FieldValidator<TValue>
```
denieler: this looks not needed 🤔 same as another ☝️ type casting above in this file

pnpm-workspace.yaml
```diff
@@ -2,6 +2,12 @@ packages:
   - 'packages/*'
   - 'packages/base/*'
 
+nodeLinker: hoisted
+strictPeerDependencies: false
+autoInstallPeers: true
+linkWorkspacePackages: true
+resolvePeersFromWorkspaceRoot: true
+
```
denieler, line 10: why do we need those here? 🤔

.changeset/typescript-5-5-upgrade.md
```diff
@@ -0,0 +1,32 @@
+---
+'@toptal/picasso': major
+'@toptal/picasso-charts': major
+'@toptal/picasso-codemod': major
+'@toptal/picasso-forms': major
+'@toptal/picasso-pictograms': major
+'@toptal/picasso-provider': major
+'@toptal/picasso-query-builder': major
+'@toptal/picasso-rich-text-editor': major
```
denieler, line 9: I think after changes those are not major anymore

pnpm-workspace.yaml
dulishkovych: This is exactly the thing that @vedrani mentioned to host in my PR [here](https://toptal-core.slack.com/archives/C0AGRJLURAR/p1778661869055969)

.changeset/typescript-5-5-upgrade.md
dulishkovych: All of these packages have TS peer dep bumped so I think they still need major

pnpm-workspace.yaml
denieler: maybe after all the changes we did this fix is not needed anymore?

vedrani: It’s ok. Pnpm actually requires it.

denieler: @vedrani but it worked without it 🤔

vedrani: Its needed because of worktrees

denieler: @vedrani can you please explain a bit more? 🙇‍♂️

packages/base/Button/src/ButtonAction/ButtonAction.tsx
dulishkovych: They look similar, but the results are used differently. In `ButtonBase` the icon goes into the children, and children are allowed to be `null`. In `ButtonAction` it's passed as the `icon` prop, and that prop only accepts an element or nothing (`undefined`), not `null`.

It compiled before because the old `OverridableComponent` type was loose about props, so the `null` slipped through. The new type checks each declared prop strictly, so the `null` now gets flagged. Returning `undefined` instead of `null` matches what the prop expects.

pnpm-workspace.yaml
vedrani: I had problems with git worktrees when I rebased on top of pnpm migration. One of things that happen was that settings in npmrc were ignored. I would run pnpm and it would start behaving different for internal workspace packages not linking them or not hoisting. 

I even run debug to return pnpm config and it was null for config of pnpm. If I remember correctly it was because when NPM_TOKEN is not set, rest gets ignored by pnpm. 

What worked for me was moving pnpm settings to pnpm-workspace (it's also required for pnpm 11 so it's good for future) and also using `pnpm clean` command to reset state.

Maybe it worked for you in local branch, but you didn't test in worktrees.

I don't think it's happening all the time, but if NPM_TOKEN is not set it breaks pnpm settings and then they pollute store, lock file etc. Having pnpm settings in separate file is much stable and cleaner. Also considering pnpm 11 requires I don't see damage to add it.

vedrani (approved):

packages/base/Button/src/ButtonAction/ButtonAction.tsx
dulishkovych: @denieler aligned all `getIcon` functions to return `ReactElement | undefined`
https://github.com/toptal/picasso/pull/4963/commits/ec76d49526d694df14ff82fd42f51cd2a9d86d3d

PR #4966, opened 2026-05-18
[PF-1994] Migrate Drawer to @base-ui/react + Tailwind

vedrani:
# Drawer migration diff

Generated: 2026-05-18 15:37:56 CEST

**Package:** `packages/base/Drawer`

## Files

## Imports

**Removed:**

```
packages/base/Drawer/src/Drawer/Drawer.tsx:import React, { useRef } from 'react'
packages/base/Drawer/src/Drawer/Drawer.tsx:import { Backdrop } from '@toptal/picasso-backdrop'
packages/base/Drawer/src/Drawer/Drawer.tsx:import { Modal } from '@mui/base/Modal'
```

**Added:**

```
packages/base/Drawer/src/Drawer/Drawer.tsx:import React from 'react'
packages/base/Drawer/src/Drawer/Drawer.tsx:import { Drawer as BaseUIDrawer } from '@base-ui/react/drawer'
```

## MUI v4 / JSS residue check

| Check | Count |
|---|---|
0 |

## package.json delta

```diff
@@ -22,9 +22,8 @@
   },
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "dependencies": {
-    "@mui/base": "5.0.0-beta.58",
+    "@base-ui/react": "^1.4.1",
     "@toptal/picasso-slide": "1.0.4",
-    "@toptal/picasso-backdrop": "2.0.0",
     "@toptal/picasso-paper": "4.0.5",
     "@toptal/picasso-button": "5.0.0",
     "@toptal/picasso-container": "3.1.4",
@@ -37,11 +36,10 @@
     "**/styles.js"
   ],
   "peerDependencies": {
-    "@material-ui/core": "4.12.4",
     "@toptal/picasso-provider": "*",
     "@toptal/picasso-tailwind": ">=2.7",
     "@toptal/picasso-tailwind-merge": "^2.0.0",
-    "react": ">=16.12.0 < 19.0.0"
+    "react": ">=16.12.0"
   },
   "exports": {
     ".": "./dist-package/src/index.js"
```

## Prop-surface diff

```diff
```

</details>

## Happo
Happo log: `migration-runs/2026-05-18/Drawer/happo.log` (0
? flagged lines).

## React 19 smoke


packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -80,83 +87,106 @@ export const Drawer = ({
...
   const { setHasDrawer } = useDrawer()
   const container = usePicassoRoot()
-  const ref = useRef<HTMLDivElement>(null)
 
   usePageScrollLock(Boolean(maintainBodyScrollLock && open))
 
   useIsomorphicLayoutEffect(() => {
     setHasDrawer(open)
 
-    const cleanup = () => {
```
denieler, line 90: I like how it was before, because it was explicit to the reader that this is cleanup and not an anonymous function in the codebase

packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -80,83 +87,106 @@ export const Drawer = ({
...
-            <Container flex className='flex-1'>
-              {children}
-            </Container>
-            <ButtonCircular
-              variant='flat'
-              icon={<CloseMinor16 />}
-              onClick={handleOnClose}
-              className='absolute right-6 top-4'
-              aria-label='Close drawer'
+      <BaseUIDrawer.Portal container={container ?? undefined} keepMounted>
```
denieler: it seems it's possible to remain `disablePortal` prop for the Drawer component, I'd suggest to keep API unchanged, because any breaking changes or any changes in the API would create huge problems in migration later

patches/@base-ui__utils@0.2.8.patch
```diff
@@ -0,0 +1,52 @@
...
++export type CreateSelectorFunction = <Args extends any[], Selectors extends ReadonlyArray<Selector<any>>, Combiner extends (...args: readonly [...ReturnTypes<Selectors>, ...Args]) => any>(...items: [...Selectors, NoOptionalParams<Combiner>]) => (...args: Selectors['length'] extends 0 ? MergeParams<ReturnTypes<Selectors>, Parameters<Combiner>> : [StateFromSelectorList<Selectors>, ...MergeParams<ReturnTypes<Selectors>, Parameters<Combiner>>]) => ReturnType<Combiner>;
+ type StateFromSelectorList<Selectors extends readonly any[]> = Selectors extends [f: infer F, ...other: infer R] ? StateFromSelector<F> extends StateFromSelectorList<R> ? StateFromSelector<F> : StateFromSelectorList<R> : {};
+ type StateFromSelector<T> = T extends ((first: infer F, ...args: any[]) => any) ? F : never;
+ type DropFirst<T> = T extends [any, ...infer Xs] ? Xs : [];
```
denieler: we don't want any patch for `base-ui` packages

packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -80,83 +87,106 @@ export const Drawer = ({
...
+            !disableBackdrop && 'fixed inset-0',
+            className
+          )}
+          style={style}
+        >
+          {!disableBackdrop && (
+            <BaseUIDrawer.Backdrop
+              className={twMerge(
+                'fixed -z-[1] inset-0',
+                '-webkit-tap-highlight-color-transparent',
```
denieler: this looks like a style that is working only for webkit, but for other engines it seems we don't have corresponding style. In this repo we provide support for all the modern browsers in our components

patches/@base-ui__utils@0.2.8.patch
```diff
@@ -0,0 +1,52 @@
+diff --git a/esm/store/ReactStore.d.ts b/esm/store/ReactStore.d.ts
```
vedrani: Why did you add this patch? Why it's needed and can we avoid pathing it?

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — restored the named `cleanup` const so the intent reads more clearly than an anonymous return. See latest commit.

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — removed the `-webkit-tap-highlight-color-transparent` class. You're right that it's webkit-only, but the bigger issue is it's not valid Tailwind syntax (real arbitrary-property syntax would be `[-webkit-tap-highlight-color:transparent]`), so it emits no CSS at all here or in the legacy `Backdrop` it was copied from. Since nothing actually depends on the tap-highlight suppression today, dropping it cleanly is safer than introducing a one-engine style. Happy to revisit with the bracketed arbitrary-property syntax if a designer flags the iOS gray-flash regression.

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Considering this. You're right that breaking the public API mid-migration is exactly what we want to avoid — I marked `disablePortal` deprecated/no-op which is still a behavior change for consumers passing `disablePortal` today.

`@base-ui/react`'s `Drawer` doesn't expose a `disablePortal` prop, but `Drawer.Portal` is just a wrapper around `ReactDOM.createPortal`. The cleanest way to preserve the prop is to conditionally skip the `Portal` and render `Viewport`/`Popup` in place:

```tsx
const PortalWrapper = disablePortal ? React.Fragment : BaseUIDrawer.Portal
const portalProps = disablePortal
  ? {}
  : { container: container ?? undefined, keepMounted: true }

<PortalWrapper {...portalProps}>
  <BaseUIDrawer.Viewport ...>
    ...
  </BaseUIDrawer.Viewport>
</PortalWrapper>
```

Trade-off: `disablePortal` was effectively unused in Picasso's own tree (verified — 0 internal callsites pass it), but external consumers might. Preserving the prop keeps the API honest rather than silently dropping behavior.

Apply this fix? 👍 to confirm, or share thoughts if you'd rather take a different shape (e.g. keep deprecated/no-op but with a runtime warning when consumers pass it).

patches/@base-ui__utils@0.2.8.patch
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

The patch is needed because Picasso pins `typescript: ~4.7.0` (root `package.json:317`), but `@base-ui/utils@0.2.8`'s shipped `.d.ts` files use the `const` modifier on type parameters — a **TS 5.0+ syntax feature**. Specifically:

```ts
// @base-ui/utils/store/ReactStore.d.ts (excerpt)
useStateSetter<const Key extends keyof State, Value extends State[Key]>(...)
//             ^^^^^ requires TS 5.0
```

`skipLibCheck: true` (which we already have, `tsconfig.base.json:8`) only skips **type-checking** of `.d.ts` files — the parser still has to parse them, and TS 4.7's parser rejects `const` type parameters with a syntax error. So even with `skipLibCheck`, `tsc --noEmit` fails on every package that transitively pulls in `@base-ui/utils` (which is anything using `@base-ui/react`).

The patch strips the `const` modifier in two `.d.ts` files (`ReactStore.d.ts` + `createSelector.d.ts`, both at the `esm/` and `store/` entry points — 4 occurrences total). It's purely declaration-file syntax stripping — no runtime change, no type-correctness change beyond TS 4.7's narrower inference.

Avoiding the patch would mean one of:
1. **Upgrade TypeScript to 5.0+** (cleanest, but out of scope for PF-1994 — affects all 28 components and ~80 packages)
2. **Vendor type shims for `@base-ui/utils`** in `@types/` overriding the offending types — fragile, drifts with each base-ui release
3. **Wait for the upstream TypeScript upgrade ticket** before continuing the @base-ui migration

See denieler's parallel comment at `patches/@base-ui__utils@0.2.8.patch:52` — they're asking we drop the patch. My read is option 1 is the only durable answer, and the patch is the bridge until then. Happy to file a follow-up ticket for the TypeScript bump and link it from a comment in the patch file. Thoughts?

patches/@base-ui__utils@0.2.8.patch
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Considering this. Agreed in principle — patches against ecosystem packages are a maintenance burden and silently drift on upgrade.

The blocker is concrete: Picasso pins `typescript: ~4.7.0`, and `@base-ui/utils@0.2.8`'s `.d.ts` files use `const`-modifier type parameters (a TS 5.0+ syntax feature). Even with `skipLibCheck: true`, the TS 4.7 **parser** rejects the syntax — so `tsc --noEmit` fails on any package transitively importing `@base-ui/react`. The patch only strips the `const` keyword in 4 places — declaration-file syntax adjustment, no runtime/type-semantic change.

Proposal: keep the patch as a temporary bridge with an inline comment pointing to a follow-up ticket for the TypeScript 5.x upgrade. Once TS is bumped, the patch goes away.

Alternative shapes if you'd prefer:
1. **Vendor type shims** under `@types/@base-ui/utils/` — fragile, has to be redone on every base-ui version bump.
2. **Block the @base-ui migration on the TS upgrade** — but that delays PF-1994 by an unknown amount.
3. **Drop the patch and ship anyway** — won't pass typecheck CI; non-starter.

Apply the "keep with comment + follow-up ticket" path? 👍 to confirm, or let me know which alternative you'd prefer. (I posted a longer technical explanation in response to @vedrani's question on patch line 1.)

.changeset/drawer-migration.md
```diff
@@ -0,0 +1,11 @@
...
+'@toptal/picasso-drawer': major
+---
+
+### Drawer
+
+- Migrate internals from @mui/base/Modal to @base-ui/react/drawer compound parts (`Drawer.Root` + `Drawer.Portal` + `Drawer.Viewport` + `Drawer.Backdrop` + `Drawer.Popup`)
+- Inline the backdrop as `Drawer.Backdrop` with Tailwind classes preserving the prior visual (drops the runtime dependency on `@toptal/picasso-backdrop`)
+- Drop `@material-ui/core` from `peerDependencies` and `@mui/base` from `dependencies`; lift the `react` peer-dependency upper bound to `>=16.12.0` (drop the `< 19.0.0` cap)
+- Gain swipe-to-dismiss gestures from `@base-ui/react/drawer` keyed to the `anchor` prop (new behavior; no existing API surface to disable it)
+- `disablePortal` is now a no-op kept for type compatibility — `@base-ui/react/drawer` always portals
```
dulishkovych: This is a breaking change. And there are some internal consumers that will be affected by this change https://github.com/search?q=org%3Atoptal%20disablePortal&type=code

Not talking about all external consumers who might be using Picasso that we don't know about.

packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -94,69 +100,94 @@ export const Drawer = ({
...
+    if (!open || typeof document === 'undefined') {
+      return
     }
-  }
+    const active = document.activeElement
+
+    if (active && active !== document.body && 'blur' in active) {
+      ;(active as HTMLElement).blur()
+    }
+  }, [open])
```
dulishkovych: This looks too hacky for me. I think we need to assess why we want to keep previous approach. Is it because it was intended like that by the design patterns to have this open behavior or was it just a library default behavior that we try now to migrate to new library that does not support the same behavior. Can you think of that? Or maybe there are other options how to achive the same result without such hacks?

.changeset/drawer-migration.md
```diff
@@ -0,0 +1,11 @@
+---
+'@toptal/picasso-drawer': major
+---
+
+### Drawer
+
+- Migrate internals from @mui/base/Modal to @base-ui/react/drawer compound parts (`Drawer.Root` + `Drawer.Portal` + `Drawer.Viewport` + `Drawer.Backdrop` + `Drawer.Popup`)
+- Inline the backdrop as `Drawer.Backdrop` with Tailwind classes preserving the prior visual (drops the runtime dependency on `@toptal/picasso-backdrop`)
+- Drop `@material-ui/core` from `peerDependencies` and `@mui/base` from `dependencies`; lift the `react` peer-dependency upper bound to `>=16.12.0` (drop the `< 19.0.0` cap)
+- Gain swipe-to-dismiss gestures from `@base-ui/react/drawer` keyed to the `anchor` prop (new behavior; no existing API surface to disable it)
```
dulishkovych: The migration adds swipe-to-dismiss with no way to turn it off. Was that an intentional product call, or just what `@base-ui/react/drawer` does by default? I ask because it's new behavior for every consumer, and a swipe that dismisses the drawer could collide with scrollable or draggable content inside it.

packages/base/Drawer/src/Drawer/Drawer.tsx
denieler: But this component provides `<BaseUIDrawer.Portal container={` so, we can emulate `disablePortal` effect (by rendering `in-place` or in desired container)

patches/@base-ui__utils@0.2.8.patch
denieler: I understand, but I think it's better to ignore type check inline vs introducing a patch to the codebase, I think patch is the worst possible solution

vedrani: Yes, but we need to not block migration and TS upgrade PR is in review so later we can remove patch after we rebase. I think shims are also ok, but imo we will clean it soon so it doesn't matter much.

.changeset/drawer-migration.md
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — `disablePortal` now works: when `true`, the Portal wrapper is skipped and the viewport renders inline. Changeset updated to reflect the preserved prop.

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Removed — it was compensating for `@base-ui/react`'s rAF-deferred focus (vs `@mui/base`'s synchronous move), but that transient state isn't visible in static Happo snapshots. Defensive-only code, and you're right it's a hack.

.changeset/drawer-migration.md
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

It's `@base-ui/react/drawer`'s default — swipe fires whenever `swipeDirection` is set (currently always, keyed to `anchor`). We can pass `swipeDirection={undefined}` when a new `disableSwipeToDismiss` prop is true. Is that worth adding, or should we accept swipe as the new default behavior?

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — implemented exactly this: `disablePortal=true` skips the Portal wrapper; `false` (default) portals via `BaseUIDrawer.Portal container={container}`.

patches/@base-ui__utils@0.2.8.patch
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Acknowledged. vedrani has decided to keep the patch as a bridge until the TS 5.x upgrade lands (see thread above). Patch goes away on that rebase.

dulishkovych:
I've tested manually in storybook and current version of drawer does not have the opening and closing animation that mui base drawer had. That is a regression. The new drawer opens by just appearing on the screen.

patches/@base-ui__utils@0.2.8.patch
denieler: please just put ignore to lines that got broken in typecheck, instead of the patch

packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -95,68 +100,82 @@ export const Drawer = ({
...
-      ref={ref}
+  const viewport = (
+    <BaseUIDrawer.Viewport
       className={twMerge(
-        className,
-        'z-drawer inset-0',
-        !disableBackdrop && 'fixed'
+        'z-drawer',
+        !disableBackdrop && 'fixed inset-0',
+        className
```
denieler, line 123: the order matters here

vedrani:
Fixed in this commit. The root cause: `@base-ui/react`'s `useAnimationsFinished` calls `element.getAnimations()` on the **Popup element only** (no `{subtree:true}`), so CSS transitions on `DrawerPaper` (a child of Popup) were invisible to it. With no detected transition, `mounted` became `false` immediately → `display:none` → the Slide exit animation was killed before it played. Same issue on enter.

The fix replaces `react-transition-group/Slide` with @base-ui/react's native transition mechanism: `transition-transform` + `data-[starting-style]:translate-x-full` / `data-[ending-style]:translate-x-full` applied directly on `BaseUIDrawer.Popup`. @base-ui/react adds `data-starting-style` on mount (Popup off-screen), removes it to trigger the enter slide-in, then adds `data-ending-style` before unmounting to trigger the exit slide-out — both directions now animate correctly.

Positioning (`fixed`, `top-0`, `h-full`, anchor-side) and `transition-transform` moved from `DrawerPaper` to `Popup`. `DrawerPaper` is now a pure surface element (`h-full flex flex-col` + Paper elevation). `@toptal/picasso-slide` dep removed from package.json.

patches/@base-ui__utils@0.2.8.patch
vedrani: Understood — I can switch to `@ts-ignore` inline. One thing to flag: the patch covers `@base-ui/utils@0.2.8` which is a transitive dep of `@base-ui/react`, and it's shared by all four migrated components: **Button, Slider, Switch, and Drawer**. The `const T` generic syntax in `@base-ui/utils` fails under TypeScript ~4.7 (our pinned version). Moving to `@ts-ignore` would mean adding suppressions to every file in those four packages that import the affected types.

Is a TypeScript 5.x upgrade already planned before these land? If so, `@ts-ignore` is cleaner and I'll add it across all four packages. If not, I want to make sure we agree on the scope before switching. 👍 to confirm, and I'll make the change.

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: Could you clarify the intent? The old `@mui/base` code always applied `inset-0` to the modal root and only made `fixed` conditional:

```tsx
twMerge(className, 'z-drawer inset-0', !disableBackdrop && 'fixed')
```

The current code makes both `fixed` and `inset-0` conditional together. Two possible interpretations:

1. **`inset-0` should always be applied** (matching old behavior), only `fixed` conditional — i.e. `twMerge('z-drawer inset-0', !disableBackdrop && 'fixed', className)`
2. **`className` override order** — `className` is currently last (wins over base classes), old code had it first (base classes win)

Which did you have in mind?

patches/@base-ui__utils@0.2.8.patch
vedrani: Update — I dug in further and **`@ts-ignore` is not a viable substitute here**:

The failure is a TypeScript **parse error** inside `node_modules/@base-ui/utils/.../{ReactStore,createSelector}.d.ts` (the `const T` generic-parameter syntax was added in TS 5.0; we're on TS ~4.7). Two facts make `@ts-ignore` unworkable:

1. **`skipLibCheck: true` only skips type checking of declaration files — it does not skip parsing.** I confirmed this empirically: removed the patch, restored `const T` in the .d.ts files, ran `pnpm --filter @toptal/picasso-drawer build:package` → 7 parse errors in `@base-ui/utils`:
   ```
   node_modules/@base-ui/utils/store/ReactStore.d.ts(73,...): error TS1005: '?' expected.
   ... TS1005, TS1109, TS1128 ...
   ```

2. **`// @ts-ignore` suppresses errors on the next line of *your own source*, not parse errors in third-party `.d.ts` files you don't author.** The errors here surface in `node_modules`, not at our import sites.

The patch is the minimal correct fix until the repo upgrades to TS 5.x (at which point it can be removed alongside any equivalent patches shared by Button, Slider, Switch).

I've kept the patch in place. Let me know if you'd prefer a different mitigation (e.g. `@base-ui/utils` overridden via `paths` mapping, or a vendored type-shim) — open to alternatives, but `@ts-ignore` specifically won't work for parse errors.

packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -95,45 +105,49 @@ export const Drawer = ({
...
+      <BaseUIDrawer.Popup
+        data-testid={testId}
+        data-private={dataPrivate}
+        style={{
+          transitionDuration: `${
+            typeof transitionProps?.timeout === 'number'
+              ? transitionProps.timeout
+              : 300
+          }ms`,
+        }}
```
dulishkovych: The popup's transition duration is driven by `transitionProps.timeout` (falling back to 300ms), but the backdrop fade is hardcoded to `duration-300`. 

So if a consumer passes a custom `timeout`, the panel slides at that duration while the backdrop keeps fading at 300ms, and the two finish at different times. The old `Slide`-based setup didn't have that split. Can we feed the same timeout into the backdrop so they stay in sync, or was the fixed 300ms intentional?

.changeset/drawer-migration.md
dulishkovych: Thanks for adding the opt-out. Checked the base-ui source, swipe is touch-only and already handles scroll/text-selection, so my earlier worries don't really apply. Just the default I'm unsure about. Old drawer had no swipe and this migration is supposed to keep behavior the same, so can we default it off here and add swipe later as its own thing? Main case I'd want tested before defaulting on is a top/bottom drawer with scrollable content, where the swipe and scroll go the same direction.

patches/@base-ui__utils@0.2.8.patch
vedrani: @denieler let's keep patch and focus on merging TS PR.

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — extracted the timeout into a single `transitionDurationStyle` and applied it to both `Drawer.Backdrop` and `Drawer.Popup`. Backdrop now fades in lockstep with the panel slide; replaced `duration-300` with the inline style.

.changeset/drawer-migration.md
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — flipped `disableSwipeToDismiss` default to `true` so swipe is off by default (parity with the old Drawer). Updated changeset to mark it as opt-in. Top/bottom scrollable-content test deferred to the follow-up that actually enables swipe.

vedrani:

Happo unrelated flake (Drawer migration is unrelated; treating as designer-review):

- CategoriesChart / default/with-column-hover — column-hover state pixels in middle-left region shifted

Other 3 Drawer diffs (drawer-{narrow,regular,ultra-wide}-width/479-default) were regressions on this PR — fixed in latest commit by capping `Drawer.Popup` width with `max-w-full` so the @base-ui/react popup respects the viewport CB (the old DrawerPaper was the positioned element with that cap; now the cap moved up the chain to Popup).

vedrani:
> 🤖 _Orchestrator agent (autonomous standards-audit)_

A stray empty `.tmp-payload.json` was committed at repo root (in `e5db78e3f`) — scratch-file leak per practices.md §"Verify before commit". Should be removed before merge (sandbox blocked me from deleting it autonomously).

vedrani:
> 🤖 _Orchestrator agent (autonomous standards-audit)_

Removed a stale `{ "path": "../Slide" }` reference in `packages/base/Drawer/tsconfig.json`. The `@toptal/picasso-slide` dep was dropped from package.json (animation is now native @base-ui/react CSS) but the tsconfig `references` entry remained — they must agree per `practices.md §tsconfig & build hygiene`. `tsc -b` verified green.

vedrani:
# 🛑 Orchestrator escalation — `Drawer`

**Trigger:** migrate-loop stuck on deterministic gate stages: audit: (identical content across 3 consecutive iters incl. stuck-recovery attempt). Worktree left dirty for operator inspection.
**Iterations:** 4 / 3
**PR:** https://github.com/toptal/picasso/pull/4966
**Worktree:** `migration-runs/2026-05-18/Drawer-v1/worktree`
**Last gate report:** `/Users/vivanac/Projects/picasso/migration-runs/2026-06-03/Drawer-v2/worktree/migration-runs/2026-06-03/Drawer/report.md`

PR #4999, opened 2026-06-10
[PF-1994] Migrate Radio to @base-ui/react + Tailwind

vedrani:
# Radio — MUI v4 + JSS → native input + Tailwind

## Summary

Migrates `packages/base/Radio` (Tier 2 heavy path) off `@material-ui/core` + JSS. `Radio` is re-implemented as a custom control — root `<span>` + visually-hidden native `<input type="radio">` + two icon `<span>`s whose circle/dot are Tailwind pseudo-elements; `Radio.Group` becomes a plain `div[role=radiogroup]` providing a Picasso-internal `RadioGroupContext` (`name`/`value`/`onChange`, with internal state for the uncontrolled `defaultValue` case), preserving MUI's group contract exactly. The 6 JSS slot keys plus the `PicassoProvider.override(MuiRadio)` block were property-parity-mapped to Tailwind (`group-hover:` tints, `duration-350 ease-in-out` transitions, legacy `transform`-based pseudo centering).

## Decisions

- **Native `<input type="radio">` instead of `@base-ui/react/radio`** (deviation from manifest `target_path` — documented in `docs/migration/components/Radio.md §"target_path deviation"` in this PR, operator review requested): Base UI's `Radio.Root` derives `checked` exclusively from `RadioGroup` context (`checked = groupContext ? checkedValue === value : value === ''`, `setCheckedValue = NOOP` standalone — see `node_modules/@base-ui/react/radio/root/RadioRoot.js`). Picasso's public contract requires standalone `checked`, uncontrolled click-to-check, per-item `onChange(event, checked)`, and external `<label htmlFor>` association — all inert or unreachable on a group-less Base UI radio. The manifest note ("@base-ui/react doesn't ship a RadioGroup primitive") is also outdated (v1.4.1 ships one), but the standalone-checked gap is the decisive blocker. Native input + own context preserves the full API; this matches the crib's "no analog → stay custom" path (Backdrop/Badge/FileInput precedent).
- **Checked visual is React-driven when `checked` is known** (controlled or group-derived); CSS `peer-checked:` only drives the uncontrolled case. Reason (Happo iter 1): pages with duplicate radio group names (Form story binds `gender` twice; QueryBuilder hardcodes `name='variableName'` per nested group) trigger native radio name-exclusivity that unchecks earlier same-name inputs — MUI's element-swapped icon survived that, a pure `:checked`-CSS visual does not.
- **Focus ring via `data-focused` attribute** (set from `event.target.matches(':focus-visible')` on focus, cleared on blur) styled with `group-data-[focused]:before:shadow-[…]`. Reason (Happo iter 1, Cypress strict): Cypress-Happo re-renders a *serialized* DOM where live pseudo-class state (`:focus-visible`) is lost; the old MUI focus-visible polyfill applied a real class that survived serialization. The attribute reproduces that, with `:focus-visible` itself as the modality heuristic (jsdom-safe: nwsapi 2.2.22 supports it).
- **`shrink-0` on the control root**: old `MuiIconButton-root` carried `flex: 0 0 auto`; without it the 16px control gets flex-shrunk inside tight labels (verified live: 15px in the "Radio group grid" overflow story → Happo text-region diff, iter 1).
- **Pseudo-element centering uses legacy `transform: translate(-50%,-50%)`** (arbitrary property) instead of Tailwind v4's `-translate-*` utilities, which emit the separate `translate:` CSS property. Computed-style diff (iter 3) showed this as the only paint-relevant divergence from baseline (`matrix(1,0,0,1,-8,-8)` vs `translate: -50% -50%`); the two rasterize through different compositing paths, the best-candidate source of the sub-threshold Cypress-Happo diff (0 px via pixelmatch). Verified locally: computed values now byte-match the baseline.
- **`classes` dropped** per `decisions/classes-audit.md` §4 / Tier 2 matrix: re-verified source (`extends StandardProps`, plumbing-only body reads), 0 internal callsites (multiline rg), 0 external (gh search code). `Omit<StandardProps, 'classes'>` + runtime `classes: _classes` backstop.
- **`RadioGroupProps` MUI type leak replaced** with a Picasso-native interface (`BaseProps` + div attributes + `name`/`value`/`defaultValue`/`onChange(event, value)`); MUI-inherited `classes`/`row` disappear → drives the `major` bump (manifest `versionBump: major`, taken verbatim).
- **Pre-existing quirks preserved**: `spacing ?? horizontal ? 16 : 0` precedence (consumer `spacing` value ignored beyond truthiness — existing-violations carve-out), `data-private` only forwarded in the labeled branch, `theme.spacing(2)` frozen as literal `16`.
- **Internal `RadioControl` sub-component** (new file, co-located): needed because `FormControlLabel` does `cloneElement(control, { disabled })` — a raw host `<span>` element would receive an invalid `disabled` DOM attribute. Element-variance between the public `ButtonOrAnchorProps` surface and the `<span>` root is bridged once at the boundary (`code-standards §"TS variance"` pattern).

## Limitations / Out-of-scope

- Old root's vestigial `color: white` / checked `color: primary` and `border-radius: 50%` (transparent bg, no border — zero paint effect) plus the disabled `boxShadow: none !important` belt-and-suspenders (disabled inputs are unfocusable + root is `pointer-events-none`) were consciously dropped.
- Hover tint `mix(blue-500, white, 0.16)` ≈ `#446AD7` has no canonical token → arbitrary value + `TODO(tokens)`.
- Icon swap on check is `display`-based (matches MUI's element-swap: no cross-state color animation; transitions still cover hover/in-state changes).
- `.changeset/radio-base-ui-migration.md` is an empty (no-op) changeset left over from a rename to the expected `radio-migration.md` filename — file deletion is approval-gated in this environment; safe to drop at merge.

## Verification

- **Local gate stages passed (iter 3)**: `pnpm --filter @toptal/picasso-radio build:package` ✓, `davinci-syntax lint --check` ✓ (one sanctioned `TODO(tokens)` warning), jest ✓ 20/20 across Radio + consumers (picasso-forms Radio/RadioGroup, ButtonRadio, query-builder CombinatorSelector — snapshots regenerated), Cypress `Radio.spec.tsx` 3/3 (iter 2).
- **Runtime check (Playwright, fresh dev server)**: computed-style diff staging-vs-local on root/input/icon/::before/::after — after the `transform` fix, zero paint-relevant differences remain (`transform: matrix(1,0,0,1,-8,-8)` / `matrix(1,0,0,1,-2,-2)`, `translate: none` — byte-identical to baseline). Screenshots: `local--forms-radio--radio.png`, `local--forms-radio--radio--focused.png`, `local--forms-radio--radio--iter3.png`.
- **Visual parity**: Storybook-Happo green since iter 2. The single Cypress-strict diff (`Radio/default/after-hovered/chrome-desktop`) has now persisted across iters 2–4 at **0 diff pixels** per the gate's own pixelmatch analyzer ("negligible — no source change needed"), surviving even the iter-3 `transform` fix that made computed styles byte-identical to baseline. Classified as environment-level rendering noise — recommend the small-residual / `needs-visual-signoff` path (one Happo accept) rather than further source iteration.

---

## Mechanical diff evidence

# Radio migration diff

Generated: 2026-06-10 10:00:56 CEST

**Package:** `packages/base/Radio`

## Files

```diff
+packages/base/Radio/src/Radio/RadioControl.tsx
+packages/base/Radio/src/RadioGroupContext/index.ts
```

## Imports

**Removed:**

```
packages/base/Radio/dist-package/src/Radio/styles.d.ts:import type { Theme } from '@material-ui/core/styles';
packages/base/Radio/dist-package/src/RadioGroup/RadioGroup.d.ts:import type { RadioGroupProps } from '@material-ui/core';
packages/base/Radio/src/Radio/Radio.tsx:import cx from 'classnames'
packages/base/Radio/src/Radio/Radio.tsx:import styles from './styles'
packages/base/Radio/src/Radio/Radio.tsx:import type { Theme } from '@material-ui/core/styles'
packages/base/Radio/src/Radio/Radio.tsx:import { Radio as MUIRadio } from '@material-ui/core'
packages/base/Radio/src/Radio/Radio.tsx:import { makeStyles } from '@material-ui/core/styles'
packages/base/Radio/src/Radio/styles.ts:import type { Theme } from '@material-ui/core/styles'
packages/base/Radio/src/Radio/styles.ts:import { PicassoProvider } from '@toptal/picasso-provider'
packages/base/Radio/src/Radio/styles.ts:import { createStyles } from '@material-ui/core/styles'
packages/base/Radio/src/Radio/styles.ts:import { rem, outline, mix } from '@toptal/picasso-shared'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import React from 'react'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import type { RadioGroupProps } from '@material-ui/core'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import { RadioGroup as MUIRadioGroup } from '@material-ui/core'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import { useTheme } from '@material-ui/core/styles'
```

**Added:**

```
packages/base/Radio/dist-package/src/Radio/RadioControl.d.ts:import React from 'react';
packages/base/Radio/dist-package/src/Radio/RadioControl.d.ts:import type { ChangeEventHandler, HTMLAttributes } from 'react';
packages/base/Radio/dist-package/src/RadioGroup/RadioGroup.d.ts:import type { BaseProps } from '@toptal/picasso-shared';
packages/base/Radio/dist-package/src/RadioGroup/RadioGroup.d.ts:import type { ChangeEvent, HTMLAttributes } from 'react';
packages/base/Radio/dist-package/src/RadioGroupContext/index.d.ts:import type { ChangeEvent } from 'react';
packages/base/Radio/src/Radio/Radio.tsx:import type { Props as RadioControlProps } from './RadioControl'
packages/base/Radio/src/Radio/Radio.tsx:import { RadioControl } from './RadioControl'
packages/base/Radio/src/Radio/Radio.tsx:import { useRadioGroupContext } from '../RadioGroupContext'
packages/base/Radio/src/Radio/RadioControl.tsx:import React, { forwardRef, useState } from 'react'
packages/base/Radio/src/Radio/RadioControl.tsx:import cx from 'classnames'
packages/base/Radio/src/Radio/RadioControl.tsx:import type { ChangeEventHandler, FocusEvent, HTMLAttributes } from 'react'
packages/base/Radio/src/Radio/RadioControl.tsx:import { twMerge } from '@toptal/picasso-tailwind-merge'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import React, { useState } from 'react'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import type { BaseProps } from '@toptal/picasso-shared'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import type { ChangeEvent, HTMLAttributes } from 'react'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import { RadioGroupContext } from '../RadioGroupContext'
packages/base/Radio/src/RadioGroup/RadioGroup.tsx:import { twMerge } from '@toptal/picasso-tailwind-merge'
packages/base/Radio/src/RadioGroupContext/index.ts:import type { ChangeEvent } from 'react'
packages/base/Radio/src/RadioGroupContext/index.ts:import { createContext, useContext } from 'react'
```

## MUI v4 / JSS residue check

| Check | Count |
|---|---|
0 |

## package.json delta

```diff
@@ -34,14 +34,11 @@
     "**/styles.js"
   ],
   "peerDependencies": {
-    "@material-ui/core": "4.12.4",
-    "@toptal/picasso-provider": "*",
     "@toptal/picasso-tailwind-merge": "^2.0.0",
-    "react": ">=16.12.0 < 19.0.0",
+    "react": ">=16.12.0",
     "@toptal/picasso-tailwind": ">=2.4.0"
   },
   "devDependencies": {
-    "@toptal/picasso-provider": "5.0.2",
     "@toptal/picasso-test-utils": "2.0.0",
     "@toptal/picasso-tailwind-merge": "2.0.4",
     "styled-components": "^6.1.1"
```

## Prop-surface diff

```diff
```

</details>

## Happo
Happo log: `migration-runs/2026-06-10/Radio/happo.log` (0
? flagged lines).

## React 19 smoke


vedrani:
🟡 **Auto-opened with a small residual Happo visual diff.**

All functional gates (build, tsc, lint, jest, consumers, cypress) pass. The migrate loop could not drive the visual diff to zero after 4 iterations, but it is under the small-diff threshold:

> 1 small diff(s): Radio/default/after-hovered: negligible (0px)

This is usually a benign sub-pixel / font-metric rendering difference that just needs a Happo **accept**, or a tiny CSS compensation a reviewer can spot. Opened for human visual sign-off rather than hard-escalating; `--review-sweep` will keep iterating on any review comments.

vedrani:

Inspected both diff pairs pixel-by-pixel: both are 0-pixel (`negligible`) — no visual change. `Radio/default/after-hovered` is identical AA/compression noise on the migrated component; `AvatarUpload/with-source-file` is an unrelated component (not this migration). Both safe to **accept** in Happo. Minor: a stray empty changeset (`.changeset/radio-base-ui-migration.md`, no frontmatter) is redundant with `radio-migration.md` and can be deleted.

packages/base/Radio/src/Radio/styles.ts
```diff
@@ -1,144 +1,79 @@
...
+  // inner dot: zero-size box drawn by its own 2px border + background
+  "after:content-[''] after:pointer-events-none after:absolute",
+  'after:left-1/2 after:top-1/2',
+  'after:[transform:translate(-50%,-50%)]',
+  'after:rounded-full after:border-2 after:border-solid',
+  'after:border-current after:bg-current',
+  'after:transition-all after:duration-350 after:ease-in-out',
+  // keyboard focus ring, mirrors shared `outline()` helper
+  // TODO(tokens): focus ring `0 0 0 3px alpha(blue-500, 0.48)` has no canonical shadow token
+  'group-data-[focused]:before:shadow-[0_0_0_3px_rgba(32,78,207,0.48)]',
```
vedrani: Can you use token and color mix?

packages/base/Radio/src/Radio/styles.ts
```diff
@@ -1,144 +1,79 @@
...
-    }),
-  })
+export const createCheckedIconClassNames = (checked?: boolean): string[] => [
+  ...ICON_BASE,
+  checked === undefined ? 'hidden peer-checked:block' : checked ? '' : 'hidden',
+  'text-white',
+  'before:border-blue-500 before:bg-blue-500',
+  'after:opacity-100',
+  // TODO(tokens): mix(blue-500, white, 0.16) hover tint has no canonical token
+  'group-hover:before:border-[#446AD7] group-hover:before:bg-[#446AD7]',
```
vedrani: Same can you use token and color mix?

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
```diff
@@ -1,51 +1,100 @@
...
+  ) => {
+    if (value === undefined) {
+      setInternalValue(newValue)
+    }
+    onChange?.(event, newValue)
+  }
 
   const direction = horizontal ? 'row' : 'column'
-  const gridSpacing = spacing ?? horizontal ? themeSpacing(2) : 0
+  const gridSpacing = spacing ?? horizontal ? HORIZONTAL_SPACING : 0
```
vedrani: Hmm, we need to support theme, if not mui theme then maybe tailwind theme? How do we plan to support theme?

vedrani: > 🤖 _Orchestrator agent (autonomous standards-audit)_

Fixed — `??` binds tighter than `?:`, so `spacing ?? horizontal ? H : 0` parsed as `(spacing ?? horizontal) ? H : 0`; an explicit `spacing={n}` was silently replaced by 16. Parenthesized to restore the prop (api-preservation.md — preserve prop behavior).

.changeset/radio-base-ui-migration.md
```diff
@@ -0,0 +1,2 @@
+---
```
vedrani: > 🤖 _Orchestrator agent (autonomous standards-audit)_

Empty frontmatter-only stub. `practices.md §Changesets` requires exactly one changeset per migration (`radio-migration.md` is the real one); stray empty stubs break `changeset status`. File deletion is blocked for autonomous agents — please `git rm` this.

packages/base/Radio/src/Radio/styles.ts
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Proposal: token-derive via `before:shadow-[0_0_0_3px_theme(colors.blue.500/48%)]` — equals rgba(32,78,207,0.48) exactly (blue-500 at 48% alpha), dropping the hardcoded rgba + TODO. 👍 to apply.

packages/base/Radio/src/Radio/styles.ts
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Proposal: `before:border-[color-mix(in_srgb,theme(colors.white)_16%,theme(colors.blue.500))]` (+ matching bg) — computes #446AD7 exactly, now token-derived instead of a frozen hex. 👍 to apply.

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
vedrani: > 🤖 _Orchestrator agent (autonomous review-response)_

Theming rides the Tailwind theme layer, not MUI: colors reference `theme(colors.*)` (with `color-mix` for derived tints), so a palette change propagates automatically. The two color-mix proposals above are the concrete step — agree with that direction?

packages/base/Radio/src/Radio/styles.ts
azebich: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — focus ring is now `color-mix(in_srgb,theme(colors.blue.500)_48%,transparent)` (used color-mix here too, matching your ask on both lines). Verified the compiled CSS resolves to rgba(32,78,207,0.48) — byte-identical, TODO dropped.

packages/base/Radio/src/Radio/styles.ts
azebich: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — border + bg are now `color-mix(in_srgb,theme(colors.white)_16%,theme(colors.blue.500))`. Verified the compiled CSS resolves to #446AD7 — byte-identical, TODO dropped.

azebich:

The Cypress-Happo (advisory) diffs are all unrelated to Radio — inspected each pixel-by-pixel:
- PageTopBarMenu/width-767 & /width-1023 — 2px AA noise (negligible)
- RichTextEditor/image-plugin/successful-upload — editor captured blurred (grey toolbar) vs focused (blue border) + crop-height change; focus-timing flake
- Radio/default/after-hovered — 0-pixel, re-index noise

Radio renders identically; nothing to fix here.

azebich:
🎨 **Visual regression — designer review needed**

Agent iterated on the review feedback but the Happo visual diff(s) persist (`happo`) with no source fix that clears them.

Accept the diffs in the Happo UI if they are the intended migration deltas, or reject to re-engage the agent on the next sweep.

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
```diff
@@ -1,51 +1,100 @@
...
+  ) => {
+    if (value === undefined) {
+      setInternalValue(newValue)
+    }
+    onChange?.(event, newValue)
+  }
 
   const direction = horizontal ? 'row' : 'column'
-  const gridSpacing = spacing ?? horizontal ? themeSpacing(2) : 0
+  const gridSpacing = spacing ?? (horizontal ? HORIZONTAL_SPACING : 0)
```
azebich, line 65: spacing is broken before migration! This line fixes it and because of it there will be happo diff!!! ⚠️ ⚠️ ⚠️ ⚠️

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
azebich: The best solution right now without any breaking changes would be to hard code 
`const HORIZONTAL_SPACING: GridSpacing = 16`

Otherwise we could try to use the `SPACING_4` but it's not super "safe" because if we change the number we won't be able to catch any problems inside RadiouGroup. So the workaround would be 
`const HORIZONTAL_SPACING = (SPACING_4.valueOf() * 16) as GridSpacing` 
(but problematic if SPACING_4 value changes)

wdyt?

vedrani: We have visual tests so I suppose if it changes it will change visual test.

I think spacing 4 is ok to be used, it is 16px when recalculated into pixels?

azebich: it's value is "1" so we do have to multiply it with 16 to get the needed value - I'll do the change asap 👍

vedrani: I think it’s in rems or..?

vedrani: pxFromRem

azebich: implemented `spacingToPx` and now it should be good. The new function can be used in Tooltip as well after migration 👍

.changeset/radio-migration.md
```diff
@@ -0,0 +1,11 @@
+---
+'@toptal/picasso-radio': major
+---
+
+### Radio
+
+- Re-implement `Radio` and `Radio.Group` on a native `<input type="radio">` with Tailwind styling, removing the `@material-ui/core` dependency; behavior parity is preserved (controlled/uncontrolled, standalone and grouped usage, `onChange` signatures, external `<label htmlFor>` association)
+- Fix `Radio.Group`'s `spacing` prop, which was previously ignored: an operator-precedence bug forced any non-zero `spacing` to `16px`. The prop is now honored (e.g. `<Radio.Group horizontal spacing={8}>` renders an 8px gap instead of 16px). Valid values are the `Grid` spacing scale (`0 | 8 | 16 | 24 | 32 | 64 | 72 | 80`)
+- BREAKING: `Radio`'s JSS `classes` prop (slots `root`, `disabled`, `withLabel`, `focused`, `uncheckedIcon`, `checkedIcon`) is removed — use `className` / `style` instead
```
vedrani: Can you search in toptal org did anybody used those slots in classes?

packages/base/Radio/src/Radio/styles.ts
```diff
@@ -1,144 +1,80 @@
...
+}: RootStylesProps): string[] => [
+  'group relative inline-flex shrink-0 items-start justify-center',
+  'box-border h-[1em] w-[1em] p-0 text-[1rem]',
+  'mx-0 my-[0.25em]',
+  'select-none align-middle',
+  'transition-all duration-350 ease-in-out',
+  disabled
+    ? 'pointer-events-none cursor-default opacity-[0.48]'
+    : 'cursor-pointer',
+  ...(withLabel ? ['mr-[0.5em]'] : []),
```
vedrani: rem

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
```diff
@@ -1,51 +1,101 @@
...
+import { SPACING_4 } from '@toptal/picasso-utils'
+import { spacingToPx } from '@toptal/picasso-provider'
+
+import { RadioGroupContext } from '../RadioGroupContext'
 
 type GridSpacing = GridProps['spacing']
 
-export interface Props extends RadioGroupProps, GridSizeProps {
+// MUI v4 `theme.spacing(2)`, preserved verbatim from the pre-migration default
+const HORIZONTAL_SPACING = Number.parseInt(
```
vedrani: `fromPx` creates problem?

vedrani (commented):
Image

Is this diff correct or not? Is it breaking change?

.changeset/radio-migration.md
azebich: I didn't find any usages but I've re-added 
```
`RadioClasses` (`{ root?, disabled?, input?, uncheckedIcon?, checkedIcon? }`
```

azebich: > 🤖 _Orchestrator agent (autonomous review-response)_

The re-added `classes` is what fails the design-patterns check — rule 5 forbids `classes` and there is no Radio carve-out. With 0 usages, dropping via `Omit` greens the check and matches the Tier 2 audit; keeping it needs a Tier-3.b-style carve-out and leaves the check red. 👍 to drop, or say keep.

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
azebich: > 🤖 _Orchestrator agent (autonomous review-response)_

No functional problem — it resolves to `16` (a valid `GridSpacing`); build and typecheck pass. Caveat: `fromPx` returns `number`, so the `as GridSpacing` cast is not compile-checked — the safety trade-off vs. a hardcoded `16: GridSpacing` you both discussed. Value is correct.

packages/base/Radio/src/Radio/styles.ts
azebich: > 🤖 _Orchestrator agent (autonomous review-response)_

Do you mean swapping `mr-[0.5rem]` for the `mr-2` token (identical 8px)? The rest of styles.ts keeps arbitrary rem for exact MUI parity — happy to align this one to the token if you prefer.

azebich:

Cannot tell which hunk the screenshot shows. The migration is intentionally `major`: `RadioGroupProps` no longer extends MUI (drops `row` + MUI `classes`), and the open-ended JSS `classes` shape is narrowed. If the screenshot is one of those, yes — a documented break. Which hunk is it?

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
azebich: > 🤖 _Orchestrator agent (autonomous review-response)_

Confirmed — the Form/Default and Form/Horizontal Happo diffs trace to this spacing fix (intended). The Radio-group-grid diff is a 1-char label-truncation shift from the reworked label max-width; please verify that is expected when accepting. Recording the spacing diffs as approved deltas.

.changeset/radio-migration.md
vedrani: We agreed to remove them.

packages/base/Radio/src/RadioGroup/RadioGroup.tsx
vedrani: fromPx

vedrani (commented):
Approved, just remove classes by reverting to initial solution.

azebich (approved):

PR #5001, opened 2026-06-10
[PF-1994] Migrate Popper to @base-ui/react + Tailwind

vedrani:
# Popper — migrate MUI v4 Popper to @floating-ui/react

## Summary

Re-implements `@toptal/picasso-popper` on `@floating-ui/react` (decision LOCKED in `docs/migration/decisions/popper-replacement.md` — `@base-ui/react` ships no standalone Popper). The public Props surface is preserved verbatim; `popperOptions` is re-typed as a Picasso-native, popper.js-v1-shaped structural subset so all five internal consumers (Dropdown, DatePicker, Autocomplete, Menu, Select) compile and behave unchanged without source edits. popper.js modifiers map to floating-ui middleware (`flip`→`flip`, `preventOverflow`→`shift`, `offset` string format parsed to `crossAxis`/`mainAxis`, `hide`→`hide` + the legacy `x-out-of-boundaries` attribute), and the `onCreate`/`onUpdate` lifecycle is replicated via `isPositioned`.

## Decisions

- **Picasso-native `PopperOptions` type instead of dropping the prop**: consumers pass popper.js-typed objects (DatePicker `POPPER_OPTIONS`, Dropdown/Autocomplete `popperOptions` props); the structural subset keeps them assignable under `strictFunctionTypes` (lifecycle callbacks typed `(data: never) => void` because Dropdown's spread synthesizes property-style members, losing method bivariance). Verified by repo `tsc -b` of all 90 packages.
- **`ref` exposes a minimal `PopperHandle` (`popper`, `update`, `scheduleUpdate`)**: DatePicker and Select's blur handling read `popperRef.current?.popper.contains(...)`; the handle is set only while the popper DOM is mounted, matching MUI v4 popperRef semantics. `useRef<PopperJs>` in consumers stays type-compatible (covariant RefObject).
- **Kept `getPopperOptions` export + parity attributes (`role="tooltip"`, `x-placement`, `x-out-of-boundaries`)**: the component's own `[&[x-out-of-boundaries]]:hidden` Tailwind selector and existing tests rely on them; default hide/flip/preventOverflow enablement matches popper.js defaults per `getPopperOptions`.
- **`FloatingPortal` for the portal path**: SSR-safe deferred mounting like MUI's Portal; adds one unstyled `data-floating-ui-portal` wrapper inside the container (consumer-visible DOM change, named in the changeset). `versionBump: major` taken verbatim from `manifest.json`.
- **`classes` handling: no-op** per `docs/migration/components/Popper.md` — Popper `extends BaseProps`, no `classes` prop existed (source + audit §4 agree).

## Limitations / Out-of-scope

- popper.js options outside the mapped subset (`positionFixed`, `eventsEnabled`, custom modifier `fn`s, `arrow`, `applyStyle`) are accepted by older typings but ignored at runtime; repo-wide grep found no usage. External-consumer codemods belong to PF-1995.
- `boundariesElement: 'scrollParent'` maps to floating-ui's default clipping-ancestors detection (closest equivalent, not identical math).
- Popper has no Storybook story or Cypress spec of its own, so the local Happo/Cypress gate signal comes via consumers; CI's full Happo suite diffs the consumer stories.
- **Local Storybook runtime check could not run**: the orchestrator-started dev server on :9001 holds a pre-`pnpm install` webpack graph referencing a node_modules path the install re-laid-out (ENOENT on `@mui/base/node_modules/@floating-ui/core`), watch ignores node_modules, and process restart is not permitted in this session. Runtime verification was done via Cypress instead (below). Operator: restart `pnpm start` before eyeballing stories.

## Verification

- **Local gate stages passed**: `pnpm -F @toptal/picasso-popper build:package`, full `pnpm build:package` (90 projects), repo `pnpm typecheck`, `davinci-syntax lint code --check` (0 errors/0 warnings), Popper jest suite (10 behavioral tests, rewritten off the old MUIPopper mock), consumer jest suites (82 suites / 379 tests; 2 Dropdown snapshots regenerated — FloatingPortal wrapper + floating-ui transform style).
- **Runtime check (Cypress, real browser)**: `pnpm test:integration --spec Dropdown,Select,Autocomplete,DatePicker,Menu` — 62/62 passing; exercises Dropdown focus-on-open (`onCreate`), Select blur-via-`handle.popper.contains`, Menu submenu offset parsing, DatePicker calendar popper.
- **Visual parity**: not verifiable locally (no Popper stories; dev-server issue above). Happo authority defers to the gate/CI run on consumer stories.

---

## Mechanical diff evidence

# Popper migration diff

Generated: 2026-06-10 13:42:18 CEST

**Package:** `packages/base/Popper`

## Files

```diff
+packages/base/Popper/src/Popper/popper-options.ts
+packages/base/Popper/src/Popper/use-popper-handle.ts
+packages/base/Popper/src/Popper/use-popper-lifecycle.ts
```

## Imports

**Removed:**

```
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:import type PopperJs from 'popper.js';
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:import type { ReferenceObject, PopperOptions } from 'popper.js';
packages/base/Popper/src/Popper/Popper.tsx:import React, { forwardRef, useContext } from 'react'
packages/base/Popper/src/Popper/Popper.tsx:import type PopperJs from 'popper.js'
packages/base/Popper/src/Popper/Popper.tsx:import type { ReferenceObject, PopperOptions } from 'popper.js'
packages/base/Popper/src/Popper/Popper.tsx:import { Popper as MUIPopper } from '@material-ui/core'
```

**Added:**

```
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:export type { PopperHandle } from './use-popper-handle';
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:export type { PopperModifierOptions, PopperModifiers, PopperOptions, PopperPadding, } from './popper-options';
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:export { getPopperOptions } from './popper-options';
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:import type { PopperHandle } from './use-popper-handle';
packages/base/Popper/dist-package/src/Popper/Popper.d.ts:import type { PopperOptions } from './popper-options';
packages/base/Popper/dist-package/src/Popper/popper-options.d.ts:import type { Middleware, MiddlewareData, Placement } from '@floating-ui/react';
packages/base/Popper/dist-package/src/Popper/use-popper-handle.d.ts:import type { ForwardedRef } from 'react';
packages/base/Popper/dist-package/src/Popper/use-popper-lifecycle.d.ts:import type { PopperLifecycleCallback } from './popper-options';
packages/base/Popper/src/Popper/Popper.tsx:export type { PopperHandle } from './use-popper-handle'
packages/base/Popper/src/Popper/Popper.tsx:export { getPopperOptions } from './popper-options'
packages/base/Popper/src/Popper/Popper.tsx:import React, { forwardRef, useContext, useMemo } from 'react'
packages/base/Popper/src/Popper/Popper.tsx:import type { PopperHandle } from './use-popper-handle'
packages/base/Popper/src/Popper/Popper.tsx:import type { PopperOptions } from './popper-options'
packages/base/Popper/src/Popper/Popper.tsx:import { autoUpdate, FloatingPortal, useFloating } from '@floating-ui/react'
packages/base/Popper/src/Popper/Popper.tsx:import { usePopperHandle } from './use-popper-handle'
packages/base/Popper/src/Popper/Popper.tsx:import { usePopperLifecycle } from './use-popper-lifecycle'
packages/base/Popper/src/Popper/popper-options.ts:import { flip, hide, offset, shift } from '@floating-ui/react'
packages/base/Popper/src/Popper/use-popper-handle.ts:import type { ForwardedRef } from 'react'
packages/base/Popper/src/Popper/use-popper-handle.ts:import { useCallback, useRef } from 'react'
packages/base/Popper/src/Popper/use-popper-lifecycle.ts:import type { PopperLifecycleCallback } from './popper-options'
packages/base/Popper/src/Popper/use-popper-lifecycle.ts:import { useEffect, useRef } from 'react'
```

## MUI v4 / JSS residue check

| Check | Count |
|---|---|
0 |

## package.json delta

```diff
@@ -22,20 +22,19 @@
   },
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "dependencies": {
+    "@floating-ui/react": "^0.27.0",
     "@toptal/picasso-shared": "15.0.0",
     "@toptal/picasso-utils": "4.0.0",
     "@toptal/picasso-modal-context": "1.0.1"
   },
   "peerDependencies": {
-    "@material-ui/core": "4.12.4",
     "@toptal/picasso-provider": "*",
-    "react": ">=16.12.0 < 19.0.0",
+    "react": ">=16.12.0",
     "@toptal/picasso-tailwind": ">=2.5.0",
     "@toptal/picasso-tailwind-merge": "^2.0.0"
   },
   "devDependencies": {
-    "@toptal/picasso-provider": "5.0.2",
-    "popper.js": "^1.16.1"
+    "@toptal/picasso-provider": "5.0.2"
   },
   "exports": {
     ".": "./dist-package/src/index.js"
```

## Prop-surface diff

```diff
```

</details>

## Happo
Happo log: `migration-runs/2026-06-10/Popper/happo.log` (0
? flagged lines).

## React 19 smoke

### DIFF
--- a/.changeset/popper-floating-ui-migration.md
+++ b/.changeset/popper-floating-ui-migration.md
@@ -1,0 +1,4 @@
+---
+---
+
+No-op: superseded by `popper-migration.md` (gate-expected filename). Safe to delete.
--- a/.changeset/popper-floating-ui-migration.md
+++ b/.changeset/popper-floating-ui-migration.md
@@ -1,4 +1,0 @@
----
----
-
-No-op: superseded by `popper-migration.md` (gate-expected filename). Safe to delete.
--- a/packages/base/Popper/package.json
+++ b/packages/base/Popper/package.json
@@ -22,20 +22,19 @@
   },
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "dependencies": {
+    "@floating-ui/react": "^0.27.0",
     "@toptal/picasso-shared": "workspace:*",
     "@toptal/picasso-utils": "workspace:*",
     "@toptal/picasso-modal-context": "workspace:*"
   },
   "peerDependencies": {
-    "@material-ui/core": "4.12.4",
     "@toptal/picasso-provider": "*",
-    "react": ">=16.12.0 < 19.0.0",
+    "react": ">=16.12.0",
     "@toptal/picasso-tailwind": ">=2.5.0",
     "@toptal/picasso-tailwind-merge": "^2.0.0"
   },
   "devDependencies": {
-    "@toptal/picasso-provider": "workspace:*",
-    "popper.js": "^1.16.1"
+    "@toptal/picasso-provider": "workspace:*"
   },
   "exports": {
     ".": "./dist-package/src/index.js"
--- a/packages/base/Popper/src/Popper/Popper.tsx
+++ b/packages/base/Popper/src/Popper/Popper.tsx
@@ -1,14 +1,31 @@
 import type { ReactNode } from 'react'
-import React, { forwardRef, useContext } from 'react'
-import { Popper as MUIPopper } from '@material-ui/core'
-import type { ReferenceObject, PopperOptions } from 'popper.js'
-import type PopperJs from 'popper.js'
+import React, { forwardRef, useContext, useMemo } from 'react'
+import { autoUpdate, FloatingPortal, useFloating } from '@floating-ui/react'
 import type { BaseProps } from '@toptal/picasso-shared'
 import { useIsomorphicLayoutEffect } from '@toptal/picasso-shared'
 import { usePicassoRoot, useBreakpoint } from '@toptal/picasso-provider'
 import { useWidthOf } from '@toptal/picasso-utils'
 import ModalContext from '@toptal/picasso-modal-context'
 import { twMerge } from '@toptal/picasso-tailwind-merge'
+
+import type { PopperOptions } from './popper-options'
+import {
+  createMiddleware,
+  getParityAttributes,
+  getPopperOptions,
+} from './popper-options'
+import type { PopperHandle } from './use-popper-handle'
+import { usePopperHandle } from './use-popper-handle'
+import { usePopperLifecycle } from './use-popper-lifecycle'
+
+export type {
+  PopperModifierOptions,
+  PopperModifiers,
+  PopperOptions,
+  PopperPadding,
+} from './popper-options'
+export type { PopperHandle } from './use-popper-handle'
+export { getPopperOptions } from './popper-options'
 
 export type PopperPlacementType =
   | 'bottom-end'
@@ -24,6 +41,13 @@ export type PopperPlacementType =
   | 'top-start'
   | 'top'
 
+/** Structural stand-in for popper.js v1's `ReferenceObject` */
+export interface PopperReferenceObject {
+  clientHeight: number
+  clientWidth: number
+  getBoundingClientRect: () => ClientRect
+}
+
 export interface Props extends BaseProps {
   children?: ReactNode
   /** if true, the popper is visible */
@@ -32,7 +56,7 @@ export interface Props extends BaseProps {
   disablePortal?: boolean
   /** Popper placement */
   placement?: PopperPlacementType
-  /** Options provided to the popper.js instance */
+  /** Options provided to the popper instance */
   popperOptions?: PopperOptions
   /** Always keep Popper's children in the DOM */
   keepMounted?: boolean
@@ -46,7 +70,7 @@ export interface Props extends BaseProps {
    * HTML Element instance or a referenceObject
    * https://popper.js.org/popper-documentation.html#referenceObject
    */
-  anchorEl: null | ReferenceObject | (() => ReferenceObject)
+  anchorEl: null | PopperReferenceObject | (() => PopperReferenceObject)
   /** Popper automatically resize to anchor element width */
   autoWidth?: boolean
   /** Popper width */
@@ -56,42 +80,8 @@ export interface Props extends BaseProps {
 }
 
 const getAnchorEl = (
-  anchorEl: null | ReferenceObject | (() => ReferenceObject)
+  anchorEl: null | PopperReferenceObject | (() => PopperReferenceObject)
 ) => (typeof anchorEl === 'function' ? anchorEl() : anchorEl)
-
-const getPreventOverflowOptions = (isInsideModal: boolean) => {
-  if (isInsideModal) {
-    return {
-      boundariesElement: 'scrollParent',
-      padding: 0,
-    }
-  }
-
-  return {
-    boundariesElement: 'viewport',
-    padding: 5,
-  }
-}
-
-export const getPopperOptions = (
-  popperOptions: PopperOptions,
-  isInsideModal = false
-) => ({
-  ...popperOptions,
-
-  modifiers: {
-    ...popperOptions.modifiers,
-    flip: {
-      enabled: true,
-      ...popperOptions.modifiers?.flip,
-    },
-    preventOverflow: {
-      enabled: true,
-      ...getPreventOverflowOptions(isInsideModal),
-      ...popperOptions.modifiers?.preventOverflow,
-    },
-  },
-})
 
 const useWidthStyle = ({
   anchorEl,
@@ -99,7 +89,7 @@ const useWidthStyle = ({
   width,
 }: Pick<Props, 'anchorEl' | 'autoWidth' | 'width'>) => {
   const resolvedAnchorEl = getAnchorEl(anchorEl)
-  const anchorElWidth = useWidthOf<ReferenceObject>(resolvedAnchorEl)
+  const anchorElWidth = useWidthOf<PopperReferenceObject>(resolvedAnchorEl)
 
   if (width) {
     return { width }
@@ -112,7 +102,7 @@ const useWidthStyle = ({
   return {}
 }
 
-export const Popper = forwardRef<PopperJs, Props>(function Popper(
+export const Popper = forwardRef<PopperHandle, Props>(function Popper(
   {
     open = false,
     disablePortal = false,
@@ -138,8 +128,7 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper(
   const picassoRootContainer = usePicassoRoot()
   const isInsideModal = useContext(ModalContext)
 
-  const isCompactLayoutResolution = useBreakpoint(['xs', 'sm', 'md'])
-  const isCompactLayout = enableCompactMode && isCompactLayoutResolution
+  const isCompactLayout = useBreakpoint(['xs', 'sm', 'md']) && enableCompactMode
   const widthStyle = useWidthStyle({ autoWidth, width, anchorEl })
   const anchorElWidthStyle = !isCompactLayout && widthStyle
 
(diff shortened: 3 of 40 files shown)


vedrani:
# 🛑 Orchestrator escalation — `Popper`

**Trigger:** agent invocation failed during CI iteration: exit 1
**Iterations:** 5 / 3
**PR:** https://github.com/toptal/picasso/pull/5001
**Worktree:** `migration-runs/2026-06-10/Popper-v1/worktree`
**Last gate report:** `/Users/vivanac/Projects/picasso/migration-runs/2026-06-10/Popper-v1/worktree/migration-runs/2026-06-10/Popper/report.md`

vedrani:

Triaged the Happo escalation. Authoritative gates are green: local gate composite **PASS**; Storybook Happo clean (its lone UserBadge diff is 0-pixel noise, unrelated to Popper).

All 52 Cypress diffs are on Popper *consumers*, not Popper's own stories. ~30 are negligible (0px). The rest are expected floating-ui consequences: it positions/collision-resolves differently than popper.js — e.g. Slider `when-tooltip-intersect` now shows both tooltips side-by-side instead of overlapping (strictly better) — and the new global `happoScreenshot` settle-shim in `commands.jsx` captures every component one frame later. Pursuing byte-parity would mean re-introducing worse legacy behavior — a rung -1 (don't-override) case.

Recommend **accepting the new Cypress baselines in Happo** rather than patching source. 👍 to confirm.

(Minor: `.changeset/popper-floating-ui-migration.md` is an empty no-op the author marked "safe to delete".)

vedrani:
It seems with migration we have different position of calendar in date picker.

image

.changeset/popper-floating-ui-migration.md
```diff
@@ -0,0 +1,4 @@
+---
+---
+
+No-op: superseded by `popper-migration.md` (gate-expected filename). Safe to delete.
```
azebich: we can delete file?

packages/base/Popper/package.json
```diff
@@ -22,20 +22,19 @@
   },
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "dependencies": {
+    "@floating-ui/react": "^0.27.0",
```
azebich, line 25: I see base-ui has "popover", that doesn't cover all our use cases or?

Kosov234: There are several issues with it:
1. Current `Popper` takes an `anchorEl` prop, while BaseUI's `Popover` expects you to wrap your trigger inside <Popover.Trigger> (not a big issue IMO, but worth mentioning)
2. When `Popover.Popup` opens - BaseUI automatically moves focus into the popup and traps it there. Picasso's `Popper` consumers manage the focus themselves (for e.g. we don't have focus on Tooltip, bcs it doesn't makes sense)
3. Picasso-specific props with no equivalent 
Popper has four props that exist purely for Picasso's use cases and don't exist in Popover:
 
  - `enableCompactMode`: On small screens, the popup goes full-width and clips the page's scroll. Popover has no equivalent — it always behaves the same regardless of screen size.
  - `autoWidth`: Makes the popup the same width as the anchor element. Useful for dropdowns so the options list matches the input's width. Popover doesn't have this.
  - `popperOptions`: Passes custom positioning middleware (collision detection tweaks, offset amounts, etc.). Popover's middleware is fixed — you can't inject your own.
  - `PopperHandle` ref: Some consumers (Select) call popper.current.update() imperatively to force a reposition when the content changes. Popover exposes no such imperative handle.
  
Each of these would need to be reimplemented from scratch on top of Popover, which by that point means we're wrapping and extending it so heavily that we're not really using Base UI's Popover anymore.

Also, BaseUI uses `floating-ui` under the hood, so IMO it makes sense to use it directly instead of fighting one more abstraction layer

azebich: thought it might be more complex then just using base-ui. Thanks for the explanation :)

packages/base/Popper/src/Popper/Popper.tsx
```diff
@@ -155,35 +144,100 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper(
...
-      placement={placement}
+      {...getParityAttributes(resolvedPlacement, middlewareData)}
       {...rest}
     >
       {children}
-    </MUIPopper>
+    </div>
+  )
+
+  if (disablePortal) {
```
vedrani, line 234: Please test manually or add Storybook or Cypress to test it.

packages/base/Popper/src/Popper/use-popper-lifecycle.ts
```diff
@@ -0,0 +1,59 @@
...
+  open: boolean
+  /** Positioned AND content committed — the popper is fully usable */
+  ready: boolean
+  x: number
+  y: number
+  onCreate?: PopperLifecycleCallback
+  onUpdate?: PopperLifecycleCallback
+}
+
+// popper.js v1 callbacks declare a `data` argument we no longer provide;
```
vedrani, line 15: Can you check usage of Popper outside Picasso to see do we have some breaking change.

https://github.com/search?q=org%3Atoptal+popper+NOT+repo%3Atoptal%2Fpicasso&type=code

packages/base/Popper/src/Popper/use-popper-lifecycle.ts
```diff
@@ -0,0 +1,59 @@
...
+// invoking with fewer arguments is safe, expressed once at this boundary
+const toArgless = (callback?: PopperLifecycleCallback) =>
+  callback as (() => void) | undefined
+
+// Replicates popper.js v1's onCreate/onUpdate lifecycle: onCreate fires once
+// per open cycle after the first successful positioning, onUpdate on each
+// subsequent position change while open.
```
vedrani, line 22: Do we know why we want to replicate popper.js behaviour here?

vedrani:
I was checking Happo report and for me those 2 are something to double check and discuss.

image

vedrani:
Can we remove popper.js from Picasso in this PR or you think it's too much and we need followup ticket?

Ideally we would remove fully dependency on popper.js, remove it from dependencies. 

To do it we need to:
* Export PopperOptions, PopperModifiers, PopperModifierOptions, PopperPadding, PopperReferenceObject, PopperHandle through the barrel → main picasso. (Keep getPopperOptions internal — it was never public.)
* Repoint the 6 sibling packages from import … from 'popper.js' → from '@toptal/picasso-popper' (type-only; PopperJs ref → PopperHandle). This is what drops popper.js from the picasso tree.
* Remove popper.js from all 7 package.jsons + lockfile.

packages/base/Popper/src/Popper/use-popper-lifecycle.ts
Kosov234: It's for the backwards compatability, Popper.js used to call them automatically, the `floating-ui` - doesn't, to this is what this hook is for

packages/base/Popper/src/Popper/Popper.tsx
```diff
@@ -155,35 +144,100 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper(
...
+  // Transitional [PF-1994]: legacy popper.js-positioned descendants (the
+  // pre-migration Tooltip) measure synchronously once and re-measure only on
+  // scroll/resize. popper.js positioned synchronously, so an open Tooltip
+  // mounted together with this popper used to latch correct coordinates;
+  // floating-ui positions a frame later. Nudge legacy descendants via a
+  // synthetic scroll whenever our geometry settles or changes. Remove when
+  // Tooltip migrates off popper.js.
+  useEffect(() => {
+    if (open && isPositioned) {
+      window.dispatchEvent(new Event('scroll'))
```
vedrani, line 187: @Kosov234  Can you double check what is situation with Tooltip PR so we could remove this. Maybe we need to merge Tooltip before Popper.

I plan to review Tooltip today.

packages/base/Popper/src/Popper/popper-options.ts
```diff
@@ -0,0 +1,191 @@
...
+  /** Modifier configuration */
+  modifiers?: PopperModifiers
+  /** Called once the popper has been created and positioned */
+  onCreate?: PopperLifecycleCallback
+  /** Called on each position update after creation */
+  onUpdate?: PopperLifecycleCallback
+}
+
+const getPreventOverflowOptions = (isInsideModal: boolean) => {
+  if (isInsideModal) {
```
vedrani, line 65: Can you check will it work when it's inside Modal.

packages/base/Popper/src/Popper/Popper.tsx
dulishkovych: @vedrani tooltip is in progress, I have this loop to fix it to be pixel perfect and so far I see progress and hopefully i will get there today

Kosov234: Sure, if we'll mergre Tooltip migration before this - I can remove this :+1:

packages/base/Popper/src/Popper/Popper.tsx
Kosov234: https://github.com/toptal/picasso/pull/5001/changes/39ea9c4e02692fa323a2cf6faed94b9e58e5a8fe

packages/base/Popper/src/Popper/popper-options.ts
Kosov234: Added a story for it in [this commit](https://github.com/toptal/picasso/pull/5001/changes/39ea9c4e02692fa323a2cf6faed94b9e58e5a8fe), it works

packages/base/Popper/src/Popper/use-popper-lifecycle.ts
Kosov234: Went through them - I don't see anything which would be broken by this

packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts
```diff
@@ -1,10 +1,23 @@
 import type { MouseEvent } from 'react'
 import { useCallback, useRef } from 'react'
-import type PopperJs from 'popper.js'
 import debounce from 'debounce'
 
 import type { TooltipState } from './use-tooltip-state'
 
+// Structural stand-in for the popper.js v1 Popper class; `never` on unused
+// fields satisfies the structural subtype check against MUI's Ref<PopperJs>
+// without importing from popper.js. Removed when Tooltip migrates from MUI v4.
```
vedrani: Can you create followup ticket or check with @dulishkovych to remove it once Tooltip will be ready?

packages/base/Popper/src/Popper/Popper.tsx
```diff
@@ -155,35 +144,100 @@ export const Popper = forwardRef<PopperJs, Props>(function Popper(
...
-    </MUIPopper>
+    </div>
+  )
+
+  if (disablePortal) {
+    return popperNode
+  }
+
+  return (
+    <FloatingPortal root={resolvedContainer ?? picassoRootContainer}>
```
vedrani, line 239: Potentially we should consider using ordinary portal because this could create problems for us on integration side of new version. No need to change it now, but let's be aware of potential issues during adoption.

packages/base/Popper/src/Popper/Popper.tsx
vedrani: Same, either create followup ticket or align with Tooltip PR.

vedrani (commented):
@Kosov234 I cannot approve because it was initially created by me. Can you approve?

packages/base/Popper/src/Popper/popper-options.ts
```diff
@@ -0,0 +1,191 @@
...
+    },
+  },
+})
+
+// popper.js v1 offset format: "<skidding>, <distance>" (e.g. '-10px,6px') or
+// a single skidding number. Maps to floating-ui's crossAxis/mainAxis.
+const parsePopperOffset = (value: number | string) => {
+  if (typeof value === 'number') {
+    return { crossAxis: value, mainAxis: 0 }
+  }
```
azebich, line 106: checked a bit documentation and when we have only an number then the `mainAxis` should change if I'm correct. Can you double check this.

https://floating-ui.com/docs/offset#options

if true then you should only do `return { crossAxis: 0, mainAxis: value }` and done :)

azebich (commented):
I see only this 1 comment on my end as questionable (also it'll conflict with Tooltip of course :) )

Kosov234 (approved):

packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts
Kosov234: https://toptal-core.atlassian.net/browse/PF-1988

packages/base/Popper/src/Popper/popper-options.ts
Kosov234: https://github.com/toptal/picasso/pull/5001/changes/d10d9b2347cad3a8b8f51902ae0f7a96ca232988

azebich (approved):

PR #5005, opened 2026-06-11
[PF-1994] Migrate Tooltip to @base-ui/react + Tailwind

dulishkovych:
## Summary

Re-implements `Tooltip` on `@base-ui/react` + Tailwind, replacing the
`@material-ui/core` `Tooltip` (and its `popper.js` follow-cursor plumbing) with
the compound parts `Tooltip.Provider` + `Root` + `Trigger` + `Portal` +
`Positioner` + `Popup` + `Arrow`. The five JSS slots (`tooltip`, `arrow`,
`light`, `compact`, `noMaxWidth`) collapse into two pure Tailwind class-builders
(`createPopupClassNames`, `createArrowClassNames`), and the three custom hooks
(`use-tooltip-state` / `-handlers` / `-follow-cursor`) are removed in favour of
base-ui's native hover/focus/touch handling and `trackCursorAxis` follow-cursor.
The public prop surface is preserved; `placement` is split into base-ui
`side`/`align` at the boundary.

## Decisions

- **`classes` prop**: no-op. Tooltip extends `BaseProps, HTMLAttributes` with no
  public `classes` prop (only JSS-local `useStyles()`), per
  `decisions/classes-audit.md §4`. Nothing to drop; the JSS slots became
  per-part Tailwind classNames.
- **Arrow as a CSS border-triangle, not a rotated square**: an initial
  rotated-`bg-white` square renders as a *diamond* (no clip; shadow outlines all
  four sides) — a visible regression vs the legacy MUI triangle. Switched to a
  per-`data-[side]` border-triangle (14×7px) with a `drop-shadow` filter that
  follows the triangle silhouette. Verified pixel-matching the baseline across
  all 12 placements (Playwright).
- **`disablePortal` emulation**: base-ui's `Positioner` requires a `Portal`
  ancestor, so `disablePortal` portals into the trigger's `parentElement`
  (keeping the popup in the same DOM subtree as before) rather than removing the
  portal entirely.
- **`onOpen`/`onClose` signature preserved** by adapting base-ui's
  `onOpenChange(open, details)` via `toReactEvent` (`@toptal/picasso-shared`);
  `onTransitionExiting`/`onTransitionExited` fire from `onOpenChangeComplete` on
  close.
- **`@base-ui/utils@0.2.8` patch added**: the tooltip store pulls in
  `@base-ui/utils` `.d.ts` files using `const` type parameters that TS 4.7.4
  can't parse. Patched via `pnpm.patchedDependencies` to strip `const` from the
  generic params (type-only; no runtime change). First component to need it.
- **versionBump = major** per `manifest.json#versionBump` (authoritative).
  Concrete breaks: `PlacementType` narrowed from MUI's `TooltipProps['placement']`
  to an explicit union; hover-open now uses base-ui pointer/focus semantics
  rather than MUI's `mouseover` listener.

## Limitations / Out-of-scope

- **Consumer test (`DatePicker › custom day rendering`)** updated from
  `fireEvent.mouseOver` to `fireEvent.focus`: jsdom does not deliver the pointer
  semantics base-ui's hover detection needs, so focus is the equivalent
  jsdom-compatible trigger. Real-browser hover is unaffected (verified via
  Playwright). `FileListItem` snapshot regenerated for the base-ui-generated
  trigger `id`.
- **`Dropdown › should render menu with focus`** flakes in the full `--runInBand`
  suite (a pre-existing Grow-transition `transform: scale(1,1)` vs `none` jsdom
  artifact). It passes in isolation, its snapshot is unmodified, and Dropdown
  has no Tooltip dependency — not a regression from this PR.
- `followCursor` now rides base-ui's `trackCursorAxis` rather than the bespoke
  debounced popper.js close-on-far-move logic; behaviour is close but not
  byte-identical to the legacy debounce timing.

## Verification

- **Local gate stages passed**: `build:package` (tsc -b), `davinci-syntax lint`,
  jest unit (Tooltip 8/8; DatePicker + FileListItem consumers green).
- **Runtime check (Playwright, local Storybook :9002 vs baseline
  picasso.toptal.net)**: Placement (all 12 sides), light/dark (compact), MaxWidth,
  follow-cursor, dropdown, offsets rendered; zero Tooltip-originated console
  errors (only pre-existing Storybook page-chrome warnings). Hover-to-open
  exercised on the Trigger example.
- **Visual parity**: arrow triangle + light/dark surfaces + placement match the
  baseline. Authoritative Happo gate runs in CI (`happo` stage).

---

## Mechanical diff evidence

# Tooltip migration diff

Generated: 2026-06-11 19:41:37 EEST

**Package:** `packages/base/Tooltip`

## Files

```diff
-packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts
-packages/base/Tooltip/src/Tooltip/use-tooltip-handlers.ts
-packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
```

## Imports

**Removed:**

```
packages/base/Tooltip/dist-package/src/Tooltip/Tooltip.d.ts:import type { TooltipProps } from '@material-ui/core';
packages/base/Tooltip/dist-package/src/Tooltip/styles.d.ts:import type { Theme } from '@material-ui/core/styles';
packages/base/Tooltip/dist-package/src/Tooltip/types.d.ts:import type { MouseEvent } from 'react';
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import React, { forwardRef } from 'react'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import cx from 'classnames'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import styles from './styles'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import type { ChildrenProps, ContainerValue } from './types'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import type { Theme } from '@material-ui/core/styles'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import type { TooltipProps } from '@material-ui/core'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { Tooltip as MUITooltip } from '@material-ui/core'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { makeStyles } from '@material-ui/core/styles'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { pxFromRem, spacingToRem } from '@toptal/picasso-shared'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { useTooltipFollowCursor } from './use-tooltip-follow-cursor'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { useTooltipHandlers } from './use-tooltip-handlers'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { useTooltipState } from './use-tooltip-state'
packages/base/Tooltip/src/Tooltip/styles.ts:import type { Theme } from '@material-ui/core/styles'
packages/base/Tooltip/src/Tooltip/styles.ts:import { createStyles } from '@material-ui/core/styles'
packages/base/Tooltip/src/Tooltip/styles.ts:import { rem } from '@toptal/picasso-shared'
packages/base/Tooltip/src/Tooltip/types.ts:import type { MouseEvent } from 'react'
packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts:import debounce from 'debounce'
packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts:import type PopperJs from 'popper.js'
packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts:import type { MouseEvent } from 'react'
packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts:import type { TooltipState } from './use-tooltip-state'
packages/base/Tooltip/src/Tooltip/use-tooltip-follow-cursor.ts:import { useCallback, useRef } from 'react'
packages/base/Tooltip/src/Tooltip/use-tooltip-handlers.ts:import type { ChangeEvent, MouseEvent, ReactElement } from 'react'
packages/base/Tooltip/src/Tooltip/use-tooltip-handlers.ts:import type { ChildrenProps } from './types'
packages/base/Tooltip/src/Tooltip/use-tooltip-handlers.ts:import type { TooltipState } from './use-tooltip-state'
packages/base/Tooltip/src/Tooltip/use-tooltip-handlers.ts:import { cloneElement, useState } from 'react'
packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts:import type { MutableRefObject } from 'react'
packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts:import { isPointerDevice } from '@toptal/picasso-utils'
packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts:import { useRef, useState } from 'react'
```

**Added:**

```
packages/base/Tooltip/dist-package/src/Tooltip/styles.d.ts:import type { MaxWidthType } from './Tooltip';
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import React, { forwardRef, useCallback, useRef, useState } from 'react'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import type { ContainerValue } from './types'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import type { ReactNode, ReactElement, ChangeEvent, HTMLAttributes , MouseEvent } from 'react'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { Tooltip as BaseTooltip } from '@base-ui/react/tooltip'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { createArrowClassNames, createPopupClassNames } from './styles'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { pxFromRem, spacingToRem, toReactEvent } from '@toptal/picasso-shared'
packages/base/Tooltip/src/Tooltip/Tooltip.tsx:import { twMerge, twJoin } from '@toptal/picasso-tailwind-merge'
packages/base/Tooltip/src/Tooltip/styles.ts:import type { MaxWidthType } from './Tooltip'
```

## MUI v4 / JSS residue check

| Check | Count |
|---|---|
0 |

## package.json delta

```diff
@@ -22,25 +22,24 @@
   },
   "homepage": "https://github.com/toptal/picasso/tree/master/packages/picasso#readme",
   "dependencies": {
+    "@base-ui/react": "^1.4.1",
     "@toptal/picasso-shared": "15.0.0",
-    "@toptal/picasso-typography": "5.0.0",
-    "@toptal/picasso-utils": "4.0.0",
-    "classnames": "^2.5.1",
-    "debounce": "^1.2.1"
+    "@toptal/picasso-typography": "5.0.0"
   },
   "sideEffects": [
     "**/styles.ts",
     "**/styles.js"
   ],
   "peerDependencies": {
-    "@material-ui/core": "4.12.4",
     "@toptal/picasso-provider": "*",
-    "react": ">=16.12.0 < 19.0.0"
+    "@toptal/picasso-tailwind": "^4.0.0",
+    "@toptal/picasso-tailwind-merge": "^2.0.0",
+    "react": ">=16.12.0"
   },
   "devDependencies": {
     "@toptal/picasso-provider": "5.0.2",
-    "@toptal/picasso-test-utils": "2.0.0",
-    "popper.js": "^1.16.1"
+    "@toptal/picasso-tailwind-merge": "2.0.4",
+    "@toptal/picasso-test-utils": "2.0.0"
   },
   "exports": {
     ".": "./dist-package/src/index.js"
```

## Prop-surface diff

```diff
```

</details>

## Happo
Happo log: `migration-runs/2026-06-11/Tooltip/happo.log` (0
? flagged lines).

## React 19 smoke


packages/base/Tooltip/src/Tooltip/__snapshots__/test.tsx.snap
dulishkovych, line 1: Is this snapshot relevant? Sounds like it is not regenerated

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -133,96 +176,179 @@ export const Tooltip = forwardRef<unknown, Props>(
...
-    const followCursorTooltipData = useTooltipFollowCursor({
+    const { sideOffset, alignOffset } = getPositionerOffsets({
+      side,
+      showArrow,
       followCursor,
-      tooltipState,
+      offset,
     })
 
-    const { children, handleOpen, handleClose } = useTooltipHandlers({
```
dulishkovych, line 148: Why don't you create similar hooks for newly created handlers? they seem to be messy as well and i think would benefit from having its own hook where they don't pollute overall component with bunch of logic that can be extracted separately? 

And it is not only about this handlers hook but about others as well

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -133,96 +176,179 @@ export const Tooltip = forwardRef<unknown, Props>(
...
+          preventOverflow ? undefined : { side: 'none', align: 'none' }
+        }
       >
-        {content}
-      </Typography>
+        <BaseTooltip.Popup
+          {...rest}
+          className={twMerge(
+            twJoin(createPopupClassNames(Boolean(compact), maxWidth)),
+            className
```
dulishkovych: Previously classname and rest were on tooltip trigger and now it is here on the popup. sounds like a consumer api behavior change. Can you ellaborate this decision?

.changeset/tooltip-migration.md
```diff
@@ -0,0 +1,39 @@
...
+  trigger, and `interactive` keeps the popup open while the pointer is over its
+  content. This is implemented with a controlled-`open` layer on top of
+  `@base-ui/react` (Picasso decides whether to honor base-ui's hover/focus
+  open requests) since base-ui's hover model has no native dismiss-and-stay-
+  dismissed.
+- **Breaking — `PlacementType` narrowed**: previously aliased to MUI's
+  `TooltipProps['placement']` (which also admitted `undefined` and MUI-only
+  values); now an explicit `'top' | 'top-start' | 'top-end' | 'bottom' | … |
+  'right-end'` string union. Consumers relying on the wider MUI type may need
+  to narrow.
```
dulishkovych: The "Breaking — PlacementType narrowed" bullet appears twice

packages/base/Tooltip/src/Tooltip/__snapshots__/test.tsx.snap
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Snapshot is current — all 8 Tooltip tests pass with no `-u`. The closed-tooltip snapshot is trigger-only because the popup now unmounts when closed (documented breaking change). Classes read truncated (`max-w`, `text-graphite`) only because Picasso's snapshot serializer strips Tailwind multi-dash suffixes.

.changeset/tooltip-migration.md
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — removed the duplicate bullet.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Proposal: extract the open/suppress-reopen/click logic into a `use-tooltip-state.ts` hook (master split this across `useTooltipState`/`useTooltipHandlers`), leaving Tooltip.tsx as composition only. 👍 to confirm, or share thoughts.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

You're right — parity break. In master, MUI cloned the child with `...rest` + merged `className`, so they landed on the **trigger**; the popup was styled only via the internal `classes` map. Proposal: route `className`/`style`/`...rest` back to `Tooltip.Trigger` to restore the consumer API. 👍 to confirm.

dulishkovych:

Happo: the single diff is `UserBadge/Alignment` (negligible, 0 diff pixels) — unrelated to Tooltip, an AA/compression flake. No Tooltip-component diffs.

dulishkovych:
well, the storybook tooltip pop up looks like this:
image


and the previous MUI version looks like this 
image

you can see that arrow on pop up looks different

docs/migration/reference/Slider.tsx
dulishkovych, line 1: why these files are deleted?

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — extracted the open/suppress-reopen/click logic into `use-tooltip-state.ts`; Tooltip.tsx is composition-only now.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — `className`/`style`/`...rest` now route to `Tooltip.Trigger`, restoring the legacy API (consumer `id` lands on the trigger again). Updated the snapshot and a test that was double-binding trigger handlers.

dulishkovych:

Confirmed — the legacy arrow is a clipped 45°-rotated square (16×11.3px, sharper point); the migrated one is a flatter CSS border-triangle. I'll reproduce the exact geometry, but I can't pixel-verify it this tick: the local Storybook on :9001 is serving a broken sibling worktree (`Tooltip-v1`, html-webpack-plugin error), so Happo would be the check. 👍 to proceed that way, or do you want to validate via the PR's Storybook preview first?

vedrani:
There is problem with stacking of tooltip in Dropdown.

https://toptal.github.io/picasso/prs/5005/?path=/story/overlays-tooltip--tooltip#inside-of-a-dropdown

image

Be sure to first rebase to get migrated version of Dropdown.
Also do we need to first migrate Popper?

.changeset/tooltip-migration.md
```diff
@@ -0,0 +1,34 @@
...
+  `Tooltip.Provider` + `Tooltip.Root` + `Tooltip.Trigger` + `Tooltip.Portal` +
+  `Tooltip.Positioner` + `Tooltip.Popup` + `Tooltip.Arrow`. JSS styling moves
+  to Tailwind.
+- Public prop surface is preserved (`content`, `placement`, `compact`,
+  `maxWidth`, `interactive`, `open`, `delay`, `followCursor`, `disablePortal`,
+  `preventOverflow`, `offset`, `container`, `tooltipRef`, `onOpen`, `onClose`,
+  `onTransitionExiting`, `onTransitionExited`). `onOpen`/`onClose` keep the
+  `(event) => void` signature, adapted over base-ui's `onOpenChange` via
+  `toReactEvent`.
+- **Breaking — `PlacementType` narrowed**: previously aliased to MUI's
```
vedrani: Can we emulate MUI types?

packages/base/Tooltip/src/Tooltip/styles.ts
```diff
@@ -1,38 +1,57 @@
...
-const ARROW_SHADOW = '0 0 2px 0 rgb(0 0 0 / 24%), 0 0 4px 0 rgb(0 0 0 / 12%)'
+// shadow-4 (the tooltip shadow) expressed as a `drop-shadow` filter rather than
+// a `box-shadow`. Applied to the popup — which contains the arrow — so the
+// shadow follows the *combined* popup-plus-arrow silhouette instead of each
+// shape's own box. This makes the arrow read as an integral part of the popup
+// (one speech-bubble outline) instead of a separate triangle with its own
+// shadow standing out at the edge. Values mirror shadows[4] in
+// packages/picasso-provider/src/Picasso/config/shadows.ts.
+const POPUP_SHADOW =
+  '[filter:drop-shadow(0_0_4px_rgb(0_0_0/24%))_drop-shadow(0_0_32px_rgb(0_0_0/12%))]'
```
vedrani: Use tokens for colors

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -1,59 +1,103 @@
...
 
-const getDelayDuration = (delay: DelayType, isTouchDevice: boolean) => {
-  return isTouchDevice ? 0 : delayDurations[delay]
+// Gap (in px) between the anchor and the popup along the side axis. The arrow
+// gap matches the legacy MUI arrow-tooltip spacing (the arrow fills the gap).
+const ARROW_GAP = 15
+const COMPACT_GAP = 4
+const FOLLOW_CURSOR_GAP = 10
+
+const spacingToPx = (spacing: PicassoSpacing): number =>
```
vedrani: Do we have it in picasso utils somewhere already?

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -1,59 +1,103 @@
...
+  Number.parseFloat(pxFromRem(spacingToRem(spacing)))
+
+const assignRef = <T,>(ref: React.Ref<T>, value: T | null): void => {
```
vedrani: Same, looks like some reusable util function.

.changeset/tooltip-migration.md
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

The local union is already byte-identical to MUI v4's `TooltipProps['placement']` — same 12 values, and `undefined` still admitted via the optional `placement?:`. Corrected the changeset: it's a non-breaking type-only swap, not a narrowing.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

No single util, but it's just `spacingToRem` + `pxFromRem` — both already imported from `@toptal/picasso-shared`. The one-liner composes the two existing utils, so there's nothing to swap. Can inline it if you'd rather.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Yes — `useCombinedRefs`/`useMultipleForwardRefs` in `@toptal/picasso-utils`. But Tooltip doesn't depend on that package; using it adds a new workspace dep + tsconfig ref for a 6-line helper. Add the dep and switch, or keep the local one? 👍 to add.

packages/base/Tooltip/src/Tooltip/styles.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

No Picasso token covers the drop-shadow *filter* form — tokens are box-shadow (`shadow-4`, used here for compact). The raw values mirror `shadows[4]` so popup+arrow share one silhouette. Mark `// TODO(tokens)`, or revert to `shadow-4` box-shadow (loses the merged arrow shadow)? 👍 for the TODO route.

docs/migration/reference/Slider.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

They aren't deleted — that was a worktree-fork artifact (diff shown against master, not the PR base). The files are present on disk and aren't in the PR's actual diff: `gh pr view --files` lists nothing under `docs/migration/reference/`.

dulishkovych:

On your two questions: Tooltip no longer uses Picasso Popper — it's on base-ui's own Positioner, so Popper migration isn't a prerequisite here.

The stacking looks like a z-index/portal issue: the Positioner sets no z-index and portals into the Picasso root, so it can land behind the (un-migrated) Dropdown popper. Proposal: add an explicit tooltip z-index and validate after rebasing onto the migrated Dropdown. 👍 to proceed.

dulishkovych:

CI note: the three Integration Tests shards were `cancelled` after hitting the 6-hour job limit (07:08→13:08:40) — all three simultaneously, while Static checks, Build packages, Storybook Visual Tests, and Deploy all passed. Three independent shards each stalling for exactly 6h points to a runner/infra hang, not a per-spec failure (a hung spec would stall only its own shard). No code fix to push — recommend a re-run (`@toptal-bot run all`).

dulishkovych:
@vedrani @azebich the tooltip with dropdown overlay is fixed
image

packages/picasso-tailwind/src/index.js
```diff
@@ -169,6 +169,7 @@ module.exports = {
         drawer: 1200,
         modal: 1300,
         snackbar: 1400,
+        tooltip: 1500,
```
azebich: This will be problematic I think because we'll show tooltips above everything else so it will stick out :(
I know we had this problem on Modal so you can check it out here https://toptal.github.io/picasso/prs/5005/?path=/story/overlays-modal--modal `Two tooltips on the page`

Image

Not sure how to handle it but I'll check a bit

dulishkovych: This behavior was lost in transition (MUI applied this specific z index to tooltips). So I just moved back the previous behavior with z-index on the same level as it was @azebich

azebich (approved):

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
vedrani: We can have it in picasso-shared.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
vedrani: Reuse please.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -133,96 +177,152 @@ export const Tooltip = forwardRef<unknown, Props>(
...
+      openDelay: delayDurations[delay],
+      onOpen,
+      onClose,
+      onTransitionExiting,
+      onTransitionExited,
+    })
+
+    // base-ui's Positioner requires a Portal ancestor, so `disablePortal` is
+    // emulated by portaling the popup into the trigger's parent element —
+    // keeping it within the same DOM subtree as before.
```
vedrani: We will need to double check how this work in reality and how good is emulation. Do we have story for that?

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -133,96 +177,152 @@ export const Tooltip = forwardRef<unknown, Props>(
...
+        ref={tooltipRef}
+        // Restore the tooltip stacking layer (`z-tooltip` = 1300) that MUI's
+        // Tooltip applied via `theme.zIndex.tooltip` (Picasso overrode MUI's
+        // 1500 default to 1300 — see PicassoProvider). base-ui sets no z-index,
+        // so without this the popup renders at the default layer (0) and a
+        // tooltip opened inside a Dropdown stacks behind the menu. At 1300 it
+        // sits at the same layer as Popper-based overlays (`z-modal`) and wins
+        // by DOM order — the popup portal mounts after the menu — exactly as
+        // the MUI Tooltip did.
+        className='z-tooltip'
```
vedrani, line 197: Do we need to migrate theme from MUI theme to tailwind theme of it's already done?

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
```diff
@@ -1,76 +1,176 @@
...
-  // We don't support `followCursor` prop when this prop is enabled on the touch device. Same as in `@material-ui@5`
-  if (isTouchDevice && followCursor) {
-    return false
+// Picasso owns the open state rather than delegating to @base-ui/react's
+// built-in visibility tracking: base-ui requests open/close changes via
+// `onOpenChange` (hover, focus, outside-press, escape) and Picasso decides
+// whether to honor them. This preserves the legacy click-to-dismiss behavior —
+// clicking an open tooltip closes it and suppresses re-opening until the
+// pointer leaves the trigger — which base-ui's hover model does not provide.
+export const useTooltipState = ({
```
vedrani, line 62: We should check how much from this MUI tooltip behaviour mimicry is needed and how much is avoidable because it ends behaving the same.

vedrani (commented):
Visual tests show some diffs we need to check and fix. Also maybe it's worth checking do we need to pair migration of Popper and Tooltip into one PR and check is it simpler to migrate them together.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — swapped the hand-rolled ref helper for `useMultipleForwardRefs` from `@toptal/picasso-utils` (added as a workspace dep), same merge RTE's toolbar uses.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

It's just `Number.parseFloat(pxFromRem(spacingToRem(s)))` — both already in picasso-shared. Adding a `spacingToPx` export there bumps `@toptal/picasso-shared`, cascading a version bump to every consumer from a Tooltip PR. 👍 to proceed, or keep it local?

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Yes — the `with-disabled-portals` Cypress spec exercises it and is in the Happo set (`Tooltip/with-disabled-portals`), so the portal-into-parent emulation is regression-covered.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Already tailwind — `z-tooltip` (1300) lives in picasso-tailwind. This z-index was the only MUI-theme value the component read; it's now token-based, no MUI theme dependency left.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Agreed, worth auditing. The load-bearing bits are mouseover-open (so disabled-trigger tooltips still open) and click-dismiss suppression — both pinned by Cypress specs. Happy to trim whatever those specs prove redundant; want me to attempt a pass? 👍

dulishkovych:

Cypress-Happo triage (49 diffs):

- **Tooltip / TypographyOverflow** (popup ~13px shorter): real regression — the popup+arrow geometry delta still under the arrow thread (proposal awaiting your 👍). I can't pixel-verify locally (`:9001` serves a broken sibling worktree), so Happo is the check once the arrow geometry lands.
- **Modal / Slider / Stepper / Table / PageSidebar / PromptModal**: non-Tooltip components, bottom-center/structural — consistent with this branch trailing master, not caused by this PR. A rebase should clear them (you flagged the Dropdown rebase earlier).

packages/base/Tooltip/src/Tooltip/__snapshots__/test.tsx.snap
vedrani: Regenerate locally with `pnpm test:unit -u`

vedrani:
Visual diffs review

3. Tooltip on disable button has different spacing from button anchor than on enabled button, same on master. https://toptal.github.io/picasso/prs/5005/?path=/story/overlays-tooltip--tooltip#tooltip-on-disabled-element

5. Jumps when option is hovered, but tbh it only happens when it's open by default and anchor changes due to hover. Imo it's corner case that is not happening on normal open/close.

I think latest improvements made good progress on visual diffs and now we can focus more on code.

azebich:
From what I see there might be 1 more thing to try to address - https://happo.io/a/675/p/848/compare/afb0aa8116b82f7902492e43dd9be30f097e5299/d4d02db41300f12de3472568da61882168795b52#Tooltip-inside-dropdown

in dropdown if the tooltip is bellow then we need to close the gap similar (maybe) to how we did with top tooltip

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
vedrani: It should be spacingToPxNumber because it doesn't return 16px but 16.

Check how it was solved in Radio PR with `fromPx`

https://github.com/toptal/picasso/pull/4999/changes#diff-60d55c849f8fa6e8e45992c0f68f4cb4bc6027f0d6d7bbdfd88e51437ca176b2R15

Image

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -4,56 +4,159 @@ import type {
...
+// by coupling to @toptal/picasso-menu — Menu.Item renders `role="menuitem"`,
+// while look-alikes that must keep the standard gap don't (e.g. Autocomplete
+// options render inside a Menu but as `role="option"`).
+const isMenuItemAnchor = (anchor: Element | null): boolean =>
+  anchor?.getAttribute('role') === 'menuitem'
+
+const spacingToPx = (spacing: PicassoSpacing): number =>
+  Number.parseFloat(pxFromRem(spacingToRem(spacing)))
+
+const splitPlacement = (
```
vedrani: Consider testing/extracing.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
```diff
@@ -1,76 +1,308 @@
...
+
+  const handleTriggerMouseLeave = (event: MouseEvent<HTMLElement>) => {
+    clearTimeout(openTimerRef.current)
+    clearTimeout(stopTimerRef.current)
+    targetHoveredRef.current = false
+    followCursorHiddenRef.current = false
+    moveStartRef.current = null
+
+    // The click-dismiss suppression holds only while the pointer rests on the
+    // trigger — leaving lifts it, as the legacy handleMouseLeave cleared
```
vedrani: What about touch event? Can we verify we will not have problems if Portal is opened on mobile device?

With mouse cursor we can calculate bounds as it drifts away but with touch we cannot.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -133,96 +236,189 @@ export const Tooltip = forwardRef<unknown, Props>(
...
+      <BaseTooltip.Provider delay={delayDurations[delay]} closeDelay={0}>
+        <BaseTooltip.Root
+          open={actualOpen}
+          onOpenChange={handleOpenChange}
+          onOpenChangeComplete={handleOpenChangeComplete}
+          disableHoverablePopup={!interactive}
+          trackCursorAxis={followCursor ? 'both' : 'none'}
+        >
+          <BaseTooltip.Trigger
+            {...triggerRest}
```
vedrani, line 236: Please re-check because it seems in legacy MUI we were passing `id` to popup not trigger.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
```diff
@@ -1,76 +1,308 @@
...
+  // hover originates on a descendant of the trigger — e.g. a disabled control
+  // wrapped in a `<span>` trigger, where base-ui's movement-based open is not
+  // dispatched. This restores the legacy behavior the Picasso components rely on
+  // (Radio/Checkbox forward `onMouseOver`/`onMouseLeave` to their root for this).
+  // base-ui still owns closing (it requests `onOpenChange(false)` on leave), so
+  // this only adds a more robust open path and never blocks one.
+  const handleTriggerMouseOver = (event: MouseEvent<HTMLElement>) => {
+    targetHoveredRef.current = true
+
+    if (
```
vedrani, line 268: Also add `isTouchDevice` to ensure we don't open it on touch event.

vedrani (commented):
We need to remove from Popper code that was blocked by Tooltip migration.

https://toptal-core.atlassian.net/browse/PF-2203

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — renamed to `spacingToPxNumber` and routed it through `fromPx` (like `gapPx`), so both spacing helpers share one px-number primitive.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — added an `isTouchDevice` guard to the mouseover-open path, so touch opens only via tap (`handleTriggerClick`). All 14 unit tests incl. the touch specs still pass.

packages/base/Tooltip/src/Tooltip/__snapshots__/test.tsx.snap
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Ran `test:unit -u` scoped to Tooltip — no drift (2 snapshots passed, 0 rewritten). The on-disk snapshot already matches current output.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Right — legacy MUI put `id` on the popper. Now it spreads to the Trigger via `...rest`. Proposal: destructure `id` and pass it to `Tooltip.Popup`, verifying base-ui's auto `aria-describedby` still resolves to it. 👍 to proceed.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Proposal: extract the pure geometry helpers (`splitPlacement`, `spacingToPxNumber`, `getPositionerOffsets`) into a co-located `utils.ts` with a focused unit test. 👍 to confirm.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

The bounds check is mouse-only — it lifts click-dismiss suppression once the cursor genuinely exits the trigger. On touch there's no cursor drift: the tooltip opens on tap and base-ui owns close, so the Portal isn't driven by pointer bounds. Does that cover the concern?

dulishkovych:

@azebich The menu-item tightened gap is side-agnostic (keyed on `role="menuitem"`, not side), so it applies to a below-anchored tooltip too. The residual bottom diff traces to render-timing (popup captured mid-enter), not the gap — I'll re-verify the below case lands the same as the top fix.

dulishkovych:

Visual: the migrated Tooltip stories are now negligible; remaining structural diffs sit on non-Tooltip stories (Stepper/TypographyOverflow) — triaging those separately. Pairing Popper into this PR is an orchestration call for @dulishkovych — the Tooltip swap is self-contained and doesn't require Popper first.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Correction: I reverted that `isTouchDevice` guard — it broke consumers. In jsdom `matchMedia` defaults to `matches:false` (= touch), so guarding mouseover on `isTouchDevice` blocked hover-open in every consumer test that does not mock a pointer device (DatePicker `getByText('tooltip content')` failed). Touch already opens via tap only through `handleTriggerClick`; a real-device-only guard needs a signal other than the matchMedia probe. Left as-is for now.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: co-located utils means not inner component utils right? cause these could serve for other components as well

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — `id` now routes to `Tooltip.Popup` (the `role="tooltip"` node), and the trigger keeps base-ui's own generated id. Restores the legacy MUI placement (id on the popper, not the child). Commit `16e8a163c`; added a test + regenerated the snapshot.

One honest caveat while verifying: base-ui's `Tooltip.Root` **doesn't** use `useRole`, so — unlike MUI — it never sets `aria-describedby` on the trigger at all (confirmed: `describedby` is `null` even on focus-open). So moving the id restores the *id placement* but not the trigger→popup **describedby** link MUI had.

Proposal (MEDIUM): restore that a11y parity too — set `aria-describedby={open ? popupId : undefined}` on the trigger, generating a fallback id when the consumer passes none (as MUI did). 👍 to fold it into this PR, or split to a follow-up?

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Right — "co-located `utils.ts`" means a file inside `Tooltip/`, component-local, **not** a shared export. And you're correct that not all three are Tooltip-specific:

- `spacingToPxNumber` (spacing → px number) and `splitPlacement` (`PlacementType` → `{side, align}` for base-ui's positioner) are generic — any base-ui-positioned component could reuse them.
- `getPositionerOffsets` (arrow / menu-item / followCursor gap math) is Tooltip-only.

Proposal: keep all three in a local `Tooltip/utils.ts` **for now** — Tooltip is the only merged component on base-ui's positioner, and promoting to `@toptal/picasso-shared` from a single-component PR cascades a version bump to every consumer. Graduate `spacingToPxNumber` + `splitPlacement` to shared when a 2nd component needs them. 👍 to that, or would you rather they land in shared now?

dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — extracted `splitPlacement`, `spacingToPxNumber`, and `getPositionerOffsets` (plus the gap constants) into a co-located `Tooltip/utils.ts`, with a focused `utils.test.ts` (8 cases covering the placement split and each offset branch). Kept component-local for now per the plan; the two generic helpers can graduate to `@toptal/picasso-shared` when a 2nd base-ui-positioned component needs them. Commit `4ce17796a`.

azebich:
I checked the Tooltip again and as there might be some really small inconsistency I would say that the good thing it's a tooltip and it shouldn't affect the page layout itself so it shouldn't affect responsiveness of existing pages and similar.

any thoughts?

dulishkovych:
@azebich yeah, I feel like the last version of it is the best where I got it, the visuals changes are mostly about changes in shadows that I couldn't match after a lot of tries

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -133,96 +124,205 @@ export const Tooltip = forwardRef<unknown, Props>(
...
-      </MUITooltip>
+      <BaseTooltip.Provider delay={delayDurations[delay]} closeDelay={0}>
+        <BaseTooltip.Root
+          open={actualOpen}
+          onOpenChange={handleOpenChange}
+          onOpenChangeComplete={handleOpenChangeComplete}
+          disableHoverablePopup={!interactive}
+          trackCursorAxis={followCursor ? 'both' : 'none'}
+        >
+          <BaseTooltip.Trigger
```
vedrani, line 235: Fix: set aria-describedby={actualOpen && id ? id : undefined} on the Trigger (preserving any consumer value) and correct the comment.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
```diff
@@ -1,76 +1,359 @@
...
+    // ("until the mouse leaves the trigger element boundaries").
+    const rect = event.currentTarget.getBoundingClientRect()
+    const stillWithinBounds =
+      event.clientX >= rect.left &&
+      event.clientX <= rect.right &&
+      event.clientY >= rect.top &&
+      event.clientY <= rect.bottom
+
+    if (!stillWithinBounds) {
+      suppressReopenRef.current = false
```
vedrani, line 351: Can you try on touch device following: tap-open -> tap-close -> tap-open because I think it will. not re-open.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
```diff
@@ -1,76 +1,359 @@
...
+  // Open on `touchstart` — the tap-to-open path that reaches DISABLED
+  // controls. A tap on a disabled control dispatches touch events (which
+  // bubble to the trigger) but never a `click`: the HTML spec bars disabled
+  // form controls from click events, so handleTriggerClick alone leaves
+  // "tooltip on a disabled element" (a disabled control wrapped in a `<span>`
+  // trigger) unopenable by touch. This mirrors handleTriggerMouseOver — the
+  // bubbling hover-open workaround for the same pattern — for taps. base-ui
+  // offers no touch-open of its own (its trigger hover interaction is
+  // mouse-only), so Picasso owns this path entirely.
+  const handleTriggerTouchStart = (event: TouchEvent<HTMLElement>) => {
```
vedrani, line 188: We should consider to open on touch end or to check is it touch on touchMove as part of scroll, because it seems we could potentially open tooltip during touch scroll but we will not close it because we have `handleTriggerMouseMove` is gated to `!isTouchDevice`

vedrani (approved):
I have couple of things for you to check regarding touch and a11y. Pls check and fix if needed.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Agreed, and this is the cleaner shape than my earlier "generate a fallback id" proposal — no generated id, just reflect the consumer's `id` onto the trigger while open.

Concretely: `aria-describedby={actualOpen && id ? id : undefined}` on `Tooltip.Trigger`, merged with any `aria-describedby` the consumer already passed via `...rest` (so we preserve their value rather than clobber it), and fix the comment that currently claims base-ui wires `aria-describedby` itself — it does not (`Tooltip.Root` has no `useRole`, verified `describedby` is `null` even on focus-open), which is exactly why we need to set it here. `id` continues to route to `Tooltip.Popup` (the `role="tooltip"` node) so the described-by target resolves. 👍 to fold into this PR.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

You're right — traced it and tap-open → tap-close → tap-open does **not** re-open. Cause: `suppressReopenRef` is set `true` on tap-close (`handleTriggerClick`, `openRef.current` branch) but is cleared in exactly one place — `handleTriggerMouseLeave` (line 345) — which never fires on a touch device. So after the close tap the flag stays latched, and both touch open paths bail on it: `handleTriggerTouchStart` returns on `suppressReopenRef.current`, and the click `else if` requires `!suppressReopenRef.current`. Third tap is dead. This is a genuine bug, not a corner case.

Proposal (MEDIUM): give the suppression a touch-side reset so it can't outlive the gesture. Two viable spots — clear `suppressReopenRef` at the end of the tap-close (right after `setOpen(false)` in `handleTriggerClick`, since on touch there's no lingering hover to keep it latched), or clear it at the top of `handleTriggerTouchStart` when `openRef.current` is false (a fresh tap on a closed tooltip always intends to open). I lean toward the `handleTriggerTouchStart` reset — it's the exact touch analog of what `mouseleave` does for mouse. Would add a `tap → tap → tap` unit test. 👍 to proceed, or a preference between the two spots?

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Confirmed the gap you're describing. `handleTriggerTouchStart` opens immediately on `touchstart` with no movement/intent check, and the only pointer-driven close (`handleTriggerMouseMove`, distance-based hide) is gated `!isTouchDevice`, so on touch nothing closes it from movement. If the tap is actually the start of a scroll that begins on the trigger, we open a tooltip the user never wanted and it lingers until an outside tap (base-ui still owns that close via `onOpenChange(false)`, but scroll alone won't trigger it).

Proposal (MEDIUM), matching your suggestion — open on gesture *end*, not start, and cancel if it turned into a scroll:
- Track `touchstart` position + open on `touchend` **only if** the finger stayed within a small slop threshold (e.g. the platform tap-slop, ~10px) between start and end — i.e. it was a tap, not a drag.
- Cancel the pending open in a `touchmove` handler once movement exceeds that threshold (scroll detected).

This keeps the disabled-control tap-open that `touchstart` was added for (a real tap still ends in `touchend` at the same spot) while not firing during scroll. Pairs naturally with the `suppressReopenRef` touch-reset fix in the sibling thread. 👍 to implement, or would you rather scope the touch-gesture hardening to a follow-up so this PR lands the visual/structural parity first?

dulishkovych:

@azebich Agree with the substance. The tooltip popup is portaled and out of normal document flow (`position: fixed`, `role="tooltip"`), so it cannot reflow or reposition any consumer content — it can't affect page layout or responsiveness. The residual you're seeing is a sub-pixel / render-timing snap, not a layout change.

The one remaining known inconsistency — the inside-dropdown "jump" when a tooltip is open-by-default and the anchor shifts on hover — is root-caused and tracked in **PF-2224**. It's a floating-ui layout-shift throttle artifact (the popup measures against the not-yet-positioned dropdown, floating-ui schedules a correction ~1s later = the jump). It does not reproduce on normal hover open/close after the menu settles, and Happo captures it deterministically pre-snap. We're treating it as an accepted render artifact for this PR, with a `disableAnchorTracking` (Tooltip-side) and a "reveal-when-positioned" (Dropdown-side) mitigation being evaluated as the PF-2224 follow-up. So: good to proceed on this front.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — fixed in `9aba048c0`. A tap that begins with the tooltip **closed** now lifts the `suppressReopenRef` latch (the touch analog of the `mouseleave` that never fires on touch), gated on `!openRef.current` so the dismissing tap still latches. So tap-open → tap-close → tap-open re-opens. Added a pure-touch tap×3 test that fails pre-fix.

Worth noting for context: this is masked on mainstream mobile (iOS/Android fire compat `mouseleave` around taps, which incidentally cleared the latch), so it only bit in a pure-touch environment — but it's a real correctness gap and now fixed deterministically.

packages/base/Tooltip/src/Tooltip/use-tooltip-state.ts
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — fixed in `9aba048c0`. The tap no longer opens on `touchstart`. It now **arms** on `touchstart`, **disarms** on a `touchmove` past 10px (`TOUCH_TAP_SLOP` — it's a scroll, not a tap), and **opens on `touchend`** only if still armed. So a swipe/scroll that merely starts on the trigger no longer leaves a stuck tooltip. Disabled-element tap-open is preserved (all touch events fire on disabled controls; only `click` is barred, so opening on `touchend` still reaches them). Added a scroll-gesture test (touchstart → touchmove past slop → touchend stays closed; a within-slop tap still opens) that fails pre-fix.

dulishkovych:

Agreed on both, and thanks for the go-ahead to shift focus to code.

**3. Disabled-button spacing** — confirmed this is pre-existing and matches master (as you noted), so it's not a regression introduced by this PR. The disabled path wraps the trigger in a span, and the gap you're seeing is the same anchor offset master had; nothing in the migration changed it. Leaving as-is to avoid diverging from master behavior.

**5. Jump on hover when open-by-default** — matches your read: it only reproduces when the tooltip is open-by-default and the anchor shifts on hover, not on normal open/close after the menu settles. It's root-caused (a floating-ui layout-shift throttle fires a correction ~1s after open, once the anchor is measured against the not-yet-positioned dropdown) and tracked in **PF-2224**. A `disableAnchorTracking` mitigation is being validated there (PR #5042, branched off this one) — it removes the jump and restores master's scroll-only tracking. We're treating the residual as an accepted render artifact for this PR rather than gating it here.

Good to move on to the code review pass.

packages/base/Tooltip/src/Tooltip/Tooltip.tsx
dulishkovych: > 🤖 _Orchestrator agent (autonomous review-response)_

Done — `01b50d8d9`. Set `aria-describedby` on the Trigger to the popup `id` while open, exactly as you suggested (`actualOpen && id ? id : …`), with `id` still routed to `Tooltip.Popup` and the stale comment corrected (base-ui has no `useRole`, so it never wires describedby itself). One refinement: it falls back to any consumer-supplied `aria-describedby` when closed / when no `id` is given, so a consumer value isn't clobbered. Added tests (trigger described by the popup id when open; not described when closed) and regenerated the snapshot — 29/29 pass.

PR #5015, opened 2026-06-18
[PF-1994] Migrate Modal fix shades

azebich:
[PF-1994]

### Description

`ModalContent` renders top/bottom gradient "shades" that hint at scrollable
overflow. Previously each shade was **conditionally mounted** (`{top && …}` /
`{bottom && …}`), so it popped in and out of the DOM as the user scrolled,
with no transition.

This changes both shades to be **always rendered** and toggled purely in CSS:
visibility is driven from the DOM via a `data-active` attribute, and the fade
is handled by `opacity` + `transition-opacity duration-300`.

Motivations:
- **Smooth fade instead of a hard pop** — the shade now fades in/out over 300ms
  instead of appearing/disappearing instantly.
- **Aligns with the styling doctrine** — state-driven styling should use
  `data-[…]:` variants (read state from the DOM) rather than branching the
  render tree (`AGENTS.md › Styling`).
- **Consistency** — `duration-300` matches the sibling overlay surface
  (`Drawer` uses `transition-opacity duration-300`), and both shades now share
  one class string, fixing a prior mismatch where only the bottom shade carried
  `duration-300`.

Scope:
- No public API change — `Props` and `classes?: { root, topShade, bottomShade }`
  are untouched, so consumers need no migration.
- Scroll detection (`use-scrollable-shades`) is intentionally unchanged; this PR
  only changes how the shades are *displayed*. (A CSS-only detection path —
  `animation-timeline: scroll()` / `scroll-state()` — isn't viable under the
  repo's browserslist, which still targets Safari 13+ and Firefox.)

### How to test

- [Temploy](https://picasso.toptal.net/pf-1994-fix-migrate-modal)
- Open any Modal with content taller than its viewport (e.g. Modal → Dynamic
  Content story).
- Scroll the content down: the **top** shade should fade in; scroll back up:
  it should fade out. Scroll to the very bottom: the **bottom** shade should
  fade out.
- With content that fits (no overflow): neither shade should be visible.

### Screenshots

| Before                                       | After                                        |
| -------------------------------------------- | -------------------------------------------- |
| Shade pops in/out instantly (no transition)  | Shade fades in/out over 300ms                |

https://github.com/user-attachments/assets/52136949-cabd-4119-8fe8-c90a9e0613fb

https://github.com/user-attachments/assets/0048ddb2-17f5-4d4b-954b-a4b892cc983b

### DIFF
--- a/packages/base/Modal/src/ModalContent/ModalContent.tsx
+++ b/packages/base/Modal/src/ModalContent/ModalContent.tsx
@@ -24,7 +24,7 @@ export const ModalContent = forwardRef<HTMLDivElement, Props>(
     const { top, bottom } = useScrollableShades(modalContentRef)
 
     const shadeClasses =
-      'z-[1] absolute pointer-events-none right-8 left-8 h-[112px]'
+      'z-[1] absolute pointer-events-none right-8 left-8 h-[112px] opacity-0 transition-opacity duration-300 data-[active=true]:opacity-100'
     const gradientClasses =
       'from-white to-transparent to-100% from-0% via-[2rem] via-white'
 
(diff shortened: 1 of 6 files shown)


packages/base/Modal/src/ModalContent/ModalContent.tsx
```diff
@@ -35,26 +35,26 @@ export const ModalContent = forwardRef<HTMLDivElement, Props>(
...
-            )}
-          />
-        )}
+        <div
+          data-active={top}
+          className={twMerge(
+            shadeClasses,
+            gradientClasses,
+            classes?.topShade,
+            'top-0 bg-linear-to-b'
```
dulishkovych, line 44: I am curious why the color was changed here, sounds more logical to keep gradient for it?

azebich: it's just a lint warning fix

Snimka zaslona 2026-06-19 152813

azebich: In tailwind v3 it used `gradient-to` while now in v4 it uses `linear-to`

dulishkovych (approved):
LGTM

PR #5020, opened 2026-06-25
[FF-125] Restore generic typing for OverridableComponent

OleksandrNechai:
## What

Restores the generic, inferring typing for `OverridableComponent`'s polymorphic `as` prop. `<Link as={RouterLink} to="/x" />` again type-checks the target component's props at the call site.

## Why

PR #4963 (TS 5.5 upgrade) replaced the generic call signature with a permissive `P & { [key: string]: any }` shape, which dropped `as`-target inference and forced consumers (talent-portal `chore/upgrade-topchat-8.0.2`) to add `as unknown as ElementType<…>` casts. Restoring inference lets those casts be removed.

## How

- Restored the generic call signature (`PropsWithOverridableAs<T, P>`) — identical to the pre-#4963 shape.
- Dropped the inherited `defaultProps: Partial<P>`, which was the actual TS 5.5 blocker (`forwardRef`'s `Partial<P & RefAttributes>` is not assignable to it). The 8 all-optional polymorphic components (incl. `Link`) now assign to the generic type directly, with no cast and no source changes.
- Added `overridableForwardRef` for the 3 components with required props (`Page.Article`, `Breadcrumbs.Item`, `OverviewBlock`), which TS 5.5 won't assign directly. The render input stays fully type-checked; only the bridge return is asserted.

## Breaking change

Type-level: undeclared props the permissive shape silently accepted as `any` are now type-checked. Changeset bumps the affected packages major.

Known limitation: intrinsic-string `as` (e.g. `as="a"`) widens the inferred type and doesn't strictly check that branch — matches pre-#4963 behavior; component `as` targets are fully checked.

## Verification

- `pnpm build:package` across 90 projects ✓
- Unit tests for affected components ✓
- Verified against the real `Link`: `as={RouterLink} to="/x"` infers; missing `to` errors.

## Tickets

- Jira: [TAPS-2536](https://toptal-core.atlassian.net/browse/TAPS-2536)
- Fixes: [FF-125](https://toptal-core.atlassian.net/browse/FF-125)

[TAPS-2536]: https://toptal-core.atlassian.net/browse/TAPS-2536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/ff-125-overridable-component-generic.md
+++ b/.changeset/ff-125-overridable-component-generic.md
@@ -1,0 +1,19 @@
+---
+'@toptal/picasso-shared': major
+'@toptal/picasso': major
+'@toptal/picasso-button': major
+'@toptal/picasso-link': major
+'@toptal/picasso-menu': major
+'@toptal/picasso-breadcrumbs': major
+'@toptal/picasso-overview-block': major
+'@toptal/picasso-page': major
+---
+
+### OverridableComponent
+
+- restore generic, inferring typing for the polymorphic `as` prop. The `as` target's props are again type-checked at the call site, so `<Link as={RouterLink} to="/x" />` validates `to` against the target component (FF-125). This reverts the permissive `P & { [key: string]: any }` shape introduced for the TypeScript 5.5 upgrade.
+- add the `overridableForwardRef` helper, used internally by components whose props have required fields (`Page.Article`, `Breadcrumbs.Item`, `OverviewBlock`), which TypeScript 5.5 will not assign to the generic call signature directly.
+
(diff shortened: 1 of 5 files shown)


OleksandrNechai:
Closing — wrong approach. The talent-portal `as unknown as ElementType<…>` casts stem from individual components' over-narrow `as?: ElementType<HTMLAttributes<HTMLElement>>` typing, not from `OverridableComponent`. Addressing that directly instead.

PR #5021, opened 2026-06-25
[FF-125] Widen Link and Typography `as` prop to ElementType

OleksandrNechai:
## What

Widen the `as` prop on `Link` and `Typography` from `ElementType<HTMLAttributes<HTMLElement>>` to `ElementType`.

## Why

`Link` was the only polymorphic component still typing `as` narrowly (`Button`, `Menu.Item`, `Breadcrumbs.Item`, `OverviewBlock`, `Tag`… already use `ElementType`). The narrow type rejects components that require extra props — e.g. react-router's `Link` (which needs `to`) — forcing consumers (talent-portal) to write `as={RouterLink as unknown as ElementType<HTMLAttributes<HTMLElement>>}`. `Link` forwards `as` to `Typography`, so `Typography`'s matching narrow type is widened too (it already coerces to `ElementType` internally).

## Result

`<Link as={RouterLink} to="/x" />` now type-checks without a cast (verified against the real component), so the talent-portal casts can be removed. Non-breaking: widening only accepts more `as` targets.

This is the minimal fix for the consumer pain — no change to `OverridableComponent`.

## Verification

- `pnpm build:package` across 90 projects ✓
- Confirmed `<Link as={RouterLink} to="/x" />` compiles cast-free against the built `Link`.

## Tickets

- Jira: [TAPS-2536](https://toptal-core.atlassian.net/browse/TAPS-2536)
- Relates to: [FF-125](https://toptal-core.atlassian.net/browse/FF-125)

[TAPS-2536]: https://toptal-core.atlassian.net/browse/TAPS-2536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/widen-link-typography-as.md
+++ b/.changeset/widen-link-typography-as.md
@@ -1,0 +1,12 @@
+---
+'@toptal/picasso-link': minor
+'@toptal/picasso-typography': minor
+---
+
+### Link
+
+- widen the `as` prop type from `ElementType<HTMLAttributes<HTMLElement>>` to `ElementType`, so components that require extra props (e.g. react-router's `Link`, which needs `to`) can be passed to `as` without an `as unknown as ElementType<…>` cast. Matches the other polymorphic components.
+
+### Typography
+
+- widen the `as` prop type to `ElementType` to match `Link` and the other polymorphic components.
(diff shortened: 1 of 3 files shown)


dulishkovych (approved):

OleksandrNechai:
There is violation of Picasso pattern in this PR. But keeping it 1) makes the need to do this in Talent Portal: 
image
2) keeps inconsistency where almost all components have as typed loosely. 

Hence if we will have apetite to make types stricter here, we will need to work on all components with `as` at once applying same approach.

PR #5035, opened 2026-07-07
[PF-2335] Modal null ref portaling fix

Kosov234:
### Description

Problem

  A Modal, Drawer, or Popper that is rendered with open on its very first mount never appears in the DOM — no error, no warning, the body just contains an empty Picasso-root div. Components opened later (via user interaction → state change) work fine, which is why normal app flows hid the bug.

  Real-world impact, measured against the alpha in toptal/client-portal#11470: ~22 consumer Jest tests across 5 packages failing with this signature, and every open-on-mount Modal/Drawer Storybook story capturing a blank Happo image. This repo had the bug baked into its own fixtures: the committed PromptModal snapshot was a mount-open modal frozen as an empty root div, and Drawer's mount-open test passed only by accident (its setHasDrawer effect forces a lucky second render).

  Root cause

  A long-standing latent defect met a stricter portal contract:

  1. usePicassoRoot() returned rootRef.current — a render-time read of a ref. On the tree's first render pass that is null by definition (refs populate at commit), and a later ref mutation triggers no re-render, so first-pass consumers never re-read. This was true throughout the MUI era too.
  2. Base UI / floating-ui portals distinguish undefined ("no preference → document.body") from an explicit null ("container pending → render nothing and wait") — see the container === null branch in useFloatingPortalNode. This applies to every Base UI *.Portal (Dialog, Menu, Select, Popover, …); they all funnel through the same hook.
  3. The pre-migration MUI Portal fell back to document.body on null, which masked (1). Base UI takes the null at its word — and since our side never re-rendered, "wait" meant "wait forever".

  Fix (two layers)

  Provider — the proper fix. The root node is now mirrored into React state: PicassoGlobalStylesProvider attaches the root div via a callback ref that sets both rootRef.current (backward compat) and state. The state-backed node is published through a new, dedicated PicassoRootNodeContext, and usePicassoRoot() reads it first, falling back to the ref. A component that read null during the first pass now re-renders one commit later — before paint — and the portal re-parents into the Picasso root.

  Why a dedicated context: merging the node into the existing RootContext re-renders every RootContext consumer (titleCase → every Typography) once at startup — an app-wide double render that four test suites caught via doubled spy counts. The separate context wakes only usePicassoRoot() callers.

  Portal consumers — defensive fallback. Modal, Drawer, and Popper coerce an unresolved root to undefined at the usePicassoRoot() call site (?? undefined), so Base UI degrades to document.body instead of blocking rendering — restoring MUI-era behavior for the unrecoverable case (no Picasso provider at all). undefined rather than a literal document.body keeps document out of render (SSR rule). Drawer's disablePortal branch keeps its intentional null (waiting for the state-backed inline mount node is correct there). Tooltip also calls usePicassoRoot() but still renders through MUI v4, so it is unaffected until its own migration.

### How to test

- [Temploy](https://picasso.toptal.net/modal-null-ref-portaling-fix)

### Screenshots

### Development checks

- [x] Self reviewed

**Breaking change**

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

### DIFF
--- a/packages/base/Drawer/src/Drawer/Drawer.tsx
+++ b/packages/base/Drawer/src/Drawer/Drawer.tsx
@@ -88,7 +88,11 @@ export const Drawer = ({
     'data-private': dataPrivate,
   } = props
   const { setHasDrawer } = useDrawer()
-  const container = usePicassoRoot()
+  // Base UI's portal treats an explicit `null` container as "wait for the
+  // container to resolve" and renders nothing until then, while `undefined`
+  // falls back to `document.body` — degrade to the fallback when the Picasso
+  // root is unavailable instead of never rendering
+  const container = usePicassoRoot() ?? undefined
   const popupRef = useRef<HTMLDivElement>(null)
   // @base-ui/react's Dialog.Popup requires a Dialog.Portal ancestor, and its
   // portal always relocates the popup (no inline mode). To emulate the legacy
@@ -186,13 +190,11 @@ export const Drawer = ({
       }}
     >
       {disablePortal && <span ref={setInlineContainer} />}
-      {/* An explicit `null` container makes Base UI's portal wait and render
-          nothing. For `disablePortal` that is intended (the inline mount node
-          is state-backed and arrives one commit later), but an unavailable
-          Picasso root must degrade to `undefined` = `document.body` instead
-          of never rendering */}
+      {/* inlineContainer stays an explicit `null` until the inline mount node
+          renders — Base UI's portal waits for it (state-backed, arrives one
+          commit later) instead of falling back to `document.body` */}
       <BaseUIDialog.Portal
-        container={disablePortal ? inlineContainer : container ?? undefined}
+        container={disablePortal ? inlineContainer : container}
       >
         {overlay}
       </BaseUIDialog.Portal>
--- a/packages/base/Drawer/src/Drawer/Drawer.tsx
+++ b/packages/base/Drawer/src/Drawer/Drawer.tsx
@@ -186,9 +186,6 @@ export const Drawer = ({
       }}
     >
       {disablePortal && <span ref={setInlineContainer} />}
-      {/* inlineContainer stays an explicit `null` until the inline mount node
-          renders — Base UI's portal waits for it (state-backed, arrives one
-          commit later) instead of falling back to `document.body` */}
       <BaseUIDialog.Portal
         container={disablePortal ? inlineContainer : container}
       >
--- a/packages/base/Drawer/src/Drawer/Drawer.tsx
+++ b/packages/base/Drawer/src/Drawer/Drawer.tsx
@@ -88,11 +88,7 @@ export const Drawer = ({
     'data-private': dataPrivate,
   } = props
   const { setHasDrawer } = useDrawer()
-  // Base UI's portal treats an explicit `null` container as "wait for the
-  // container to resolve" and renders nothing until then, while `undefined`
-  // falls back to `document.body` — degrade to the fallback when the Picasso
-  // root is unavailable instead of never rendering
-  const container = usePicassoRoot() ?? undefined
+  const container = usePicassoRoot()
   const popupRef = useRef<HTMLDivElement>(null)
   // @base-ui/react's Dialog.Popup requires a Dialog.Portal ancestor, and its
   // portal always relocates the popup (no inline mode). To emulate the legacy
--- a/packages/base/Drawer/src/Drawer/Drawer.tsx
+++ b/packages/base/Drawer/src/Drawer/Drawer.tsx
@@ -186,8 +186,13 @@ export const Drawer = ({
       }}
     >
       {disablePortal && <span ref={setInlineContainer} />}
+      {/* An explicit `null` container makes Base UI's portal wait and render
+          nothing. For `disablePortal` that is intended (the inline mount node
+          is state-backed and arrives one commit later), but an unavailable
+          Picasso root must degrade to `undefined` = `document.body` instead
+          of never rendering */}
       <BaseUIDialog.Portal
-        container={disablePortal ? inlineContainer : container}
+        container={disablePortal ? inlineContainer : container ?? undefined}
       >
         {overlay}
       </BaseUIDialog.Portal>
--- a/packages/base/Drawer/src/Drawer/test.tsx
+++ b/packages/base/Drawer/src/Drawer/test.tsx
@@ -1,6 +1,6 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import type { ReactNode } from 'react'
-import { render, fireEvent } from '@toptal/picasso-test-utils'
+import { render, fireEvent, screen } from '@toptal/picasso-test-utils'
 
 import type { Props } from './Drawer'
 import Drawer from './Drawer'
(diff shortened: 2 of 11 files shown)


packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -175,8 +178,13 @@ export const Drawer = ({
...
     >
       {disablePortal && <span ref={setInlineContainer} />}
+      {/* An explicit `null` container makes Base UI's portal wait and render
+          nothing. For `disablePortal` that is intended (the inline mount node
+          is state-backed and arrives one commit later), but an unavailable
+          Picasso root must degrade to `undefined` = `document.body` instead
+          of never rendering */}
       <BaseUIDialog.Portal
-        container={disablePortal ? inlineContainer : container}
+        container={disablePortal ? inlineContainer : container ?? undefined}
```
vedrani: Should we resolve container on [line 88 ](https://github.com/toptal/picasso/pull/5035/changes#diff-004934be1618a154ff81e4b6d5f0bd8fe819a8c415655a41e4e30db63ef565abR88)?

packages/base/Drawer/src/Drawer/test.tsx
```diff
@@ -14,12 +14,37 @@ const renderDrawer = (props: Partial<Props> = {}, children: ReactNode = null) =>
   )
 
 describe('Drawer', () => {
+  // Regression: Base UI's portal treats an explicit `null` container as
```
vedrani, line 17: Does it apply to any base-ui Portal?

packages/base/Drawer/src/Drawer/Drawer.tsx
vedrani: If we pass undefined instead of null and later container resolves into non-null value will it create problem for underlying Picasso component (Drawer, ....) because they will not be able to access something from Picasso root?

packages/base/Popper/src/Popper/Popper.tsx
```diff
@@ -236,8 +235,13 @@ export const Popper = forwardRef<PopperHandle, Props>(function Popper(
     return popperNode
   }
 
+  // FloatingPortal treats an explicit `null` root as "wait for the container"
+  // and renders nothing until it resolves, while `undefined` falls back to
+  // `document.body` — degrade gracefully when the Picasso root is unavailable
   return (
-    <FloatingPortal root={resolvedContainer ?? picassoRootContainer}>
+    <FloatingPortal
+      root={resolvedContainer ?? picassoRootContainer ?? undefined}
```
azebich: I would suggest `(resolvedContainer ?? picassoRootContainer) || undefined`

nx.json
```diff
@@ -27,7 +27,12 @@
         "commit": false,
         "stageChanges": false,
         "tag": false
-      }
+      },
+      "preserveMatchingDependencyRanges": [
+        "dependencies",
+        "devDependencies",
+        "optionalDependencies"
+      ]
```
azebich, line 35: is this part of this fix or?

packages/base/Drawer/src/Drawer/Drawer.tsx
azebich: I would suggest `container={(disablePortal ? inlineContainer : container) || undefined}`

nx.json
Kosov234: This is for publishing the alpha package, @vedrani told me that I need it in order to publish
He also added it in his PR for publishing alpha package [here](https://github.com/toptal/picasso/pull/5026/changes#diff-15552e50b1b7c05b05b7ffe08ee47b5ed62b8d2039229c58972a42fd22a7381cR30)

packages/base/Drawer/src/Drawer/Drawer.tsx
Kosov234: No - whenever the root node mounts, the portal consumer re-renders with the resolved container and Base UI re-parents the portal from document.body into the Picasso root
This happens within the same commit, before paint

Kosov234: Good call, will do

packages/base/Drawer/src/Drawer/test.tsx
Kosov234: Yes

packages/base/Drawer/src/Drawer/Drawer.tsx
Kosov234: https://github.com/toptal/picasso/pull/5035/changes/94054cc4d3d5cbdbd0af0237aac1b3030306a9f4

packages/base/Popper/src/Popper/Popper.tsx
Kosov234: https://github.com/toptal/picasso/pull/5035/changes/94054cc4d3d5cbdbd0af0237aac1b3030306a9f4

packages/base/Drawer/src/Drawer/Drawer.tsx
Kosov234: https://github.com/toptal/picasso/pull/5035/changes/94054cc4d3d5cbdbd0af0237aac1b3030306a9f4

nx.json
azebich: ah right, let's check with him if this is intended to be merged into master or just for testing 👍 
cc @vedrani

packages/base/Drawer/src/Drawer/test.tsx
vedrani: Hmm, does it make sense in that case to add some rule (not sure where right now, maybe base-ui) to define that every usage of base-ui portal with picasso container requries undefined or null.

Or even better did you consider to change usePicassoRoot and return `undefined` not `null`, it would we would'nt need to solve it per component and portal usage.

nx.json
vedrani: I still didn't have time to check, but I guess it's fine to merge it into picasso-modernization-temp for now and recheck as part of final testing. Let me merge it.

packages/base/Drawer/src/Drawer/test.tsx
Kosov234: Fair point, done in https://github.com/toptal/picasso/pull/5035/changes/27ca5b8a91545a4bd8c7be038600df44d6c56996

packages/base/Drawer/src/Drawer/Drawer.tsx
```diff
@@ -175,6 +178,9 @@ export const Drawer = ({
       }}
     >
       {disablePortal && <span ref={setInlineContainer} />}
+      {/* inlineContainer stays an explicit `null` until the inline mount node
+          renders — Base UI's portal waits for it (state-backed, arrives one
+          commit later) instead of falling back to `document.body` */}
```
vedrani: Remove comment?

vedrani (approved):

packages/picasso-provider/src/Picasso/RootContext.ts
```diff
@@ -31,10 +31,24 @@ export const RootContext = React.createContext<RootContextProps>({
...
-export const usePicassoRoot = () => {
-  const context = useContext(RootContext)
+// State-backed Picasso root node. Kept in a dedicated context (instead of a
+// field on RootContext) so that only `usePicassoRoot` consumers re-render
+// when the node becomes available after the first mount.
+export const PicassoRootNodeContext =
+  React.createContext<HTMLDivElement | null>(null)
 
-  return context && context.rootRef ? context.rootRef.current : null
+export const usePicassoRoot = (): HTMLDivElement | undefined => {
```
vedrani, line 40: Do you need to define return types or it will be implicitly defined?

`: HTMLDivElement | undefined `

azebich (approved):

packages/picasso-provider/src/Picasso/RootContext.ts
Kosov234: Just thought to keep it as a contract guard, so that if anyone would try to return `null` in the future - he will get a TS error

PR #5039, opened 2026-07-08
[PF-1994] Tweak cypress transitions and wait logic

azebich:
[PF-1994] Tweak cypress transitions and wait logic

### Description

Follow-up to #5028, which fixed the flaky Cypress-Happo captures on `PageTopBarMenu` by gating the `happoScreenshot` override on `@floating-ui` poppers settling. This PR extends the same doctrine across the rest of the Cypress component suite: every screenshot that captures async content now waits for that content to be **present and settled** before serializing, so `happo-cypress` can't freeze a mid-flight frame.

**Root cause (recap).** `happo-cypress` serializes the live DOM into a single static frame. Any content that mounts or animates asynchronously — a popper positioning over a couple of `@floating-ui` `autoUpdate` frames, a base64 image still decoding, a notistack notification sliding in, a recharts tooltip mounting on `mousemove` — can land mid-flight in that one instant and diff against the baseline. The global override only ever waited for `data-starting-style` (a Base UI signal) and, after #5028, for open poppers' geometry — neither of which covers image decode, notification slide-in, or the case where the popper **hasn't mounted yet** (an empty match set reports "settled" vacuously, so the capture races the async open).

**What this PR does:**

- **Generalizes the settle gate to notifications.** `waitForPoppersToSettle` → `waitForTransientGeometryToSettle`: the selector now also matches `[role="alert"]`, so notistack notifications (which slide in via inline-style transforms that `happo-cypress` captures verbatim) must reach a stable bounding box before serialization — same two-consecutive-identical-reads mechanism, still a no-op when nothing matches. Covers `NotificationStream`, `Form`, `QueryBuilder`, `Page`.
- **Adds `cy.waitForImagesDecoded(selector?)`.** A shared command that gates on every matched `` having decoded (`naturalWidth > 0`) rather than merely being laid out (`be.visible` passes on an `object-cover` image before its pixels paint). Replaces the same assertion hand-rolled across `Avatar`, `AvatarUpload`, `PageTopBarMenu`, and (scoped to the editor) `RichTextEditor/ImagePlugin`.
- **Adds `cy.waitForCalendarOpen()`.** Gates on the `DatePicker` calendar being mounted (`.rdp-month` visible) before its screenshots. The calendar opens via async state and its Popper renders a frame after mount, so the geometry-settle would otherwise pass vacuously and capture an empty page.
- **Adds popper/tooltip/alert presence gates in specs the settle gate can't cover.** `Select` (via its `openSelect()` helper — one place, all callers), `TagSelector` (open combobox + filtered "other option"), and `BarChart` (recharts tooltip) now assert their async content is visible **before** the screenshot instead of after. `PageTopBarMenu` gates on the opened dropdown's content before letting the avatars decode.

**No `cy.wait(ms)` / hardcoded timeouts** — every wait is a `.should()`-retried DOM condition, resolving the instant it's satisfied and only failing at the existing `defaultCommandTimeout` ceiling. **Test-infra only:** no component source is touched (`cypress/` exclusively), so no changeset is needed and no consumer-visible behavior changes.

### How to test

- [Temploy](https://picasso.toptal.net/tweak/cypress-tests-transitions)
- The change is capture-timing only; the intended effect is fewer non-deterministic 1-diff Happo flakes on `DatePicker`, `Select`, `TagSelector`, `BarChart`, `NotificationStream`, `Avatar`, `AvatarUpload`, `PageTopBarMenu`, and `ImagePlugin` snapshots. Re-run the Visual Tests job (or `pnpm test:integration`) and confirm those suites are stable across repeated runs.

### Screenshots

| Before                                  | After                                   |
| N/A — no visual change; this PR removes non-deterministic capture races, so baselines are unchanged | N/A |

### Development checks

- [x] Add changeset according to guidelines (if needed) — _not needed; `cypress/`-only, no package code changed_
- [x] Self reviewed
- [x] Covered with tests (visual tests included) — _this PR is test-infra hardening for the visual suite itself_

### DIFF
--- a/cypress/component/RichTextEditor/ImagePlugin.spec.tsx
+++ b/cypress/component/RichTextEditor/ImagePlugin.spec.tsx
@@ -107,6 +107,10 @@ describe('ImagePlugin', () => {
         `<p><img src="${uploadedFileContent}" alt="${uploadedFileAltText}"></p>`
       )
 
+      // the assertion above only checks the serialized HTML string — also let
+      // the <img> rendered in the editor decode before capturing
+      cy.waitForImagesDecoded(`${editorSelector} img`)
+
       cy.get('body').happoScreenshot({
         component,
         variant: 'image-plugin/successful-upload',
--- a/cypress/component/RichTextEditor/ImagePlugin.spec.tsx
+++ b/cypress/component/RichTextEditor/ImagePlugin.spec.tsx
@@ -1,41 +1,20 @@
-import React, { useState } from 'react'
-import type {
-  RichTextEditorProps,
-  UploadedImage,
-} from '@toptal/picasso-rich-text-editor'
-import { ImagePlugin, RichTextEditor } from '@toptal/picasso-rich-text-editor'
-import { Container } from '@toptal/picasso'
+import React from 'react'
+import type { UploadedImage } from '@toptal/picasso-rich-text-editor'
+import { ImagePlugin } from '@toptal/picasso-rich-text-editor'
 
-const editorTestId = 'editor'
+import {
+  Editor,
+  component,
+  editorSelector,
+  makeEditorProps,
+  resultContainerTestId,
+} from './test-helpers'
+
 const imageUploadButtonTestId = 'image-upload-button'
-const resultContainerTestId = 'result-container'
 
-const defaultProps = {
-  id: 'foo',
-  onChange: () => {},
-  placeholder: 'placeholder',
-  testIds: {
-    editor: editorTestId,
-    imageUploadButton: imageUploadButtonTestId,
-  },
-}
-
-const editorSelector = `#${defaultProps.id}`
-
-const Editor = (props: RichTextEditorProps) => {
-  const [value, setValue] = useState('')
-
-  return (
-    <Container style={{ maxWidth: '600px' }} padded='small'>
-      <RichTextEditor {...props} onChange={value => setValue(value)} />
-      <Container padded='small' data-testid={resultContainerTestId}>
-        {value}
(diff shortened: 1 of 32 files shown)


vedrani (approved):
Seems ok, but check how it works with https://github.com/toptal/picasso/pull/5035

azebich:
> Seems ok, but check how it works with #5035

I'll wait for it to be merged before doing final tweaks and making the PR rdy ;)

cypress/component/RichTextEditor/ImagePlugin.spec.tsx
```diff
@@ -151,11 +134,15 @@ describe('ImagePlugin', () => {
...
-      //   variant: 'image-plugin/failed-upload',
-      // })
+      // Screenshot was disabled pre-migration ("fonts stopped working in
+      // modals") and targeted MUI's role="presentation" portal wrapper; the
+      // @base-ui/react Modal emits role="dialog" instead (the passing
+      // successful-upload test above already queries it).
+      cy.waitForOverlayOpen()
+      cy.getByRole('dialog').happoScreenshot({
+        component,
+        variant: 'image-plugin/failed-upload',
```
vedrani, line 143: Are fonts showing now in deterministic way?

vedrani (approved):

cypress/component/RichTextEditor/ImagePlugin.spec.tsx
azebich: it should be good now, I also see the font has changed from previous happo to current happo

PR #5042, opened 2026-07-09
[PF-2224] Track tooltip anchor on scroll only to remove inside-dropdown jump

dulishkovych:
[PF-2224]

### Description

Fixes the inside-a-dropdown Tooltip position **"jump"**: when a tooltip opens inside a Dropdown, it paints slightly off and then snaps ~4px vertically **exactly 1 second after open**.

Root cause: the Dropdown's Popper positions one frame after mount, so the tooltip's base-ui `Positioner` first measures against a not-yet-positioned anchor. floating-ui's layout-shift `IntersectionObserver` sees `intersectionRatio: 0` and, to avoid an infinite loop, schedules the correction via `setTimeout(refresh, 1000)` — that 1s-late correction is the visible jump.

The fix sets **`disableAnchorTracking`** on the tooltip's `<BaseTooltip.Positioner>`. This turns off floating-ui's `ResizeObserver` + layout-shift `IntersectionObserver` (so the 1s throttle never arms) while keeping scroll tracking (`ancestorScroll`) on. The popup settles once and stays put.

**Not a regression from master.** The legacy popper.js v1 Tooltip repositioned only on scroll/window-resize — it had neither observer. This change removes exactly the two trackers base-ui added on top, restoring master's behavior. The only dropped behavior (re-measuring on an anchor **resize** or a **non-scroll layout shift** while open) is something master never did either; scroll-following is preserved.

Alternative considered: the more surgical cure is Dropdown-side — reveal the Paper only once positioned (the existing `[PF-1994]` TODO), which also removes the residual ~4px first-paint offset. This PR is the cheap, Tooltip-only, fully revertable mitigation that removes the *motion*; if the Dropdown-side fix lands later this prop becomes unnecessary.

Branches off `migrate-Tooltip-v2` (#5005) so the diff is a single prop + a changeset bullet.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2224-tooltip-disable-anchor-tracking)
- Open **Overlays / Tooltip → Inside of a Dropdown**, open the dropdown, and watch the tooltip on "Option 2": it should settle immediately with **no 1-second jump**. Compare against #5005's Temploy, where the same tooltip still snaps ~4px ~1s after opening.
- Regression check: hover tooltips, `followCursor`, and placements still position correctly, and a page scroll still repositions an open tooltip.

### Screenshots

**Before**

https://github.com/user-attachments/assets/5d5cba5b-0112-47cb-9912-9f0bf9dc88c8 

**After**

https://github.com/user-attachments/assets/7a1c87e5-82f1-4620-8373-d32e9a873678

### Development checks

- [x] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`) — N/A, no tailwind-merge change
- [x] Annotate all `props` in component with documentation — N/A, no new public props
- [x] Create `examples` for component — N/A, existing "Inside of a Dropdown" example covers it
- [x] Self reviewed
- [x] Covered with tests ([visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md) included) — 27/27 unit + Cypress inside-dropdown/Happo

**Breaking change** — N/A (folds into the unreleased Tooltip migration #5005; no independent consumer break)

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[PF-2224]: https://toptal-core.atlassian.net/browse/PF-2224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/base/Tooltip/src/Tooltip/Tooltip.tsx
```diff
@@ -213,6 +213,19 @@ export const Tooltip = forwardRef<HTMLElement, Props>(
...
+        // that IO's anti-infinite-loop throttle is what makes a tooltip opened
+        // inside a still-opening Dropdown snap ~4px exactly 1s after open (the
+        // anchor is measured before the Dropdown's Popper positions, the IO
+        // sees `intersectionRatio: 0`, and `setTimeout(refresh, 1000)` fires
+        // the late correction). Turning the extra trackers off makes the popup
+        // settle once and stay put. NOT a regression from master: the legacy
+        // popper.js Tooltip likewise repositioned only on scroll/window-resize
+        // (popper.js v1 has neither observer) — `ancestorScroll` stays on, so
+        // scroll-following is preserved; this restores that master behavior.
+        disableAnchorTracking
```
azebich: do we need to enable/disable it depending on the use case?

dulishkovych: i don't think so. As I found out this flag behaves in the same way as previous MUI implementation was and it fixes the behavior of the jumping tooltip within the dropdown.

packages/base/Tooltip/src/Tooltip/utils.ts
```diff
@@ -9,38 +9,250 @@ import type { OffsetType, PlacementType } from './Tooltip'
...
+): AnchorRect => {
+  if (!anchor) {
+    return makeRect({ left: 0, top: 0, width: 0, height: 0 })
+  }
+
+  const taintedAncestor = findScaleTaintedAncestor(anchor)
+
+  if (!taintedAncestor) {
+    return anchor.getBoundingClientRect()
+  }
```
vedrani, line 242: Also add check for this to ensure it can understand difference between mid-animation and persistent one.

```ts
// A static (non-animating) transform is deliberate geometry, not a
// mid-reveal taint — position against it like master did.
if (taintedAncestor.getAnimations().length === 0) {
  return anchor.getBoundingClientRect()
}
```

packages/base/Tooltip/src/Tooltip/utils.ts
```diff
@@ -9,38 +9,250 @@ import type { OffsetType, PlacementType } from './Tooltip'
...
+}
+
+// True while the element is scaled away from identity — via the standalone
+// `scale` property (what Tailwind v4's `scale-*` sets, and what the Dropdown's
+// Paper animates) or via a `transform` whose matrix has non-1 scale components.
+// A pure translate (`matrix(1, 0, 0, 1, x, y)`, e.g. popper.js positioning)
+// is NOT a taint: it moves boxes without distorting them, and
+// getBoundingClientRect reports it faithfully.
+const hasNonIdentityScale = (element: Element): boolean => {
+  const computed = getComputedStyle(element)
```
vedrani, line 62: Be careful about this, it's calculated on every ancestor on every scroll tick. Can we do some early exit or add boundary so it doesn't go each time to html element?

We can potentially can pass boundary from Tooltip into useTooltipAnchor.

Just idea, not sure will it be big improvement.

packages/base/Popper/src/Popper/Popper.tsx
```diff
@@ -175,14 +175,18 @@ export const Popper = forwardRef<PopperHandle, Props>(function Popper(
...
+  // it disables the resize/layout-shift observers for menu-item anchors — see
+  // its Positioner), but — because it is portaled to the picasso root, not into
+  // this popper's scroll-ancestor chain — a synthetic window scroll is the ONLY
+  // signal that reaches it when we move (most load-bearing for the menu-item
+  // case, whose observers are off). So this is not a transitional shim: it
+  // serves both the legacy popper.js Tooltip AND the migrated one
+  // (container-scroll following). [PF-2203][PF-1994][PF-2224]
   useEffect(() => {
     if (open && isPositioned) {
       window.dispatchEvent(new Event('scroll'))
```
vedrani, line 192: I'm not happy we are using this dispatch because it triggers lot of stuff across global scope of app. 

Can we do some spike after we finish with Tooltip bug fixes and investigate can it be refactored without it?

vedrani (approved):

azebich (approved):

packages/base/Tooltip/src/Tooltip/utils.ts
dulishkovych: Addressed and added a document-level fast path (if nothing is animating, return the live rect) so the common steady-state scroll re-solve skips the ancestor walk entirely rather than climbing to `<html>`.

packages/base/Popper/src/Popper/Popper.tsx
dulishkovych: I created this https://toptal-core.atlassian.net/browse/PF-2252

PR #5044, opened 2026-07-09
Fix twMerge className prop position

azebich:
### Description

Consumer `className` is passed to `twMerge(...)` **before** the component's own
base/variant classes in a number of migrated components. `twMerge` resolves
conflicting Tailwind utilities left-to-right with the **rightmost** class
winning, so a consumer `className` placed first is silently overridden by the
component's built-in classes whenever both target the same property. This breaks
the styling contract — the consumer `className` must win on conflicts.

This PR moves `className` to the last `twMerge(...)` argument so consumer
overrides take precedence again, in:

- **@toptal/picasso-avatar** — `IconAvatar`
- **@toptal/picasso-drawer** — `DrawerPaper`
- **@toptal/picasso-modal** — `Modal`, `ModalPaper`
- **@toptal/picasso-page** — `Page`
- **@toptal/picasso-select** — `NativeSelect`, `NonNativeSelect`

On the touched lines, a few Tailwind arbitrary-value classes are also normalized
to their Tailwind v4 bare-value / CSS-variable-shorthand form. These compile to
byte-identical CSS against the repo's Tailwind config, so there is no visual
change:

- `MenuItem`: `sm:min-h-[auto]` → `sm:min-h-auto`
- `PageSidebar`: `delay-[225ms]` → `delay-225`
- `PageTopBar`: `z-[1100]` → `z-1100`,
  `h-[var(--header-height,3.5rem)]` → `h-(--header-height,3.5rem)`,
  `min-h-[var(--header-height,3.5rem)]` → `min-h-(--header-height,3.5rem)`

The public Props API is unchanged; the consumer-visible impact is a bug fix
(overrides that were being dropped now apply), so all affected packages bump
`patch`.

### How to test

- Render any affected component with a `className` that conflicts with a
  built-in utility (e.g. `<Tag className="bg-blue-500" />`, or a `<Page>` /
  `<Modal>` with an overriding background/spacing class) and confirm the consumer
  class now wins instead of being dropped.

### Development checks

- [x] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`) — not touched
- [x] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core — no design change
- [x] Create `examples` for component — no new API
- [x] Ensure the changed/created components have not caused accessibility issues — no DOM/behavior change
- [ ] Self reviewed
- [ ] Covered with tests ([visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md) included)

### DIFF
--- a/packages/base/Menu/src/MenuItem/MenuItem.tsx
+++ b/packages/base/Menu/src/MenuItem/MenuItem.tsx
@@ -173,13 +173,13 @@ export const MenuItem: OverridableComponent<Props> = forwardRef<
           getFontColor({ variant, highlighted }),
           getBackgroundAndActionStyles({ variant, highlighted, nonSelectable }),
           disableGutters ? 'p-0' : 'px-4 py-[0.375rem]',
-          'min-w-[9rem] w-auto min-h-[unset] sm:min-h-[auto] md:min-h-0 relative cursor-pointer',
+          'min-w-[9rem] w-auto min-h-[unset] sm:min-h-auto md:min-h-0 relative cursor-pointer',
           'transition-colors duration-150 ease-in-out',
           'overflow-hidden whitespace-normal text-left no-underline',
           'flex items-center justify-start',
           'outline-hidden appearance-none leading-4',
-          className,
-          disabled && 'text-gray-600 opacity-100 pointer-events-none'
+          disabled && 'text-gray-600 opacity-100 pointer-events-none',
+          className
         )}
         aria-disabled={Boolean(disabled)}
         disabled={disabled}
--- a/packages/base/Menu/src/MenuItem/MenuItem.tsx
+++ b/packages/base/Menu/src/MenuItem/MenuItem.tsx
@@ -173,7 +173,7 @@ export const MenuItem: OverridableComponent<Props> = forwardRef<
           getFontColor({ variant, highlighted }),
           getBackgroundAndActionStyles({ variant, highlighted, nonSelectable }),
           disableGutters ? 'p-0' : 'px-4 py-[0.375rem]',
-          'min-w-[9rem] w-auto min-h-[unset] sm:min-h-auto md:min-h-0 relative cursor-pointer',
+          'min-w-[9rem] w-auto min-h-[unset] sm:min-h-[auto] md:min-h-0 relative cursor-pointer',
           'transition-colors duration-150 ease-in-out',
           'overflow-hidden whitespace-normal text-left no-underline',
           'flex items-center justify-start',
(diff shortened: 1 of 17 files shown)


azebich:
Previously the component used `cx` which didn't filter out duplicate classes so they needed to be applied before

packages/base/Menu/src/MenuItem/MenuItem.tsx
```diff
@@ -173,7 +173,7 @@ export const MenuItem: OverridableComponent<Props> = forwardRef<
           getFontColor({ variant, highlighted }),
           getBackgroundAndActionStyles({ variant, highlighted, nonSelectable }),
           disableGutters ? 'p-0' : 'px-4 py-[0.375rem]',
-          'min-w-[9rem] w-auto min-h-[unset] sm:min-h-[auto] md:min-h-0 relative cursor-pointer',
+          'min-w-[9rem] w-auto min-h-[unset] sm:min-h-auto md:min-h-0 relative cursor-pointer',
```
vedrani: Do you have concerns it uses tailwind-merge v2 instead of new v3?

It seems v2 and v3 have different behaviour and I'm not actually sure is v2 compatible with tailwind v4.

Also will tailwind-merge correctly handle it without brackets. It seems potentially it will not handle it properly and resolve it. Pls check in browser.

packages/base/Page/src/PageTopBar/PageTopBar.tsx
```diff
@@ -133,12 +133,12 @@ export const PageTopBar = forwardRef<HTMLElement, Props>(function PageTopBar(
 
   return (
     <PageTopBarContext.Provider value={{ variant }}>
-      <div className='relative h-[var(--header-height,3.5rem)] min-h-[var(--header-height,3.5rem)]'>
+      <div className='relative h-(--header-height,3.5rem) min-h-(--header-height,3.5rem)'>
```
vedrani: Seems again tw-merge v2 can have problems with bare formats, in this case it's fine because we are not using it, but still not good if in future we add it here.

vedrani (commented):
PR looks fine, only concerns is about stale tailwind-merge and how it manages bare values. Pls check to revert and/or create followup ticket for tailwind-merge upgrade.

packages/base/Menu/src/MenuItem/MenuItem.tsx
azebich: awesome catch!

vedrani (approved):
Please create followup upgrade ticket for tw-merge to v3.

PR #5002, opened 2026-06-10
[PF-1994] Migrate Accordion to @base-ui/react + Tailwind

azebich:
# [PF-1994] Migrate Accordion to @base-ui/react + Tailwind

## Summary

Migrates `@toptal/picasso-accordion` (Tier 3.a) off MUI v4 (`@material-ui/core` Accordion + JSS) onto `@base-ui/react/accordion` parts (`Root` + `Item` + `Header` + `Trigger` + `Panel`) with all styling rewritten as Tailwind class arrays in `styles.ts`. The single-item Picasso API (`expanded` boolean, `content` prop, `borders` variant) is preserved by wrapping Base UI's multi-item value model behind a fixed one-item value. The public `classes` prop is dropped on all three subcomponents per the Tier 3.a decision, and the one internal consumer (`Page/SidebarItemAccordion`) is migrated to `className`-based slot targeting in the same PR.

## Decisions

- **`classes` dropped via `Omit<StandardProps, 'classes'>` + runtime backstop** on Accordion, AccordionSummary, AccordionDetails: audit-verified 0 external/0 internal public callsites (`decisions/classes-audit.md` §5, Tier 3.a). The only internal caller, `SidebarItemAccordion`, now routes its `summary`/`content` slot styling through the root `className` with `[&_[data-component-type=…]]:` selectors until Page's own migration (per-item plan §5, option "single PR").
- **`major` bump** taken verbatim from `manifest.json#versionBump`: removing `classes` and ignoring `transitionProps.timeout` are concrete consumer-visible breaks; both are named in the changeset. The Page changeset is `patch` (internal styling route only, no visual change).
- **Expand/collapse animation is CSS-driven** (`transition-[height]` on `Panel` with `--accordion-panel-height` + `data-[starting-style]`/`data-[ending-style]`), so `transitionProps.timeout` is ignored; `transitionProps.onExited` parity is kept by firing it from `onTransitionEnd` on collapse.
- **Border-collapse between stacked accordions** uses a CSS sibling selector (`[[data-component-type="accordion"]+&]:before:hidden`) instead of the legacy JSS `&$expanded`/sibling rules — verified at runtime for both `all` and `middle` border variants.
- **`onChange` event bridge**: Base UI's `onValueChange` eventDetails bridge to Picasso's `(event: ChangeEvent, expanded)` signature via `toReactEvent` from `@toptal/picasso-shared` (canonical boundary pattern).

## Limitations / Out-of-scope

[...]

PR #5045, opened 2026-07-13
[PF-2234] Tweak Accordion

azebich:
[PF-2240]

### Description

Post-migration fixes for the `@base-ui/react` Accordion (#5002). Three
consumer-visible behaviors regressed or were missing versus the pre-migration
MUI version; this PR restores them. Public Props API is unchanged → **patch**.

1. **Keyboard-focus background restored.** MUI's `AccordionSummary` (a
   `ButtonBase`) painted `palette.action.focus` (`rgba(0,0,0,0.12)`) on keyboard
   focus. The migrated summary is a plain `<div>` with `outline-none` and nothing
   replacing it, so focusing a summary showed **no indicator at all** (WCAG 2.4.7
   regression). Re-added as `focus-visible:bg-black/12` — matches the old value,
   and `focus-visible` keeps it keyboard-only like MUI's `focusVisible`.

2. **Panel overflow clip released once fully open.** MUI `Collapse` switched to
   `overflow: visible` after the open transition finished (its `entered` state).
   The migrated Panel kept `overflow-hidden` permanently, so content that
   legitimately overflows the open panel — dropdowns, tooltips, poppers, focus
   rings on the last row — was **clipped**. Restored by releasing the clip only
   after the open height-transition settles. An `isAnimating` flag (derived with
   `summaryExpanded`) keeps the clip active for the whole open/close animation, so
   content no longer "pops" out before the box finishes growing.

3. **Expand icon removed from the tab order** (`tabIndex={-1}`). The icon
   `ButtonAction` was a redundant, unlabeled tab stop; the summary is the sole
   interactive control. Also added `data-component-type='accordion-summary-icon'`
   to the icon wrappers, matching the component's existing addressing scheme
   (`accordion`, `accordion-summary`, `accordion-summary-content`).

Files: `Accordion.tsx` (`isAnimating` state + `transitionend` rewiring,
`tabIndex`, `data-component-type`), `Accordion/styles.ts` (new
`panelOverflowVisibleClass`), `AccordionSummary/styles.ts` (focus background).

### How to test

- [Temploy](https://picasso.toptal.net/tweak/pf-2234-accordion)
- Keyboard-focus an accordion summary (Tab) → visible focus background appears;
  Tab again → focus skips the arrow and moves past the accordion.
- Open an accordion whose content overflows near the bottom edge (e.g. a
  `Dropdown` or `Tooltip`) → the popup is **not clipped** when open, and there is
  **no content "pop"** during the open animation.
- Collapse the accordion → content clips correctly while animating; if
  `transitionProps.onExited` is provided, it still fires after collapse.
- Stacked accordions and `borders` variants render unchanged.

### Screenshots

| No focus ring on keyboard focus; open-panel dropdown/tooltip clipped | Focus background on keyboard focus; open-panel content not clipped |

### Development checks

- [x] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`) — no new API used, no bump needed
- [x] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core — restores pre-migration behavior; confirm focus-token choice with design
- [ ] Create `examples` for component — N/A, no new API
- [x] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md). — this PR fixes two a11y issues (focus visibility, tab order)
- [x] Self reviewed
- [ ] Covered with tests ([visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md) included) — snapshots regenerated; Happo reviewed on CI

**Breaking change**

[PF-2240]: https://toptal-core.atlassian.net/browse/PF-2240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/base/Accordion/src/Accordion/Accordion.tsx
+++ b/packages/base/Accordion/src/Accordion/Accordion.tsx
@@ -24,6 +24,7 @@ import {
   createExpandIconClassNames,
   expandIconAlignTopClasses,
   panelClasses,
+  panelOverflowVisibleClass,
 } from './styles'
 
 export type { Borders } from './styles'
@@ -117,11 +118,14 @@ export const Accordion = forwardRef<HTMLElement, Props>(function Accordion(
 
   const [summaryExpanded, setSummaryExpanded] = useState(defaultExpanded)
   const [prevExpanded, setPrevExpanded] = useState(defaultExpanded)
+  // Releases the panel's overflow clip only after the open transition settles.
+  const [isAnimating, setIsAnimating] = useState(false)
 
   // getDerivedStateFromProps implementation to allow expanded to be controlled
   if (expanded !== undefined && expanded !== prevExpanded) {
     setSummaryExpanded(expanded)
     setPrevExpanded(expanded)
+    setIsAnimating(true)
   }
 
   const handleValueChange = (
@@ -131,6 +135,7 @@ export const Accordion = forwardRef<HTMLElement, Props>(function Accordion(
     const newExpanded = value.length > 0
 
     setSummaryExpanded(newExpanded)
+    setIsAnimating(true)
     onChange(
       toReactEvent<ChangeEvent<Element>>(eventDetails.event),
       newExpanded
@@ -139,10 +144,15 @@ export const Accordion = forwardRef<HTMLElement, Props>(function Accordion(
 
   const handlePanelTransitionEnd = (event: TransitionEvent<HTMLDivElement>) => {
     if (
-      !summaryExpanded &&
-      event.target === event.currentTarget &&
-      event.propertyName === 'height'
+      event.target !== event.currentTarget ||
+      event.propertyName !== 'height'
     ) {
+      return
+    }
+
+    setIsAnimating(false)
+
+    if (!summaryExpanded) {
       transitionProps?.onExited?.(event.currentTarget)
     }
   }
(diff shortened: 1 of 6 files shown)


packages/base/Accordion/src/Accordion/Accordion.tsx
```diff
@@ -117,11 +118,14 @@ export const Accordion = forwardRef<HTMLElement, Props>(function Accordion(
 
   const [summaryExpanded, setSummaryExpanded] = useState(defaultExpanded)
   const [prevExpanded, setPrevExpanded] = useState(defaultExpanded)
+  // Releases the panel's overflow clip only after the open transition settles.
+  const [isAnimating, setIsAnimating] = useState(false)
```
vedrani: What if in controlled usage it's gets mounted with `expanded = true`? I think it will be stuck with overflow-hidden. Pls check.

Use `useState(expanded ?? defaultExpanded)` for summaryExpanded, prevExpanded.

Pls check.

packages/base/Accordion/src/Accordion/Accordion.tsx
```diff
@@ -189,15 +199,21 @@ export const Accordion = forwardRef<HTMLElement, Props>(function Accordion(
               {children}
               {expandIcon ? (
                 <ButtonAction
+                  data-component-type='accordion-summary-icon'
```
vedrani: Seems ButtonBase is overriding it, pls check.

packages/base/Accordion/src/Accordion/Accordion.tsx
vedrani, line 79: Do we need to animate if 

```
transitionProps={{ timeout: 0 }}
```

and will it trigger `handlePanelTransitionEnd`

packages/base/Accordion/src/Accordion/Accordion.tsx
azebich: If it get's expanded by default it will set the overflow correctly because we have `[panelOverflowVisibleClass]: summaryExpanded && !isAnimating`. The `isAnimating` is only used so that we don't set the overflow during transition because it will show the full content while the accordion is expanding

packages/base/Accordion/src/Accordion/Accordion.tsx
azebich: tested and it should work fine with latest changes

packages/base/Accordion/src/Accordion/Accordion.tsx
azebich: fixed the problem by adding an div wrapper similar to what we have when we don't use `expandIcon`

vedrani (approved):

PR #5048, opened 2026-07-14
[PF-2230] Fix Popper default role

azebich:
[PF-2230]

### Description

`Popper` is Picasso's generic positioning primitive, but it hardcoded `role="tooltip"` on its floating element. Since the `@base-ui/react` migration, the real Tooltip renders through Base UI's `Tooltip.Popup` (which sets its own `role="tooltip"`), so **no consumer actually wants the tooltip role from `Popper`** — yet all five of its consumers inherited it. A calendar, a menu, and two listboxes were each announced to assistive tech as a `tooltip` (a passive description region), misrepresenting interactive content.

This PR keeps `role="tooltip"` as `Popper`'s default (so any un-migrated caller stays correct) but makes it an explicit, typed, overridable prop, and sets the correct role at each consumer site.

**`@toptal/picasso-popper`**
- Add a typed `role?: React.AriaRole` prop, defaulting to `'tooltip'`. Previously `role` was not part of `Popper`'s `Props` (it extends `BaseProps`, not `HTMLAttributes`), so consumers could not pass one — it now applies on the floating element via `role={role}`.

**Consumers — correct role per popup content:**
| Component | Role | Rationale |
| --- | --- | --- |
| `DatePicker` | `dialog` | The calendar popup is a dialog; nothing inside claims the role. |
| `Menu` (submenu via `MenuItem`) | `presentation` | The inner `Menu` already owns `role="menu"`. |
| `Dropdown` | `presentation` | Content (usually a `Menu`) supplies its own semantics. |
| `Select` (`NonNativeSelect`) | `presentation` | The options list already owns `role="listbox"`. |
| `Autocomplete` | `presentation` | The popup already sits inside a `role="listbox"` wrapper. |

The four wrapper cases use `role="presentation"` rather than a content-specific role on purpose: putting `listbox`/`menu` on the `Popper` wrapper would nest a second, competing role and create a new a11y bug, since those roles already exist on the inner elements.

**Tests**
- Add a `role` block to `Popper`'s unit tests: default `tooltip`, custom role override, and `presentation` removing the element from the accessibility tree.
- Update `Dropdown` unit test + snapshot to the new `presentation` role.
- Fix `Select.spec.tsx` — it located the open dropdown via `[role="tooltip"]`, which no longer exists; switched to the `searchInput` testId (matching the sibling test).

Motivation: found during an ARIA-role audit of `packages/base`. This is a pre-existing issue (the `tooltip` default predates the migration), fixed here because it's one primitive feeding five consumers — high leverage, low blast radius. No visual change — ARIA roles are invisible to Happo.

### How to test

- Open a `DatePicker` and inspect the calendar popup — it should expose `role="dialog"`, not `role="tooltip"`.
- Open a `Dropdown`, a nested `Menu` submenu, a `Select`, and an `Autocomplete`; inspect the floating container — it should be `role="presentation"`, with the inner `menu`/`listbox` roles intact and singular (no nested duplicate `listbox`).
- Confirm a plain `Tooltip` still exposes `role="tooltip"` (unchanged default).
- Run the Storybook accessibility addon on the affected components — no new violations.

### Screenshots

_No visual change — ARIA roles only._

| Before | After |
| --- | --- |
| `Popper` floating element: `role="tooltip"` for all consumers | `dialog` (DatePicker) / `presentation` (Dropdown, Menu submenu, Select, Autocomplete); Tooltip unchanged |

### Development checks

- [ ] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`)
- [x] Self reviewed

**Breaking change**

[PF-2230]: https://toptal-core.atlassian.net/browse/PF-2230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/popper-role-prop.md
+++ b/.changeset/popper-role-prop.md
@@ -1,0 +1,20 @@
+---
+'@toptal/picasso-popper': minor
+'@toptal/picasso-select': patch
+'@toptal/picasso-autocomplete': patch
+'@toptal/picasso-dropdown': patch
+'@toptal/picasso-menu': patch
+'@toptal/picasso-date-picker': patch
+---
+
+### Popper
+
+- add a `role` prop for the floating element (defaults to `tooltip`)
+
+### DatePicker
+
+- expose the calendar popup as `role="dialog"` instead of Popper's default `tooltip`
+
+### Select / Autocomplete / Dropdown / Menu
+
+- mark the popup positioning wrapper as `role="presentation"` — the inner listbox/menu owns the popup semantics, so assistive tech no longer announces these popups as tooltips
--- a/.changeset/popup-focus-fixes.md
+++ b/.changeset/popup-focus-fixes.md
@@ -1,0 +1,13 @@
+---
+'@toptal/picasso-modal': patch
+'@toptal/picasso-date-picker': patch
+---
+
+### Modal
+
+- exempt Popper-based popups (Select, DatePicker, Dropdown, Menu) from the focus trap, so focus is no longer stolen from a popup opened inside a Modal (e.g. a Select's search input)
+
+### DatePicker
+
+- fix Tab dead-ending on the input when entering an open calendar a second time
+- fix Shift+Tab on the input jumping forward into the calendar instead of moving back to the previous field
--- a/cypress/component/DatePicker.spec.tsx
+++ b/cypress/component/DatePicker.spec.tsx
@@ -225,6 +225,14 @@ describe('DatePicker', () => {
     cy.mount(<DrawerExample />)
 
     cy.getByTestId(DATEPICKER_INPUT_TESTID).as('input').click()
+
+    // the visually risky state — the calendar popper layering above the drawer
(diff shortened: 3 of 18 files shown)


dulishkovych (approved):

vedrani (commented):
Check how Modal works with `focus`, I think it needs also changing.

https://github.com/toptal/picasso/blob/feature/picasso-modernization-temp/docs/migration/reference/Modal.tsx#L75

https://github.com/toptal/picasso/blob/feature/picasso-modernization-temp/docs/migration/reference/Modal.tsx#L98

Suggestion: decouple Modal's check from ARIA semantics — have Popper always render a stable marker attribute on its floating div (e.g. data-picasso-popper='') and widen Modal's selector to '[role=tooltip], [data-picasso-popper]'

vedrani (approved):

PR #5049, opened 2026-07-14
[PF-2243] Fix Modal duplicate role attribute

azebich:
[PF-2243]

### Description

Modal rendered **two nested `role="dialog"`** elements: Base UI's `Dialog.Popup` supplies one by default, and `ModalPaper` also set its own — a regression introduced by the `@base-ui/react` migration. Previously `@mui/base`'s Modal root had no role, so the paper legitimately carried it; after the swap, Base UI's popup provides the role natively, making the paper's copy a duplicate.

Nested duplicate dialog roles are an accessibility anti-pattern (a screen reader sees two dialogs) and break test/automation that assumes a single match — `getByRole('dialog')` throws on multiple matches, and `[role="dialog"]` selectors (e.g. the shared `waitForOverlayOpen` Cypress command) become ambiguous.

**Changes (`@toptal/picasso-modal`)**
- Remove `role="dialog"` from `ModalPaper`. The role now lives only on Base UI's `Dialog.Popup` — the element that actually hosts the focus trap and aria wiring, so semantics and behavior stay on the same node.
- Route a consumer's `paperProps.role` override onto the `Dialog.Popup` instead of the inner Paper, so "changing role" still works and still targets the true dialog element. The role key is applied conditionally (`paperProps.role !== undefined`) — passing it through unconditionally would send `role={undefined}` and clobber Base UI's default, because Base UI's `mergeProps` copies `undefined` keys.
- Alias the import to `BaseUIDialog` for readability against the local `Modal`/`ModalContext` names.

**Snapshots**
- Updated `Modal`, `PromptModal`, and `Utils/Modal` jest snapshots to drop the removed duplicate `role="dialog"` from the paper node. Each now shows a single `role="dialog"` (on the popup).

Found during an ARIA-role audit of `packages/base`. No visual change — ARIA roles are invisible to Happo.

### How to test

- Open any Modal (and PromptModal) and inspect the DOM — there should be exactly **one** `role="dialog"` (on Base UI's popup), not two nested.
- Pass `paperProps={{ role: 'alertdialog' }}` — the override should land on the dialog popup, and there should still be only one dialog role in the tree.
- Confirm existing Modal/Drawer/PromptModal Cypress specs still pass (`waitForOverlayOpen` resolves a single `[role="dialog"]`).
- Run the Storybook accessibility addon on Modal — no new violations.

### Screenshots

_No visual change — ARIA roles only._

| Before | After |
| --- | --- |
| Two nested `role="dialog"` (Base UI popup **and** inner Paper) | Single `role="dialog"` on the Base UI popup; `paperProps.role` override applied there |

### Development checks

- [ ] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`)
- [x] Self reviewed

**Breaking change**

[PF-2243]: https://toptal-core.atlassian.net/browse/PF-2243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/base/Modal/src/Modal/Modal.tsx
+++ b/packages/base/Modal/src/Modal/Modal.tsx
@@ -213,12 +213,6 @@ export const Modal = forwardRef<HTMLDivElement, Props>(function Modal(
     (typeof container === 'function' ? container() : container) ||
     picassoRootContainer
 
-  // Base UI's Dialog.Popup already renders `role="dialog"`. Honor a consumer's
-  // `paperProps.role` override on the popup (the element Base UI treats as the
-  // dialog) rather than the inner Paper.
-  const { role: paperRole, ...restPaperProps } = paperProps ?? {}
-  const popupRoleProps = paperRole !== undefined ? { role: paperRole } : {}
-
   const handlePopupClick = (event: React.MouseEvent<HTMLDivElement>) => {
     if (event.target === event.currentTarget && !disableBackdropClick) {
       onBackdropClick?.()
@@ -253,7 +247,6 @@ export const Modal = forwardRef<HTMLDivElement, Props>(function Modal(
         )}
         <BaseUIDialog.Popup
           {...rest}
-          {...popupRoleProps}
           ref={modalRef}
           initialFocus={() =>
             modalRef.current?.contains(document.activeElement)
@@ -267,12 +260,7 @@ export const Modal = forwardRef<HTMLDivElement, Props>(function Modal(
           style={{ ...style, ...durationStyle }}
           onClick={handlePopupClick}
         >
-          <ModalPaper
-            size={size}
-            align={align}
-            tabIndex={-1}
-            {...restPaperProps}
-          >
+          <ModalPaper size={size} align={align} tabIndex={-1} {...paperProps}>
             <ModalContext.Provider value>
               {children}
               {onClose && (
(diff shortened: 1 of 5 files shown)


packages/base/Modal/src/Modal/Modal.tsx
```diff
@@ -213,6 +213,12 @@ export const Modal = forwardRef<HTMLDivElement, Props>(function Modal(
     (typeof container === 'function' ? container() : container) ||
     picassoRootContainer
 
+  // Base UI's Dialog.Popup already renders `role="dialog"`. Honor a consumer's
+  // `paperProps.role` override on the popup (the element Base UI treats as the
+  // dialog) rather than the inner Paper.
+  const { role: paperRole, ...restPaperProps } = paperProps ?? {}
```
vedrani: This are paperProps not popupProps, I think we shouldn't route it like this but just spread into paper props.

Please change.

vedrani (approved):

packages/base/Modal/src/Modal/Modal.tsx
azebich: ah makes sense, we still have the popup role so we don't need to overwrite the paper role into popup :D

dulishkovych (approved):

PR #5057, opened 2026-07-20
[PF-2253] Refactor Tooltip open-arbitration and fix pointer-focus open

dulishkovych:
[PF-2253](https://toptal-core.atlassian.net/browse/PF-2253)

### Description

Fixes PF-2253 (the migrated Tooltip opening on **pointer-initiated focus**) and, in
the same pass, de-accretes the open-arbitration logic so the fix lands on a clean
foundation instead of as another inline special-case.

**The bug.** A `mousedown` that focuses a tooltip trigger made base-ui open the popup
mid-click (its `useFocus` reports `trigger-focus`, and `matchesFocusVisible` treats
that focus as focus-visible under jsdom and for typeable triggers such as inputs in a
real browser). The popup could flash open over its own trigger and swallow the
trailing click, leaving the wrapped control effectively unclickable.

**The refactor.** `handleOpenChange` had become a referee vetoing base-ui's open
requests one reason at a time — an accreting pile of inline `if (...) return` guards.
This PR replaces that with:

- **`isPointerModality` shared util** (`@toptal/picasso-utils`) — a ref-counted,
  SSR-guarded, window-capture input-modality tracker. It is the single home for the
  jsdom/Cypress/`matchesFocusVisible` rationale, and the reusable primitive other
  base-ui-migrated components can adopt when they hit the same `:focus-visible`
  unreliability. Tooltip is its first consumer.
- **`shouldHonorOpen` pure arbiter** — the veto pile collapsed into one pure,
  exhaustively-tested truth table.

**The fix, expressed as one arbiter row:** pointer-initiated `trigger-focus` opens are
vetoed; keyboard focus still opens (accessibility preserved).

**Controlled-mode note (please read).** The pointer-focus veto applies in **both**
controlled and uncontrolled mode, so a controlled tooltip no longer opens on
pointer-initiated focus either. Every **other** arbitration rule (hover-timing,
click-dismiss latch, follow-cursor roam-hide) keeps its original **uncontrolled-only**
gating — controlled consumers such as `TypographyOverflow` rely on base-ui's forwarded
hover-open (`onOpen`) because Picasso's own hover timer is disabled in controlled mode.
This is the sole controlled-mode behavior change. See
`docs/decisions/20-tooltip-open-arbitration.md`.

`pnpm build:package` (90 projects) and lint pass; 47 unit tests across the arbiter,
the util, and the Tooltip integration suite are green.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2253-tooltip-pointer-focus-open)

**Manual steps** (Storybook → `Tooltip → Default` and `Tooltip → Interactive`):

1. **Click the trigger.** The tooltip must **not** flash open mid-click. *(This is the
   fix.)* The click already focuses the trigger — so don't press Tab expecting it to
   re-focus the same trigger; Tab moves focus **forward** to the next element, which is
   normal browser behavior.
2. **Hover the trigger.** Still opens after the delay — unchanged.
3. **Keyboard focus opens it:** click an empty area to blur, then **Tab** (or
   **Shift+Tab**) until the focus ring lands **on the trigger** → the tooltip appears.
   Blur (Tab away) closes it. Accessibility preserved.
4. **Modality resets:** click the trigger (no tooltip) → blur → Tab back onto it →
   tooltip opens. A prior pointer click must not poison later keyboard focus.
5. **Tooltip-wrapped `<input>`** (the real-Chrome case — buttons don't get
   `:focus-visible` on mouse focus, so a plain button won't visibly differ in Chrome):
   clicking into the input must **not** open the tooltip; **Tab**-focusing it **must**.
6. **Controlled consumer:** hover an ellipsed **`TypographyOverflow`** — the tooltip must
   still open on hover (guards the controlled-mode path, which relies on base-ui's
   forwarded `onOpen`).

**Consumer validation:** client-portal `TalentAvailabilityNotificationBadge.cy.tsx` must
pass **after reverting** its `click({ force: true })` PF-2253 workarounds.

### Screenshots

| Before                                  | After                                   |
| Tooltip flashes open on click; trailing click can be lost | Click does not open the tooltip; wrapped control stays clickable |

### Development checks

- [x] Add changeset according to [guidelines](https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md) (if needed) — `@toptal/picasso-tooltip` patch, `@toptal/picasso-utils` minor (new public export)
- [ ] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`) — N/A
- [ ] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/) — N/A, behavior fix only
- [ ] Annotate all `props` in component with documentation — N/A, no API change
- [ ] Create `examples` for component — N/A, no new component
- [x] Ensure the changed/created components have not caused accessibility issues — keyboard focus still opens; only pointer-initiated focus is suppressed
- [x] Self reviewed
- [x] Covered with tests — pure arbiter truth-table tests + modality-util tests + Tooltip integration tests (incl. controlled-mode honor and veto paths); rationale recorded in ADR-20

**Breaking change** — N/A (patch/minor). Note: stricter than the pre-migration build
for tooltip-wrapped **inputs**, which previously also flashed open on click; and a
latched pointer modality suppresses focus-opens for programmatic `.focus()` after a
pointer interaction until the next keydown (disclosed in ADR-20 drawbacks).

[PF-2253]: https://toptal-core.atlassian.net/browse/PF-2253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/tooltip-suppress-pointer-focus-open.md
+++ b/.changeset/tooltip-suppress-pointer-focus-open.md
@@ -1,0 +1,24 @@
+---
+'@toptal/picasso-tooltip': patch
+---
+
+### Tooltip
+
+- Suppress opening on pointer-initiated focus. A `mousedown` that focuses a
+  tooltip trigger made base-ui open the tooltip mid-click (its `useFocus`
+  reports `trigger-focus`, and `matchesFocusVisible` treats that focus as
+  focus-visible under jsdom and for typeable triggers such as inputs). The
+  popup could then flash open over its own trigger and swallow the trailing
+  click, leaving the wrapped control effectively unclickable. Picasso now
+  tracks the input modality and honors base-ui's focus-open only when the focus
+  was keyboard-driven; pointer-initiated focus no longer opens the tooltip.
+  Keyboard focus still opens it, preserving accessibility. Note this is
+  stricter than the pre-migration build for tooltip-wrapped inputs, which
+  previously also flashed open on click. [PF-2253]
+- The pointer-focus suppression now also applies to controlled tooltips: a
+  pointer-initiated focus no longer calls `onOpen` on a controlled tooltip
+  either. This is the only controlled-mode behavior change — hover-timing, the
+  click-dismiss latch and follow-cursor roam-hide keep their original
+  uncontrolled-only gating, so a controlled tooltip still forwards base-ui's
+  hover-open to `onOpen` as before. See
+  `docs/decisions/20-tooltip-open-arbitration.md`. [PF-2253]
--- a/.changeset/utils-pointer-modality.md
+++ b/.changeset/utils-pointer-modality.md
@@ -1,0 +1,12 @@
+---
+'@toptal/picasso-utils': minor
+---
+
+### Utils
+
+- Add `isPointerModality`, `subscribePointerModality` and
+  `unsubscribePointerModality` — a shared, ref-counted input-modality tracker
+  (window-capture `keydown`/`pointerdown` listeners) for telling
+  pointer-initiated focus from keyboard-initiated focus in environments where
+  `:focus-visible` cannot. First consumer: Tooltip's pointer-focus open
+  suppression. [PF-2253]
--- a/docs/decisions/20-tooltip-open-arbitration.md
+++ b/docs/decisions/20-tooltip-open-arbitration.md
@@ -1,0 +1,141 @@
+# Tooltip open arbitration
+
+## Problem
+
+The migrated (base-ui) Tooltip runs two open-timing models at once. base-ui's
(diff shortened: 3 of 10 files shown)


azebich (approved):

vedrani (approved):
Pls add test into Tooltip/test to test Pf-2245 behaviour to ensure it's not broken after this fix. I checked locally and it's ok.

Image

Also rebase again.

PR #5068, opened 2026-07-27
[PF-2287] Unlock Form package react v19 support

azebich:
[PF-2287]

### Description

Upgrades the final-form ecosystem in `@toptal/picasso-forms` to the versions that support React 19. The versions we were on do not declare React 19 in their peer ranges, which blocks consumers from adopting React 19:

| Package | Before | After |
| --- | --- | --- |
| `final-form` | `^4.20.9` | `^5.0.1` |
| `final-form-arrays` | `3.0.2` | `^4.0.1` |
| `react-final-form` | `^6.5.9` | `^7.0.1` |
| `react-final-form-arrays` | `^3.1.4` | `^5.0.0` |
| `react-final-form-listeners` | `^1.0.3` | `^3.0.1` |

Upstream will not patch the old lines — the request to widen `react-final-form@6.5.9`'s peer range ([final-form/react-final-form#1043](https://github.com/final-form/react-final-form/issues/1043)) was closed in favour of the v7 major. These majors are upstream's Flow-to-TypeScript rewrite and carry no intentional API changes ("There _should_ be **NO BREAKING CHANGES**, but because so much code was touched, I'm bumping to a major version out of precaution"). The upgrade is all-or-nothing because the new majors peer-depend on each other, which is why `final-form` and `final-form-arrays` move too.

The public API of `@toptal/picasso-forms` is unchanged — the re-exported `FieldArray`, `useFieldArray`, `OnChange`, `OnFocus`, `OnBlur` and `ExternallyChanged` surfaces are identical.

**Notes for reviewers**

- **`patches/react-final-form-listeners@3.0.1.patch` (new).** Upstream v3.0.1 has a broken publish: `package.json` declares `"types": "dist/index.d.ts"` but the file ships at `dist/src/index.d.ts`, so TypeScript treats the module as untyped and the build fails with TS7016. The patch corrects that one path. It also lets us drop the `@types/react-final-form-listeners` dev dependency, since the package now ships its own types. To be reported upstream; the patch and its `pnpm-workspace.yaml` entry can be deleted once a fixed version is published.
- **Mutator type signatures.** `final-form` v5 constrains `MutableState`'s second type parameter (`InitialFormValues extends Partial<FormValues>`) and types `Mutator`'s first argument as `any[]`. `setActiveFieldTouched` and `setHasMultilineCounter` were adjusted to match; in the latter the argument tuple moved from the signature into a local cast. Runtime behaviour is unchanged.
- **Checkbox snapshots (3 updated).** v7 changed what `input.value` returns for `checkbox`/`radio` fields: v6 returned the `_value` prop unconditionally (`undefined` for a standalone checkbox, so React omitted the attribute), whereas v7 returns it only when defined and otherwise falls through to the field value. A standalone checkbox's visually-hidden native input therefore now renders `value=""`. Verified as cosmetic — submitted values are byte-identical on both versions (`{ 'single-checkbox': true }` standalone, `{ 'checkbox-group': ['checkbox-value-1'] }` in a group), and checked-state handling is untouched. Consumers with their own snapshots covering a `Form.Checkbox` should expect a one-time regeneration.
- **`NumberInput` test assertion removed.** v7 renders the field twice on mount instead of once, breaking `expect(numberInputMock).toHaveBeenCalledTimes(1)`. Both calls are identical and carry the correct props, so rather than re-pin the test to an upstream render count that can shift again on any minor, the count assertion was dropped. The `toHaveBeenCalledWith` assertion below it still covers what the test is named for — that `min`/`max` reach the underlying component.

**Out of scope.** Widening the `react`/`react-dom` peer range of `@toptal/picasso-forms` itself (still `>=17.0.0 < 19.0.0`). This PR is the prerequisite; declaring React 19 support is a separate decision affecting the whole Picasso release line and will be its own ticket.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2287-form-package-react-support)
- `pnpm install` — no unmet-peer warnings for the final-form family
- `pnpm typecheck` and `pnpm test:unit` — both green locally (full workspace: 306 suites / 1411 tests / 274 snapshots, 0 failures)
- In the deployed Storybook, exercise the behaviour that touches final-form internals most directly, since unit coverage there is thinner:
  - TextArea with a character counter (uses the `setHasMultilineCounter` mutator)
  - `validateOnBlur` flows (uses the `setActiveFieldTouched` mutator)
  - scroll-to-error on submit, and form autosave
  - `OnChange` / `OnBlur` / `OnFocus` listeners, and `FieldArray` add/remove
  - single checkboxes and checkbox groups — verify submitted values and click-to-toggle

### Screenshots

| No visual change — dependency upgrade only; Happo expected clean | No visual change — dependency upgrade only; Happo expected clean |

### Development checks

- [x] Self reviewed

**Breaking change**

[PF-2287]: https://toptal-core.atlassian.net/browse/PF-2287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/picasso-forms/src/NumberInput/test.tsx
+++ b/packages/picasso-forms/src/NumberInput/test.tsx
@@ -29,7 +29,6 @@ describe('Form.NumberInput', () => {
         name: 'test-input',
       })
 
-      expect(numberInputMock).toHaveBeenCalledTimes(1)
       expect(numberInputMock).toHaveBeenCalledWith(
         expect.objectContaining({
           min: '8',
--- a/.changeset/final-form-v5-upgrade.md
+++ b/.changeset/final-form-v5-upgrade.md
@@ -1,0 +1,8 @@
+---
+'@toptal/picasso-forms': patch
+---
+
+### Form
+
+- upgrade the final-form ecosystem to the versions that support React 19: `final-form@^5.0.1`, `final-form-arrays@^4.0.1`, `react-final-form@^7.0.1`, `react-final-form-arrays@^5.0.0` and `react-final-form-listeners@^3.0.1`. These majors are upstream's Flow-to-TypeScript rewrite and carry no intentional API changes; the re-exported `FieldArray`, `useFieldArray`, `OnChange`, `OnFocus`, `OnBlur` and `ExternallyChanged` surfaces are unchanged. `react-final-form-listeners@3.0.1` ships a broken `types` path, corrected by a pnpm patch
+- drop the `@types/react-final-form-listeners` dev dependency — the package now ships its own types
--- a/packages/picasso-forms/package.json
+++ b/packages/picasso-forms/package.json
@@ -56,17 +56,16 @@
     "classnames": "^2.5.1",
     "debounce": "^1.2.1",
     "detect-browser": "^5.3.0",
-    "final-form": "^4.20.9",
-    "final-form-arrays": "3.0.2",
-    "react-final-form": "^6.5.9",
-    "react-final-form-arrays": "^3.1.4",
-    "react-final-form-listeners": "^1.0.3"
+    "final-form": "^5.0.1",
+    "final-form-arrays": "^4.0.1",
+    "react-final-form": "^7.0.1",
+    "react-final-form-arrays": "^5.0.0",
+    "react-final-form-listeners": "^3.0.1"
   },
   "devDependencies": {
     "@testing-library/react-hooks": "^8.0.1",
     "@toptal/picasso-test-utils": "workspace:*",
     "@types/classnames": "^2.3.1",
-    "@types/react-final-form-listeners": "^1.0.0",
     "type-fest": "^4.15.0"
   },
   "sideEffects": false,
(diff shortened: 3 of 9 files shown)


packages/picasso-forms/src/NumberInput/test.tsx
```diff
@@ -29,7 +29,6 @@ describe('Form.NumberInput', () => {
         name: 'test-input',
       })
 
-      expect(numberInputMock).toHaveBeenCalledTimes(1)
```
dulishkovych, line 32: this is either regression or AI just decided that this test is not worth keeping 🤔

dulishkovych (approved):
Checked manually on picasso temploy, looks good to me

packages/picasso-forms/src/NumberInput/test.tsx
azebich: ```NumberInput test assertion removed. v7 renders the field twice on mount instead of once, breaking expect(numberInputMock).toHaveBeenCalledTimes(1). Both calls are identical and carry the correct props, so rather than re-pin the test to an upstream render count that can shift again on any minor, the count assertion was dropped. The toHaveBeenCalledWith assertion below it still covers what the test is named for — that min/max reach the underlying component.```

PR #5091, opened 2026-08-18
Stop Button keeping its focus ring after a mouse click

PiotrRashkouski:
### Description

`createCoreClassNames` pushes **both** `focus-visible:shadow-[0_0_0_3px_rgba(32,78,207,0.48)]` and `focus-within:shadow-[0_0_0_3px_rgba(32,78,207,0.48)]` onto every non-disabled `Button`.

`:focus-within` matches the focused element *itself*, not only an ancestor of one. Chrome focuses a `<button>` on mouse-down, so clicking any Button paints the 3px blue ring and leaves it there until the button blurs — precisely what the `focus-visible` variant sitting next to it exists to prevent.

It reads worst on `ButtonGroup` toggles, where the leftover ring looks like a second, stuck "selected" state next to the genuinely active item. Reported by a Staff Portal user against the Demand Dashboard view toggle:

> when I click the button to switch between `Default View` and `Air Traffic Control`, the button I click remains "selected" after clicking it with a blue outline. Is this by design? I would expect this to appear only during click and then disappear.

#### Why the rule is there

It arrived in f9388aaf4 `[FX-5061] Migrate ButtonControlLabel component`. `ButtonControlLabel` renders `<Button as='label'>` wrapping a `Radio` or `Checkbox`, so the focusable element really is the nested `<input>` and the label genuinely needs `:focus-within` to ring. Adding it to the shared helper gave that one component what it needed and leaked the ring to every other Button.

#### The change

Move the rule from `Button` to `ButtonControlLabel`, where it is needed and correct.

`ButtonControlLabel` is the only `as='label'` Button in the repo, and `Button/styles.ts` was the only place using `focus-within`, so nothing else depends on it. (`FileInput` renders its `<input type='file'>` as a *sibling* of the Button, not inside it.)

| | before | after |
|---|---|---|
| Button — mouse click | ring sticks until blur | no ring |
| Button — Tab | ring | ring |
| ButtonRadio / ButtonCheckbox — Tab | ring | ring |
| ButtonRadio / ButtonCheckbox — disabled | no ring | no ring |

### How to test

- [Storybook preview](https://toptal.github.io/picasso/prs/5091/)
- Open any Button story, click a button with the mouse and move the pointer away — no ring should remain. Previously the 3px blue ring stayed until the button lost focus.
- Tab to the same button — the ring must still appear.
- Open the ButtonRadio / ButtonCheckbox stories and Tab to a control — the label must still ring (this is the case the rule exists for).
- `ButtonGroup` story is the clearest reproduction: click one segment and the old build leaves two "selected"-looking items.

Snapshot fallout is exactly the removal of that one class: of the 55 changed snapshot lines, all 55 dropped `focus-within:shadow-[…]`, and the only two that keep it are `ButtonRadio` and `ButtonCheckbox`, where it moved onto the label.

#### About the Happo diffs

`Happo (Picasso/Storybook)` reports **no diffs** — nothing about static rendering changed. `Happo (Picasso/Cypress)` reports **20**, and that is the fix showing up: the Cypress suite is the one that clicks things, and a trigger button keeps DOM focus after the click, so those screenshots used to carry the stray ring. Removing one `focus-within:` class can only change rendering while a Button holds focus, so every diff should be "3px blue ring gone" and nothing else. They need accepting.

Worth a specific look while accepting: any `ButtonRadio` / `ButtonCheckbox` shot must **keep** its ring — that is the case the rule exists for, and unit snapshots confirm those two are the only ones that still carry `focus-within:shadow-[...]`.

### Screenshots

| Before | After |
| --- | --- |
| Ring persists after a mouse click (see reporter's screenshot above) | No ring after a mouse click; unchanged on Tab |

### Development checks

- [x] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`) — not needed, no new utility classes
- [x] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/), and it's been already discussed with designers at #-base-core — restores the intended `focus-visible` behaviour, no new design decision
- [x] Create `examples` for component — no new component
- [x] Ensure the changed/created components have not caused accessibility issues. [How to use accessibility plugin in storybook](https://github.com/toptal/picasso/blob/master/docs/contribution/accessibility.md). — keyboard focus is preserved on both Button and ButtonControlLabel; only the mouse-click ring goes away
- [x] Self reviewed
- [x] Covered with tests ([visual tests](https://github.com/toptal/picasso/blob/master/docs/contribution/visual-testing.md) included) — new unit tests on `Button` and `ButtonRadio`; Happo covers the rest

**Breaking change**

[FX-5061]: https://toptal-core.atlassian.net/browse/FX-5061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/button-click-focus-ring.md
+++ b/.changeset/button-click-focus-ring.md
@@ -1,0 +1,15 @@
+---
+'@toptal/picasso-button': patch
+---
+
+---
+
+### Button
+
+- remove the `focus-within` ring so a mouse click no longer leaves the button
+  ringed until it blurs; keyboard focus still rings through `focus-visible`
+
+### ButtonControlLabel
+
+- ring the label while the Radio or Checkbox inside it holds focus, keeping the
+  behaviour that previously came from Button
--- a/packages/base/Alert/src/Alert/__snapshots__/test.tsx.snap
+++ b/packages/base/Alert/src/Alert/__snapshots__/test.tsx.snap
@@ -42,7 +42,7 @@ exports[`Alert renders 1`] = `
           >
             <button
               aria-disabled="false"
-              class="base-Button text-lg inline-flex items-center justify-center select-none appearance-none m-0 relative normal-case align-middle transition-colors duration-350 ease-out shrink-0 outline-hidden [[data-component-type="button"]+&]:ml cursor-pointer no-underline hover:no-underline rounded-sm shadow-none focus-visible:shadow-[0_0_0_3px_rgba(32,78,207,0.48)] focus-within:shadow-[0_0_0_3px_rgba(32,78,207,0.48)] border-none text-white visited:text-white hover:bg-[#4269D6] active:bg-[#1A41AB] bg-blue min-w h-6 py-0 px-3"
+              class="base-Button text-lg inline-flex items-center justify-center select-none appearance-none m-0 relative normal-case align-middle transition-colors duration-350 ease-out shrink-0 outline-hidden [[data-component-type="button"]+&]:ml cursor-pointer no-underline hover:no-underline rounded-sm shadow-none focus-visible:shadow-[0_0_0_3px_rgba(32,78,207,0.48)] border-none text-white visited:text-white hover:bg-[#4269D6] active:bg-[#1A41AB] bg-blue min-w h-6 py-0 px-3"
               data-component-type="button"
               data-testid="action-button"
               role="button"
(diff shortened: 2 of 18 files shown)


azebich:
hey @PiotrRashkouski , let's wait with this change a bit because we've released the newest picasso version (+v100) which aren't used yet anywhere. So this change can't be easily used unless you update all picasso packages to the latest version in the repository you need it

PiotrRashkouski:
> hey @PiotrRashkouski , let's wait with this change a bit because we've released the newest picasso version (+v100) which aren't used yet anywhere. So this change can't be easily used unless you update all picasso packages to the latest version in the repository you need it

Ok, is see, thanks!

PR #5069, opened 2026-07-28
[PF-2289] Upgrade picasso-charts to support React v19

azebich:
[PF-2262]

### Description

Upgrade `recharts` from `^2.12.3` to `^2.15.4` in `@toptal/picasso-charts` so the
package can run under React 19. This is one step towards lifting the
`< 19.0.0` React peer cap library-wide (PF-2262); the peer range itself is
**not** touched here.

`recharts@2.12.3` blocked React 19 in two independent ways:

1. **Declared peer range excluded it** — `^16.0.0 || ^17.0.0 || ^18.0.0`.
   Upstream added React 19 in `2.15.0`; every version from 2.12.4 through
   2.14.1 still excludes it, so 2.15.x is the minimum.
2. **Runtime break.** React 19 no longer applies `defaultProps` on function
   components, and 2.12.3 set `defaultProps` on seven of them — `XAxis`,
   `YAxis`, `ReferenceArea`, `ReferenceLine`, `ReferenceDot`, `ZAxis`,
   `ErrorBar`. `LineChart` and `BarChart` use four of these, so defaults such
   as `xAxisId`/`yAxisId: 0`, `orientation`, `type`, `tickCount` and
   `ifOverflow` would silently become `undefined` — axis lookup fails and
   charts render blank or misaligned. In 2.15.4 all seven are class components
   or read a module-level defaults constant, so React 19 applies them
   correctly.

Transitive effects, both resolved by the same bump:

- `react-smooth` 4.0.0 → **4.0.4** (4.0.4 is the first release whose peer range
  admits React 19)
- `react-is` 16.13.1 → **18.3.1** inside recharts' dependency tree

The `findDOMNode` hazard (API removed in React 19) lives only in react-smooth's
`AnimateGroup`/`AnimateGroupChild` via react-transition-group. recharts imports
neither module in 2.12.3 or 2.15.4, so that path is unreachable.

No `@toptal/picasso-charts` API or behaviour change. Deliberately **not** in
scope: widening the `react` peer range, upgrading the repo's own React,
recharts 3.x (breaking major, not needed for React 19), and the unrelated
`d3-array`/`debounce` bumps.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2289-upgrade-charts-package-to-allow-react-v19-support)
- Open the **LineChart** and **BarChart** stories and confirm axes render with
  correct orientation, tick count and labels — a dropped `defaultProps` would
  show up as a missing or misplaced axis.
[...]

PR #5096, opened 2026-08-27
[PF-2262] Improvements for Lift the react 19

vedrani:
[PF-2262]

### Description

Follow-ups from the review sweep of the six merged React 19 sub-PRs (#5068, #5069, #5072, #5074, #5076, #5079), plus the first working pieces of workstream E: a React 19 unit-test harness and a React 19 Storybook mode. **No `react` peer range changes in this PR** — the cap lift itself stays with the PF-2262 sweep.

#### 1. Code fixes from the review sweep

- **`useTransitionStatus`: settle an `appear` transition under StrictMode** (`packages/base/Utils/src/utils/use-transition-status.ts`). StrictMode replays the mount effect (setup → cleanup → setup); the cleanup cancelled the pending settle timer and the replay's `prevIn === inProp` early-return never rescheduled it, so status stuck at `entering` forever and `onEntered` never fired. The effect now tracks the pending settle in a ref and reschedules it on replay — without re-firing the start callbacks (react-transition-group re-runs `performEnter` here and double-fires `onEnter` in dev; we settle without the double-fire, and the previous comment claiming rtg parity was wrong). Regression test added; the suite passes under React 18 and 19
- **Fix accumulating `keyup` listeners** in `RichTextEditorEmojiPicker.tsx` — `removeEventListener` received a fresh arrow function so it never removed anything; every picker open leaked a permanent listener on `document.body`. Pre-existing on master (not introduced by this branch); own changeset
- **Restore the render-count guard** in `packages/picasso-forms/src/NumberInput/test.tsx` as `toHaveBeenCalledTimes(2)` with an explanation: react-final-form v7 registers fields in an effect and re-renders once after mount (v6 registered during render — one render). The assertion had been deleted without a trace
- **`any[]` → `unknown[]`** in both Form mutators (`set-has-multiline-counter.ts`, `set-active-field-touched.ts`) — `unknown[]` still satisfies final-form v5's `Mutator` signature contravariantly and keeps the files `any`-free per AGENTS.md; the cast-at-the-boundary tuple stays
- Comment cleanup in `Slide/styles.ts` (historical narration removed) and `Fade.tsx` (compressed to one line — it guards a deliberate twMerge-ordering deviation)
- Remove a stray `--` scaffolding artifact from `README.md`

#### 2. Changeset corrections (release metadata — two retiers to major, please review these deliberately)

- **`final-form-v5-upgrade.md`: `patch` → `major`.** `react-final-form` is a regular dependency and its context is instance-bound: consumers also depending on `react-final-form@^6` can no longer dedupe with our `^7`, get two instances, and `useForm`/`useField` inside a Picasso `<Form>` throws. The changeset now carries the consumer action, the v7 checkbox/radio `checked` `format` → `parse` change, the `value=""` snapshot regeneration, and the mount double-render
- **`date-fns-v4-upgrade.md`: `@toptal/picasso-calendar` `patch` → `major`.** The old text claimed the peer swap was "lateral" — it is not: `@base-ui/react`'s date-fns peer is *optional* while `react-day-picker@8.x`'s is *required*, so shipping `date-fns@^4` next to it turns a warning into a real `ERESOLVE` for npm 7+/yarn consumers. The changeset now states the remedy (`--legacy-peer-deps` / override until PF-2297), the `isWithinInterval` reversed-interval semantic reachable via public `minDate`/`maxDate`, and clarifies that the `pnpm-workspace.yaml` exception protects only this repo (the file is not published)
- **`transitions-tailwind-migration.md`:** Backdrop `patch` → `minor` (consumer `className` now wins via twMerge — new behavior per the tiering guidelines); disclose that Fade/Slide `onEnter`/`onExited` now receive the DOM node as their types always declared (under rtg's `nodeRef` mode the old runtime actually called `onEnter(isAppearing, undefined)` and `onExited()` — consumers reading argument 0 as the boolean must switch to argument 1); disclose that Fade no longer forces inline `opacity: 1`, so a child's own `opacity-*` class applies while visible; drop the inaccurate "identical callback timing" wording
- **`calendar-react-day-picker-react-19.md`:** add `'@toptal/picasso': patch` — the PR also removed the unused direct `react-day-picker` dependency from the aggregate, which previously reached no changelog
- New `rte-emoji-picker-keyup-listener-leak.md` (patch) for the fix above
- `pnpm-workspace.yaml`: cite **PF-2297** in the react-day-picker>date-fns exception comment so it has a removal trigger

#### 3. React 19 unit-test harness (workstream E groundwork)

`pnpm test:react19` runs the existing jest suite against **React 19.2.8 + @testing-library/react 16** via `jest.react19.mjs` (extends `jest.spec.mjs` with moduleNameMapper rewrites). Mechanics worth knowing:

- npm aliases (`react19: npm:react@^19.2.0`, …) escape the workspace-wide `react: '^18.2.0'` override — overrides match the dependency *name*, not the aliased target
- The aliases live in a new private workspace member **`react19/`**, not root devDeps: with `resolvePeersFromWorkspaceRoot: true`, root-level aliases silently rebind other packages' `react-dom` peers to 19 (observed: 49 lockfile bindings flipped before isolating). In the member, all React-19 peer bindings stay confined to its own importer — the React 18 graph is untouched
- Mappings use `<rootDir>/node_modules/...` absolute paths so every file gets the same hoisted copy (bare specifiers resolved from inside `react19/` find a nested duplicate → dual-React `null.useState` crash)
- `react19/sanity.test.tsx` proves the runtime (versions + StrictMode render); it only runs under the react19 config

Verified: all touched suites green under React 18 (67/67 across Utils/Fade/Slide/Backdrop/NumberInput/EmojiMartPicker) **and** under React 19. First genuine harness finding, deliberately not fixed here: **`base/Slider` fails 2 snapshot tests under React 19** (small rendered-DOM diff) — first entry of the workstream E triage list.

#### 4. Storybook React 19 mode

`pnpm start:storybook:react19` (or `STORYBOOK_REACT_19=1 pnpm start:storybook`) runs the Storybook preview on React 19; plain `pnpm start:storybook` is byte-identical to before — every change is env-gated.

- `.storybook/main.js`: gated aliases that overwrite Storybook's **own** plain-key `react`/`react-dom` aliases (builder-webpack5 injects them in `iframe-webpack.config.js:202`; plain keys shadow `$`-exact ones, so the same keys must be replaced)
- `.storybook/react-dom-19-shim/`: re-exports react-dom 19 with an `18.`-prefixed `version`, because `@storybook/react@6.5` picks its renderer via `version.startsWith('18')` (`render.js:111`) and the legacy branch calls `ReactDOM.render`/`unmountComponentAtNode` — both removed in React 19. Also covers a real SB 6.5 edge hit during verification: `unmountElement` calls legacy `unmountComponentAtNode` even in new-root mode for elements with no tracked root (`render.js:118-123`) — the shim answers `() => false`, which is exactly the legacy semantic for "nothing mounted". Plus `client`/`server`/`test-utils` passthroughs (`react-source-render` in CodeExample imports `react-dom/server`)
- Per-mode webpack `cache.version` — the persistent cache does not key on the alias swap, so without this, toggling the env var serves modules resolved against the other React (observed twice during setup). First run after switching modes is a cold build; warm after
- `fastRefresh` off in 19 mode only — the installed react-refresh 0.11 predates React 19
- The preview logs `[react19] preview running on React <version>` at boot so the active mode is self-evident

Verified live: Button chapter renders fully on 19.2.8 (222 rendered buttons across all example sections, all props-table tabs), Base UI Accordion expands interactively, no console errors beyond the pre-existing noise that also appears on 18. Note: Accordion no longer uses picasso-Collapse (Base UI migration), and Collapse has no story — its remaining rtg/`findDOMNode` exposure surfaces through the jest harness, not Storybook. This mode also doubles as the StrictMode story renderer from the workstream E plan (`reactOptions.strictMode` is already on)

### How to test

- [Temploy](https://picasso.toptal.net/feature/pf-2262-lift-the-react-19-improvements)
- `pnpm test:react19 -- --testPathPattern 'react19/sanity|use-transition-status'` — sanity spec asserts React 19.2.8; the hook suite includes the new StrictMode regression test
- `pnpm test:unit -- --testPathPattern 'use-transition-status|packages/base/Fade|packages/base/Slide|packages/base/Backdrop|NumberInput|EmojiMartPicker'` — same suites on React 18
- `pnpm start:storybook:react19` → console shows `[react19] preview running on React 19.2.8`, Button/Accordion stories render and interact
- `pnpm start:storybook` → unchanged React 18 behavior

### Screenshots

N/A — no visual changes on the React 18 path (code fixes are behavioral/dev-only; everything else is release metadata and opt-in tooling).

### Development checks

- [ ] Self reviewed

**Breaking change**

> The two `major` entries are corrected release metadata for changes already merged on the base branch (#5068's final-form stack, #5076's date-fns v4), not new API breaks introduced here — no codemod applies (nothing renames). Testing the Picasso alpha in StaffPortal is recommended before the feature branch goes to master, tracked with the PF-2262 close-out.

[PF-2262]: https://toptal-core.atlassian.net/browse/PF-2262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jest.react19.mjs
```diff
@@ -0,0 +1,42 @@
...
+// The mappings below rewrite every react / react-dom /
+// @testing-library/react import to that member for this run only
+// (@testing-library/react@16 is the first line whose peers admit React 19;
+// it needs @testing-library/dom@10 alongside).
+//
+// Usage: pnpm test:react19 [jest args], e.g.
+//   pnpm test:react19 --testPathPattern react19/sanity
+const config = {
+  ...baseConfig,
+  roots: [...baseConfig.roots, '<rootDir>/react19'],
```
azebich, line 24: looks like the file import isn't good and should be `<rootDir>/node_modules_react19`

check it locally by `pnpm clean` and then `pnpm install && pnpm test:react19 --testPathPattern react19/sanity`

PR #5099, opened 2026-09-01
[PF-2317] Migrate Collapse to Tailwind from react-transition-group

vedrani:
[PF-2317]

### Description

Removes the last react-transition-group consumer in the kit and closes out workstream C of the React 19 plan (PF-2262). Two commits:

**1. Collapse → Tailwind + `useTransitionStatus` (PF-2317)**

- `Collapse` reimplemented on the shared timer-driven `useTransitionStatus` hook (PF-2145 precedent) — the old `<Transition>` ran without `nodeRef`, so it reached `findDOMNode`, which React 19 removes. This crashed Section's jest suite under the React 19 harness and Table expandable rows on CI
- height engine: status-driven `useIsomorphicLayoutEffect` + `requestAnimationFrame` two-frame handoff, replacing the old 50 ms `setTimeout` hacks (which also never cleaned up); the inner wrapper keeps its natural `clientHeight` while the outer div is collapsed, so measurement stays synchronous
- DOM structure, class strings and inline-style shape preserved byte-for-byte — Section's three snapshots are untouched (parity proof)
- riders from the review sweep: object-form `timeout` now produces a valid per-direction `transitionDuration` (was `[object Object]ms`), unrecognized props/`data-private` now reach the DOM node, `displayName` set, `CollapseProps` exported with the misnamed `FadeProps` kept as a deprecated alias, phantom `@toptal/picasso-shared` dep declared, rtg + `@types/react-transition-group` dropped

**2. Version-aware element refs (PF-2262 C-scope)**

- `getElementRef` reads `props.ref` on React 19+ (where `element.ref` is deprecated with a removal warning on every access) and `element.ref` below (where dev builds warn on `props.ref` instead) — each major warns on the other's location, so a fallback chain was never an option
- `ClickAwayListener` consolidates onto the helper (was a hand-rolled `element.ref` cast) — DatePicker, Dropdown and MenuItem inherit the fix; the helper is now the single element-ref read site in the repo

### How to test

- [Temploy](https://picasso.toptal.net/pf-2317-collapse-tailwind-migration)
- `pnpm test:unit -- --testPathPattern 'packages/base/Collapse|packages/base/Section|get-element-ref|ClickAwayListener'` — React 18 (59 tests green locally, zero snapshot churn)
- `pnpm test:react19 -- --testPathPattern 'packages/base/Collapse|packages/base/Section|get-element-ref'` — Section flips red→green under React 19; the get-element-ref suite asserts the read is warning-free on the running major under both configs
- `pnpm test:setup cypress run --component --spec cypress/component/Table.spec.tsx` — expandable rows exercise `appear` + interaction (6/6 locally)
- real-browser interpolation was traced frame-by-frame (0 → 12 → 91 → 165 → 192 → 200 px across the 400 ms window, then `auto`)

### Screenshots

N/A — visual parity is the goal; Section's Happo stories should show zero diffs.

### Development checks

- [ ] Self reviewed

**Breaking change**

> Note: Base UI `Collapsible` was evaluated as the engine and deferred to PF-2318 — it cannot animate mount-open panels (TableExpandableRow's `appear` contract) and its `transitionend` settle never fires in jsdom.

[PF-2317]: https://toptal-core.atlassian.net/browse/PF-2317

### DIFF
--- a/.changeset/collapse-tailwind-migration.md
+++ b/.changeset/collapse-tailwind-migration.md
@@ -6,7 +6,8 @@
 
 - drop the react-transition-group dependency — the last consumer in the kit; its `Transition` ran without `nodeRef`, so it reached `findDOMNode`, which React 19 removes — and reimplement on the shared timer-driven `useTransitionStatus` hook. Public props are unchanged
 - the height handoff now uses `requestAnimationFrame` instead of an internal 50 ms delay, so the transition starts about 50 ms sooner and pending frames are cleaned up on unmount
-- the object form of `timeout` now sets the CSS transition duration correctly per direction (previously it produced an invalid inline value)
+- the object form of `timeout` now sets the CSS transition duration correctly per direction, including a distinct `appear` duration for the mount transition (previously it produced an invalid inline value)
 - unrecognized props (including `data-private`) now reach the rendered DOM node — previously they were silently dropped
 - add the `CollapseProps` type export and deprecate the misnamed `FadeProps` re-export (kept as an alias, so no action is required yet)
+- prop descriptions are corrected (`children`, `appear` and `onEnter` were stale or wrong) and promoted to JSDoc, so they surface in IDE tooltips and generated docs
 - the `react` peer range stays capped at `< 19.0.0`; lifting that cap library-wide is tracked separately in PF-2262
--- a/.changeset/collapse-tailwind-migration.md
+++ b/.changeset/collapse-tailwind-migration.md
@@ -1,0 +1,12 @@
+---
+'@toptal/picasso-collapse': minor
+---
+
+### Collapse
+
+- drop the react-transition-group dependency — the last consumer in the kit; its `Transition` ran without `nodeRef`, so it reached `findDOMNode`, which React 19 removes — and reimplement on the shared timer-driven `useTransitionStatus` hook. Public props are unchanged
+- the height handoff now uses `requestAnimationFrame` instead of an internal 50 ms delay, so the transition starts about 50 ms sooner and pending frames are cleaned up on unmount
+- the object form of `timeout` now sets the CSS transition duration correctly per direction (previously it produced an invalid inline value)
+- unrecognized props (including `data-private`) now reach the rendered DOM node — previously they were silently dropped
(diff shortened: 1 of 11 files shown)


packages/base/Collapse/src/Collapse/Collapse.tsx
```diff
@@ -1,126 +1,138 @@
...
+      // height 'auto' after the transition supports dynamic content inside
+      setHeight('auto')
+
+      return
+    }
+
+    if (status === 'exiting') {
+      setHeight(measured())
+
+      const frame = requestAnimationFrame(() => setHeight('0px'))
```
azebich:
```
 const frame = requestAnimationFrame(() => {
    // force the from-value to be computed before the to-value is written:
    // on a legacy root this setState lands synchronously, pre-recalc
    void nodeRef.current?.offsetHeight
    setHeight('0px')
  })
```
looks like there are some cases where the exit direction isn't valid although I couldn't verify it manually. So check this suggestion which should also be what MUI's `reflow(node)` helper does

.changeset/collapse-tailwind-migration.md
```diff
@@ -0,0 +1,13 @@
+---
+'@toptal/picasso-collapse': minor
+---
+
+### Collapse
+
+- drop the react-transition-group dependency — the last consumer in the kit; its `Transition` ran without `nodeRef`, so it reached `findDOMNode`, which React 19 removes — and reimplement on the shared timer-driven `useTransitionStatus` hook. Public props are unchanged
```
azebich: recharts → react-smooth@4.0.4 still pulls react-transition-group@4.4.5, and react-smooth uses Transition/TransitionGroup without nodeRef.

will this be a problem for us?

azebich (approved):

packages/base/Collapse/src/Collapse/Collapse.tsx
vedrani: Thanks, will check.

.changeset/collapse-tailwind-migration.md
vedrani: It seems it will not, I checked and we are not using that part of react-smooth and both libraries defining react 19 as possible peer dependency.

PR #5100, opened 2026-09-02
[PF-2262] Allow React 19: peer range, @types/react 19, jest and Cypress suites on both majors

vedrani:
[PF-2262]

### Description

Makes React 19 a validated target for Picasso and lifts the peer cap. Grouped by change — the repo squash-merges, so the commit log is not a reading order:

**1. React 19 test harness hardening**

- `react19-validate.yml`: runs the whole unit suite against the standalone React 19 install from #5096 on every PR. **This PR is its first CI run**, and it is blocking on purpose — a non-blocking gate trains everyone to ignore it
- React 19 renders three things differently that are pure formatting, so one snapshot set now serves both majors instead of every base-ui id becoming noise: `useId` (`_r_1_` vs `:r1:`), the `value=""` React 18 reflects on empty inputs, and the `react.transitional.element` tag pretty-format 29 cannot print. Each normalizer lives in `jest/react-compat/` with a test and a "retire when" note (the README is the handoff list for davinci-qa)

**2. A real React 19 bug**

- React 19 decides whether a `<link>`/`<meta>` can be hoisted into `<head>` **when it mounts**; one mounted with `href={undefined}` stays in the body for good. Favicon loads its icons asynchronously, so on React 19 consumers would have shipped with no favicons. It now renders the links once the URLs exist (React 18 end state unchanged). Any `Page.Head` consumer with async attributes has the same trap — called out in the changeset
- FixViewport's test hand-removed a React-owned hoisted `<meta>` before unmount; it unmounts first now

**3. Test-stack alignment**

- React 19 no longer passes function components the legacy-context `{}` argument. `renderedProps(mock)` in `@toptal/picasso-test-utils` (new, minor) reads the props of every render; the six query-builder/RTE suites use it instead of `toHaveBeenCalledWith(props, {})`
- davinci-qa's JSS snapshot serializer threw on the `null`s React 19 elements expose — guarded locally until the upstream fix lands
- 14 snapshot files lose their `value=""` lines (audited: nothing else changed); Avatar's tests pass an `alt`, since testing-library 16 no longer maps an alt-less `` to `role="img"`

**4. Peer range and types**

- every one of the 88 packages: `react`/`react-dom` peer `>=17.0.0 < 19.0.0` → `^17.0.0 || ^18.0.0 || ^19.0.0` (explicit majors, so React 20 stays opted out until validated the same way); one 88-package `minor` changeset
- `pnpm typecheck:react19` compiles all sources against `@types/react` 19 (also a step in `react19-validate.yml`) — it went from **408 errors to 0** while the `@types/react` 17 build stays green, so consumers on types 17, 18 and 19 all keep compiling. The recurring fixes:
  - `JSX.Element` → `React.ReactElement` (the global namespace is gone in 19; `React.JSX` does not exist in 17) — 308 of the 315 files are the generated `*Responsive.tsx` icons, one identical line each
  - element props declared wherever `cloneElement`/`.props` read them (React 19 defaults `ReactElement` props to `unknown`). The two shapes shared across packages are named once in `@toptal/picasso-shared` and re-exported from `@toptal/picasso`: `IconElement` (`ReactElement<{ className?: string }>`) for the 21 `icon`/`expandIcon`/`image` props in 17 files, and `TransitionChild` for the single child Fade and Slide clone. Notification keeps its own `{ className?; color? }` shape because it also reads `color`; EmptyState gains the `@toptal/picasso-shared` dependency it lacked; `Menu` reads its children as `isValidElement<MenuItemProps>` instead of an inline shape
  - `RefObject<T | null>` for hook and context parameters that only read a ref. Three refs are also passed to a JSX `ref` prop, which `@types/react` 17 rejects for `RefObject<T | null>` (`TS2322`) — they are typed `NullableRefObject<T>`, new in `@toptal/picasso-utils`: the `{ readonly current: T | null }` shape `useRef<T>(null)` returns on every types major, so no ref is cast on either major
  - `Omit<HTMLAttributes, 'content' | 'children'>` where they collided with Picasso props, and `ReturnType<ExoticComponent<P>>` for `OverridableComponent`
- only two runtime deltas, both disclosed: `Form` now calls a function child (react-final-form's contract) and `Dropdown` declares the render-function `children` it already supported — that function receives `{ open }` (design-patterns rule 14; the untyped form passed `{ isOpen }`; `ButtonSplit` is the only in-repo caller and passes its `renderMenuButton` straight through)

**5. Component tests on Cypress 14 and React 19**

- Cypress 13.6 → **14.5.4** (+ `cypress-real-events` 1.15). Cypress 13's `cypress/react` mounts through `ReactDOM.render`, which React 19 removed; 14's uses `createRoot`. The only 14 incompatibility found: `retries.openMode: null` must be a number
- `CYPRESS_REACT_19=1` aliases `react`, `react-dom`, `@testing-library/react` and `@testing-library/dom` to the standalone `react19/` install; `ReactRuntime.spec.tsx` asserts which React major actually mounted, so a broken alias cannot silently test React 18 twice. Scripts: `test:integration:react19[:open]`; CI: `react19-component`, three shards, no Happo (behavior only — the React 18 run owns visuals)
- `experimentalMemoryManagement: true`: with davinci-qa's single-tab run mode, Cypress 14's Chromium ran out of renderer memory around the 27th spec and the run hung rather than failing
- Two spec fixes that React 19 surfaced: Autocomplete's tab helper now waits for the field (React 19 commits the mount asynchronously; `.tab()` does not retry), and Select's "focuses the input" uses `realClick()` for wrapper clicks while the popup is open — Cypress's synthetic click fires blur and click in one task, so the click could see the popup still open and close it instead of reopening (failed locally on macOS on Cypress 13 too; passed on Linux CI by flush timing)
- Cypress 14 pulls `@cypress/request@3.0.10`, whose `qs ~6.14.1` resolves to `qs@6.14.2` (GHSA-q8mj-m7cp-5q26, GHSA-x5fp-wj9c-mxmx, GHSA-4mjr-xmp4-gh2g): a version new to the tree, which Dependency Review rejects. Every Cypress 14.x accepts 3.0.10, so an older 14 does not help; instead a root devDependency pins `@cypress/request@3.0.9` (identical apart from its `qs 6.14.0` pin) and pnpm dedupes Cypress onto it, so the lockfile's `qs` set stays exactly master's. No pnpm config change; the pin goes away with Cypress 15+, whose request client resolves a patched `qs` on its own (rationale also in AGENTS.md)
- Result: 61/61 specs on React 18 and on React 19

**6. Changeset audit**

- tiers follow consumer impact: `@toptal/picasso-date-picker` and the `@toptal/picasso` aggregate go `major` alongside Calendar in the date-fns changeset — they install `@toptal/picasso-calendar`, so the `react-day-picker@8` / `date-fns@4` install-time peer conflict reaches their consumers too; `@toptal/picasso-show-more` and `@toptal/picasso-provider` go `minor` for the behavior changes their own changesets already disclose; ShowMore's superseded peer-cap bullet is dropped (the 88-package changeset owns the range)

**Review round 1** — applied in full, with one substitution: the review proposed widening the three JSX-bound refs to `RefObject<T | null>` and dropping their casts, but that fails the `@types/react` 17 build (`TS2322` at `TreeViewSvg.tsx:54`, reproduced in isolation on 17.0.39 and accepted on 19.2.18) — hence `NullableRefObject<T>` above, which drops the casts without it. The rest of the round: `IconElement`/`TransitionChild`, `isValidElement<MenuItemProps>` in Menu, `ButtonSplit` passing `renderMenuButton` directly, `pnpm react19:install` shared by the five `*:react19` scripts, `toMatchObject` in `RadioOptions.test.tsx`, the `snapshotSerializers`/`files` why-comments in `jest.spec.mjs` and `tsconfig.react19.json`, the plain `davinciResolve.alias` spread in `cypress.config.mjs`, and the workflow header wording

**7. `@testing-library/react-hooks` → `@testing-library/react`** (`Replace @testing-library/react-hooks with @testing-library/react renderHook`)

- 46 hook tests imported `renderHook`/`act` from `@testing-library/react-hooks`, which is archived, peers React ≤ 17 and ran here through the root `react-test-renderer@19` — deprecated in React 19 and slated for removal, so all 46 would break together. `@testing-library/react` ships the same `renderHook(callback, { initialProps, wrapper })` since 13.1, and none of the tests used the helpers RTL lacks (`waitForNextUpdate`, `waitForValueToChange`, `result.error`)
- import swap in 46 files (two hand fixes: RTL's hook result type is `RenderHookResult<…>['result']`, and RTL's `wrapper` must take only `children`, so `SubMenuContextProvider` is wrapped in a closure); `@testing-library/react-hooks` dropped from the four packages that declared it, `@testing-library/react` declared where it was missing, `react-test-renderer` removed from the root along with its `pnpm-workspace.yaml` override
- verified: the ten affected packages pass 136 suites / 642 tests / 81 snapshots on React 18 and on React 19; `not wrapped in act` warnings are unchanged against a pre-swap baseline (9 / 7), the 228 react-test-renderer deprecation warnings on React 19 are gone

### Support matrix after this PR

| Consumer on | Runtime | Types |
| --- | --- | --- |
| React 17 / types 17 | supported, declared (never had a test run — same as before) | unchanged |
| React 18 / types 18 | supported, tested | unchanged |
| React 19 / types 19 | supported, tested | compiles; a variable typed bare `ReactElement` (props `unknown` in 19) no longer satisfies `icon?: IconElement` (`ReactElement<{ className?: string }>`) — inline JSX is fine |

### How to test

- [Temploy](https://picasso.toptal.net/pf-react-19-leftovers-1)
- `pnpm test:unit` — React 18: 314 suites / 1546 tests, 272 snapshots
- `pnpm test:react19` — React 19: 315 suites / 1549 tests, the same 272 snapshots
- `pnpm typecheck:react19` — 0 errors; `pnpm build:package` (types 17) — 0 errors. Every `*:react19` script installs the harness through the shared `pnpm react19:install`
- `pnpm start:storybook:react19` for a manual click-through on React 19
- `pnpm test:integration:react19` — 61 component specs on Cypress 14 + React 19 (`pnpm test:integration` for the React 18 run)
- Review aid: `git diff origin/feature/pf-2262-lift-the-react-19-peer-dep -- '**/__snapshots__/*'` shows only removed `value=""` lines; the 315-file `React.ReactElement` swap is mechanical, the hand-written type fixes are the remaining `packages/**` source changes. Besides Cypress 14 and the EmptyState link, the lockfile diff drops five entries the Cypress 14 upgrade had left orphaned (`@types/node@18.19.17`, `http-signature@1.3.6`, `sshpk@1.17.0`, `tmp@0.2.1`, `undici-types@5.26.5`), pruned by a full `pnpm install`; `pnpm install --frozen-lockfile` passes

### Screenshots

N/A — no visual change; Happo is expected clean.

### Development checks

- [ ] Self reviewed

**Breaking change**

Not a breaking change from this PR's own code: the peer range widens, the types stay consumable on `@types/react` 17–19, and the two runtime deltas are the documented library contracts. The changeset audit does raise `@toptal/picasso` and `@toptal/picasso-date-picker` to `major` — for #5076's date-fns 4 install-time peer conflict, not for anything here.

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


package.json
```diff
@@ -53,8 +53,12 @@
     "test:integration": "pnpm build:package && pnpm test:setup cypress run --component",
     "test:integration:ci": "pnpm build:package && happo-e2e -- pnpm test:setup cypress run --component",
     "test:integration:open": "pnpm build:package && concurrently --names build,test --raw 'pnpm build:package:watch' 'pnpm test:setup cypress open --browser=electron --component'",
+    "test:integration:react19": "pnpm -C react19 install --ignore-workspace && pnpm build:package && pnpm test:setup:react19 cypress run --component",
+    "test:integration:react19:open": "pnpm -C react19 install --ignore-workspace && pnpm build:package && pnpm test:setup:react19 cypress open --browser=electron --component",
+    "typecheck:react19": "pnpm -C react19 install --ignore-workspace && tsc -p tsconfig.react19.json",
```
azebich: Check if this script does need `pnpm build:package`

packages/picasso-forms/src/Input/Input.tsx
```diff
@@ -40,7 +40,7 @@ export const Input = React.forwardRef<HTMLInputElement, Props>((props, ref) => {
     mutators: { setHasMultilineCounter },
   } = useForm()
 
-  const { label, labelEndAdornment, titleCase, ...rest } = props
+  const { label, labelEndAdornment, titleCase, name, ...rest } = props
```
azebich, line 49: do we need to take out the name prop from rest?

packages/picasso-provider/src/Favicon/test.tsx
```diff
@@ -16,25 +16,40 @@ const render: typeof baseRender = ui =>
...
 
   it('renders with environment specified', async () => {
     render(<Favicon environment='staging' />)
 
-    await waitFor(() => {
-      expect(document.querySelectorAll('head > link')).not.toHaveLength(0)
-    })
-
-    expect(document.querySelectorAll('head > link')).toMatchSnapshot()
+    await expectFaviconLinks()
```
azebich: The jest file mock returns test-file-stub for every environment, so this verifies nothing the renders case doesn't. Either spy on getIcons to assert the environment actually selects different icons, or merge the two.

packages/base/Dropdown/src/Dropdown/Dropdown.tsx
```diff
@@ -30,10 +30,10 @@ type StyleProps = {
 
 interface InternalProps
   extends StandardProps,
-    HTMLAttributes<HTMLDivElement>,
+    Omit<HTMLAttributes<HTMLDivElement>, 'content' | 'children'>,
     StyleProps {
   /** Anchor element that opens content on click */
-  children: ReactNode
+  children: ReactNode | ((props: { isOpen: boolean }) => ReactNode)
```
azebich: The type now declares the render-function form; the doc still describes only the element form, and this string is what shows in the Storybook props table.

packages/shared/src/types.ts
```diff
@@ -73,7 +74,7 @@ export interface NamedComponent<P> {
 // TS 5.5 variance change.
 export interface OverridableComponent<P = {}> extends NamedComponent<P> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  (props: P & { [key: string]: any }): JSX.Element | null
+  (props: P & { [key: string]: any }): ReturnType<ExoticComponent<P>>
```
azebich, line 88: worth maybe updating the comment on line 74 for line 77?

azebich (approved):
check if tsconfig.jest.json:11-16  new tests aren't type-checked

```
"include": [
  "packages/**/test.tsx",
  "packages/**/*.test.tsx",
  "packages/**/test.ts",
  "packages/**/*.test.ts"
]
```

The three new `jest/react-compat/*.test.ts` files fall outside every tsconfig, so `pnpm typecheck` skips them. Adding `jest/**/*.test.ts` also covers `react19/sanity.test.tsx`.

packages/picasso-forms/src/Input/Input.tsx
vedrani: Yes, it seems due to @types/react v19 there is type problem.

packages/shared/src/types.ts
vedrani: This is done to satisfy both React 17 and 19

PR #5105, opened 2026-09-07
[PF-2434] Fix TableExpandableRow animation

azebich:
[PF-2434]

### Description

Three defects in `Table.ExpandableRow`:

- the content cell was hardcoded `colSpan={100}` — under `table-layout: fixed` Chrome renders the excess as phantom columns and crushes the real ones
- no collapse animation — the content row unmounted the moment `expanded` turned `false`
- opening stuttered — `Collapse` mounted with `in` + `appear`, so it started at `height: auto`, was reset to `0` on mount, and only then grew

Fix:

- add `colSpan?: number`, defaulting to the previous `100` so existing consumers are unchanged, forwarded to the content `Table.Cell`
- keep the content row mounted while it closes and unmount it in `Collapse`'s `onExited`, so closing animates too
- mount `Collapse` closed and open it one commit later from `useEffect`, so the height always grows from zero. `defaultExpanded` still renders at full height with no transition
- content cell is `h-auto p-0`; the content row drops its bottom border while closed, so a collapsed row is zero height with no hairline

`h-auto` is load-bearing rather than cosmetic: `TableCell` sets `h-10`, and `height` on a `<td>` acts as a minimum, so a mounted-but-closed row would have been 40px tall.

Opening is flipped from `useEffect`, not `requestAnimationFrame` — after a click an rAF flip batches with the mount render, `Collapse` mounts already open and nothing animates.

One behavior narrowing, also called out in the changeset: `defaultExpanded` now skips the transition only for a row already expanded on the first render. It used to also skip the first expansion triggered later by a click, which now animates like any other.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2434-fix-tableexpandablerow-column-animation)
- Storybook → `base/Table` → **Expandable rows**: expand and collapse a row. Both directions animate, and opening no longer jumps to full height before growing. The example now passes `colSpan={6}`.
- **Expandable rows, expanded by default**: toggle the table on — rows appear already expanded, with no transition.
- Collapse a row fully and confirm no leftover 1px line between it and the next row.
- For the `colSpan` defect, add `className='table-fixed'` to the `Table` in either example: on master the columns collapse, with `colSpan={6}` they hold.
- `pnpm test:unit -- packages/base/Table`

### Development checks

- [ ] Self reviewed

**Breaking change**

[PF-2434]: https://toptal-core.atlassian.net/browse/PF-2434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
+++ b/packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
@@ -1,5 +1,5 @@
 import type { ReactNode, HTMLAttributes } from 'react'
-import React, { forwardRef, useRef, useEffect } from 'react'
+import React, { forwardRef, useEffect, useState } from 'react'
 import { Collapse } from '@toptal/picasso-collapse'
 import { type BaseProps } from '@toptal/picasso-shared'
 import { twJoin } from '@toptal/picasso-tailwind-merge'
@@ -7,7 +7,7 @@ import { twJoin } from '@toptal/picasso-tailwind-merge'
 import { TableRow } from '../TableRow'
 import { TableCell } from '../TableCell'
 
-const MAX_COL_SPAN = 100
+const DEFAULT_COL_SPAN = 100
 
 export interface Props extends BaseProps, HTMLAttributes<HTMLTableRowElement> {
   /** Should be valid `<tr>` children such as `Table.Cell`. */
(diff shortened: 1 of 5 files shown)


packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
```diff
@@ -16,51 +16,65 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLTableRowElement> {
...
-  /** Makes the row appear without transition when it is expanded the very first time */
+  /** Skips the opening transition for a row already expanded on first render */
   defaultExpanded?: boolean
 }
 
 export const TableExpandableRow = forwardRef<HTMLTableRowElement, Props>(
   function TableExpandableRow(
-    { expanded = false, stripeEven = false, ...props },
+    {
+      expanded = false,
```
vedrani, line 30: Also `defaultExpanded =  false`?

packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
```diff
@@ -16,51 +16,65 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLTableRowElement> {
...
+          className={className}
+          style={style}
+          stripeEven={stripeEven}
+        >
+          {children}
+        </TableRow>
+        {mounted && (
           <TableRow
-            className={twJoin(className, stripeEven && 'bg-gray-200/[0.32]')}
+            className={twJoin(
```
vedrani: Use twMerge from `@toptal/picasso-tailwind-merge` to avoid collisions with `className`

packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
```diff
@@ -16,51 +16,65 @@ export interface Props extends BaseProps, HTMLAttributes<HTMLTableRowElement> {
...
+    const [open, setOpen] = useState(Boolean(expanded && defaultExpanded))
 
     useEffect(() => {
-      if (!wasExpandedOnce.current && expanded) {
-        wasExpandedOnce.current = true
+      if (!expanded) {
+        setOpen(false)
+      } else if (mounted) {
+        setOpen(true)
+      } else {
```
vedrani, line 50: Should we break it into 2 effects where each effect is for `setMounted` and separate for `setOpen`

vedrani (approved):

packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
azebich: I've seen you did that change in react v19 so it would be a conflict for sure, that's why I skipped it here?

Gtosta96-Toptal (approved):

packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
azebich: I would keep it as one for now but I'll add a follow up ticket for the react v19 PR because we should tweak it with the new baseUI collapse logic there

packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
vedrani: I can easily resolve conflicts later in React 19. I would suggest just to check against current master to avoid problem of collision with className content and follow same patter as other components are using.

azebich: I did the change :)

packages/base/Table/src/TableExpandableRow/TableExpandableRow.tsx
vedrani: Ok, makes sense, I was also optimizing Fade/Slide/Collapse so it will maybe need more time to check it and fix it.

PR #5059, opened 2026-07-23
[PF-1988] Migrate Picasso to Base UI + Tailwind

vedrani:
[PF-1988]

### Description

Lands the **Base UI + Tailwind modernization** of Picasso onto `master`. This swaps Picasso's internals from `@material-ui/core` v4 / `@mui/base` / JSS to **[`@base-ui/react`](https://base-ui.com) + Tailwind CSS**, keeping component public APIs stable wherever possible (with codemods/changeset notes where a break was unavoidable).

**Why now:** `@material-ui/core` v4 is deprecated (no security patches) and was the hard blocker for React 19; the kit also carried two parallel styling systems (JSS + partial Tailwind). Full decision, alternatives weighed (Radix / React Aria / shadcn / full rewrite), and rationale are in **[ADR 22 — Modernizing Picasso onto Base UI and Tailwind](https://github.com/toptal/picasso/blob/feature/picasso-modernization/docs/decisions/22-base-ui-tailwind-modernization.md)**.

#### What's in it

- **~30 components/packages migrated** to `@base-ui/react` + Tailwind: Accordion, Backdrop, Badge, Button, Checkbox, Container, Drawer, Dropdown, FileInput, FormLabel, FormLayout, Grid, Menu, Modal, ModalContext, Notification, OutlinedInput, Page, Popper, Radio, Slider, Switch, Tabs, Tooltip, Typography, Utils — plus `picasso-provider` (v2), `picasso-charts`, `picasso-query-builder`, `picasso-rich-text-editor`.
- **MUI v4 runtime decommissioned** — `@material-ui/core` peer dependency removed; styling is now **Tailwind-only** (no JSS/`createStyles`), with component state expressed through `data-*` variants and consumer `className` always winning via `twMerge`.
- **Where Base UI ships no primitive**, Picasso keeps a thin custom implementation: Popper → `@floating-ui/react` ([ADR 20](https://github.com/toptal/picasso/blob/feature/picasso-modernization/docs/decisions/20-popper-floating-ui.md)); Backdrop → custom portaled `<div>` ([ADR 21](https://github.com/toptal/picasso/blob/feature/picasso-modernization/docs/decisions/21-backdrop-custom-implementation.md)); Tooltip open-timing arbitration in [ADR 23](https://github.com/toptal/picasso/blob/feature/picasso-modernization/docs/decisions/23-tooltip-open-arbitration.md).

#### Consumer impact — 59 changesets (100 major · 15 minor · 57 patch bump lines)

[...]

PR #5106, opened 2026-09-07
[PF] Fix emoji picker clicks falling through in RichTextEditor

vedrani:
### Description

Picking an emoji in the RichTextEditor emoji picker did nothing: the popper closed and the editor kept its text. This is broken on master and on every deploy since the Tailwind migration (#5059).

**Cause.** The picker wrapper always carried `pointer-events-none` and toggled `[pointer-events:all]` on open through `cx`. Both utilities have the same specificity, so stylesheet order decides, and the arbitrary-property rule is emitted three rules before `.pointer-events-none`. The open picker was therefore transparent to clicks: each click landed on the editor underneath, emoji-mart's document-level listener treated it as a click outside and closed the picker, and `onEmojiSelect` never ran. Insertion itself was fine; forcing `pointer-events: auto` on the wrapper in DevTools made the same click insert the emoji.

**Fix.** Compose the wrapper with `twMerge` and toggle `opacity-100 pointer-events-auto` while open, so the hidden-state classes are dropped instead of competing on order. The private `classes` constant goes with it. Public props, the `EmojiPlugin` API, the `#trigger-emoji-picker` id and the closed-state markup are unchanged; only the open picker's class list differs, which is the fix.

**Tests.**

- `RichTextEditorEmojiPicker.test.tsx` (jsdom) asserts the wrapper is hidden and click-through until opened, and carries a single pointer-events class once open.
- `cypress/component/RichTextEditor/EmojiPlugin.spec.tsx` opens the picker with a real click, clicks 👍 inside emoji-mart's shadow root without forcing, and asserts the emoji lands in the editor and in the serialized value. Against the previous code it fails at the click: Cypress reports the emoji button covered by the editor.

**Why nothing caught it.** The plugin's unit test mocks all of Lexical, no Cypress spec covered the picker, and Happo only sees that the open picker looks right.

### How to test

- [Temploy](https://picasso.toptal.net/pf-fix-emoji-picker-rte)
- On the temploy Storybook open Forms → RichTextEditor → Emojis (`?path=/story/forms-richtexteditor--richtexteditor#emojis`): click into the editor, type some text, click 🙂, then click any emoji. Expected: the emoji is inserted at the caret, shows up in the value panel below the editor, and the picker closes. On [master's Storybook](https://toptal.github.io/picasso/?path=/story/forms-richtexteditor--richtexteditor#emojis) the same steps only close the picker.
- `pnpm test:unit -- packages/picasso-rich-text-editor/src/RichTextEditorEmojiPicker`
- `pnpm test:setup cypress run --component --spec cypress/component/RichTextEditor/EmojiPlugin.spec.tsx`

### Screenshots

| Before                                                                  | After                                                  |
| ----------------------------------------------------------------------- | ------------------------------------------------------ |
| Clicking an emoji only closes the picker; the editor still reads `hello ` | The emoji is inserted and the editor reads `hello 👍` |

### Development checks

- [ ] Self reviewed

**Breaking change**

None. The changeset is a `patch` for `@toptal/picasso-rich-text-editor`; no codemod or alpha verification is needed.

### DIFF
--- a/packages/picasso-rich-text-editor/src/RichTextEditorEmojiPicker/RichTextEditorEmojiPicker.tsx
+++ b/packages/picasso-rich-text-editor/src/RichTextEditorEmojiPicker/RichTextEditorEmojiPicker.tsx
@@ -2,7 +2,6 @@
 import React, { useEffect } from 'react'
 import data from '@emoji-mart/data'
 import Picker from '@emoji-mart/react'
-import cx from 'classnames'
 import { Container } from '@toptal/picasso-container'
 import { twMerge } from '@toptal/picasso-tailwind-merge'
 
@@ -76,7 +75,7 @@ export const RichTextEditorEmojiPicker = ({
         // and it wins, so clicks fall through the open picker to the editor
         className={twMerge(
           'absolute top-[34px] left-0 z-10 opacity-0 pointer-events-none',
-          cx({ 'opacity-100 pointer-events-auto': showEmojiPicker })
+          showEmojiPicker && 'opacity-100 pointer-events-auto'
         )}
       >
         <Picker
--- a/packages/picasso-rich-text-editor/src/RichTextEditorEmojiPicker/RichTextEditorEmojiPicker.tsx
+++ b/packages/picasso-rich-text-editor/src/RichTextEditorEmojiPicker/RichTextEditorEmojiPicker.tsx
@@ -4,6 +4,7 @@ import data from '@emoji-mart/data'
 import Picker from '@emoji-mart/react'
 import cx from 'classnames'
 import { Container } from '@toptal/picasso-container'
+import { twMerge } from '@toptal/picasso-tailwind-merge'
 
 import RichTextEditorButton from '../RichTextEditorButton'
 import type { CustomEmojiGroup, Emoji } from '../plugins/EmojiPlugin'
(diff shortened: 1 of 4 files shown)


packages/picasso-rich-text-editor/src/RichTextEditorEmojiPicker/RichTextEditorEmojiPicker.tsx
```diff
@@ -76,10 +71,12 @@ export const RichTextEditorEmojiPicker = ({
...
-        className={cx(
-          classes.emojiPicker,
-          showEmojiPicker && classes.activeOpacity,
-          showEmojiPicker && classes.activePointers
+        // twMerge drops the hidden-state classes when open; leaving
+        // `pointer-events-none` in place would let stylesheet order decide,
+        // and it wins, so clicks fall through the open picker to the editor
+        className={twMerge(
+          'absolute top-[34px] left-0 z-10 opacity-0 pointer-events-none',
+          cx({ 'opacity-100 pointer-events-auto': showEmojiPicker })
```
azebich: we should be able to write this as 
```
className={twMerge(
 'absolute top-[34px] left-0 z-10 opacity-0 pointer-events-none',
 showEmojiPicker && 'opacity-100 pointer-events-auto',
)}
```
 and remove cx from file

azebich (approved):

PR #5110, opened 2026-09-08
[PF-2444] Tweak Modal focus trap

azebich:
[PF-2444]

### Description

Fixes a regression from #5103 (`@toptal/picasso-modal@100.1.3`): opening a
`Select` or `Autocomplete` inside a `Modal` closes its popup immediately —
`[data-picasso-popper]` never becomes visible.

`handleDocumentFocus` is registered on `document` in the **capture** phase, but
both of its guards read `document.activeElement`. During capture-phase `focus`
dispatch `activeElement` has not moved to the incoming element yet, so both
guards test the *outgoing* one. When focus moves into a popper portaled outside
the modal, neither guard recognises it and the trap yanks focus back, closing
the popup.

The guards now take the incoming node, read off `event.target`.

The `input:not([disabled]):not([type="hidden"])` selector from #5103 is **kept
as-is** — it is not the bug, it only unmasked it. Before #5103 the selector
usually matched an `input[type="hidden"]` first and `.focus()` on a hidden input
is a no-op, so the broken guard had no visible effect. The trap has therefore
been inert for many modals; this makes it work for the first time, which may
surface other focus behaviour.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2444-modal-focus-trap)
- Storybook → `Modal` → `Default`, open the modal, click the **State** select.
  The popup must open and stay open.
- `Modal` → `Tooltips` still keeps focus in tooltips.
- Unit: `pnpm test:unit -- packages/base/Modal/src`. The new test
  `leaves focus inside an exempt popup before activeElement moves` reproduces
  the browser ordering (blur to `body`, then a capture-phase `focus` targeting a
  `[data-picasso-popper]`) and fails on `master`.

### Screenshots

No visual change — behaviour only, no Happo diffs expected.

### Development checks

- [x] Double check if `picasso-tailwind-merge` requires major update (check its `README.md`) — no styling change
- [x] Make sure that additions and changes on the design follow [Toptal's BASE design](https://design.toptal.net/) — n/a, no design change
- [x] Annotate all `props` in component with documentation — n/a, no prop change
- [x] Create `examples` for component — n/a, existing `Default` example covers it
- [x] Ensure the changed/created components have not caused accessibility issues — fixes an a11y bug (focus trap fighting popovers)
- [x] Self reviewed

[PF-2444]: https://toptal-core.atlassian.net/browse/PF-2444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/base/Modal/src/Modal/Modal.tsx
+++ b/packages/base/Modal/src/Modal/Modal.tsx
@@ -75,32 +75,22 @@ const focusFirstFocusableElement = (node: Element) => {
   node.querySelector<HTMLElement>(focusableElementsString)?.focus()
 }
 
-const isFocusInsideModal = (modalNode: Element) => {
-  const modalContainsFocusedElement = modalNode.contains(document.activeElement)
+// The incoming element is read off the focus event rather than from
+// `document.activeElement`, which has not moved yet during capture-phase
+// dispatch and would make both guards below test the outgoing element.
+const isFocusInsideModal = (modalNode: Element, focused: Node | null) =>
+  focused !== null && modalNode.contains(focused)
 
-  if (modalContainsFocusedElement) {
-    return true
-  }
-
-  return false
-}
-
-const isFocusInsideExemptPopup = () => {
-  const popupContainers = document.querySelectorAll(exemptPopupContainerString)
-
-  if (popupContainers.length === 0) {
+const isFocusInsideExemptPopup = (focused: Node | null) => {
+  if (focused === null) {
     return false
   }
 
-  const popupContainsFocusedElement = Array.from(popupContainers).some(
-    container => container.contains(document.activeElement)
+  const popupContainers = document.querySelectorAll(exemptPopupContainerString)
+
+  return Array.from(popupContainers).some(container =>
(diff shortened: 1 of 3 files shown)


azebich:

Gtosta96-Toptal (approved):
looks good

packages/base/Modal/src/Modal/Modal.tsx
```diff
@@ -75,32 +75,22 @@ const focusFirstFocusableElement = (node: Element) => {
   node.querySelector<HTMLElement>(focusableElementsString)?.focus()
 }
 
-const isFocusInsideModal = (modalNode: Element) => {
-  const modalContainsFocusedElement = modalNode.contains(document.activeElement)
-
-  if (modalContainsFocusedElement) {
-    return true
+// The incoming element is read off the focus event rather than from
```
vedrani: Remove comment

vedrani (approved):

PR #5070, opened 2026-07-28
[PF-2262] Lift React v19 peer dep cap

azebich:
[PF-2262]

### Description

Makes React 19 a supported, validated target for the whole library and widens every package's peer range to `react`/`react-dom` `^17.0.0 || ^18.0.0 || ^19.0.0` (objective **O2**; spike and plan of record: **PF-2236**). Merges to `master` as one release once the sub-PRs below are all in.

This body describes what actually landed; the original plan text (v9 day-picker, `@types/react` bump, story-based harness) is superseded where the work went a different way — each difference is called out under its workstream.

### What landed

| PR | Scope |
| --- | --- |
| #5067 | ShowMore drops `react-truncate` (React 19 incompatible) for CSS `line-clamp` |
| #5068 | Forms: `final-form@5` + `react-final-form@7` |
| #5069 | Charts: `recharts@2.15` |
| #5071 | `notistack@3.0.2`, `react-helmet-async@3` |
| #5072 | Rich text editor: drops `@emoji-mart/react` for a local wrapper over `emoji-mart@5` |
| #5074 | Calendar: `react-day-picker@8.10.2`, stable day `data-testid`s |
| #5075 | `tailwind-merge@3` |
| #5076 | `date-fns@4` (+ `date-fns-tz@3`) |
| #5079 | Backdrop, Fade, Slide: Tailwind transitions, no `react-transition-group` |
| #5096 | Review follow-ups: version-aware element refs, Favicon under React 19 head hoisting, Provider and test-utils fixes |
| #5099 | Collapse: Tailwind transition, removing the last `findDOMNode` path |
| #5100 | Peer range, `@types/react` 19 typecheck, jest + Cypress suites on both majors, Cypress 14, blocking CI (open) |
| #5102 | Query builder: `react-querybuilder` 6 → 8 (PF-2398, open, in flight) |

### Workstreams — outcome

**A. Peer range** — all **88** packages declare `^17.0.0 || ^18.0.0 || ^19.0.0` for `react` and, where present, `react-dom`. Explicit majors rather than an open floor, so React 20 stays opted out until it is validated the same way. One 88-package `minor` changeset. (#5100)

[...]

PR #5111, opened 2026-09-09
[PF-2262] Absorb react-final-form 7.0.1 regressions and fix DatePicker

vedrani:
[PF-2262]

### Description

Follow-ups to the React 19 branch (#5070) from consuming its alpha in Staff Portal, first on React 18, then on React 19.2.8, then twice more as this branch changed. Everything that broke there was in the `@toptal/picasso-forms` major (`react-final-form@7`, `final-form@5`), plus one Safari-only React 19 crash in TimePicker. The PR fixes the parts Picasso owns, documents the upstream type changes, and works around two `react-final-form@7.0.1` regressions.

#### The five decisions, in short

- **Remounted fields lost the user's edit.** `react-final-form@7.0.1` refills a field from `initialValues` every time it mounts and finds no field state, so hiding and showing a field wiped what was typed, with no error. We keep a throwaway subscriber alive across the mount so final-form holds the state, and drop it as soon as the real field registers. One file, deleted when upstream ships the fix.
- **A checkbox rendered the opposite of its value.** 7.0.1 computes `checked` by running the stored value through `parse`, which is the DOM-to-state direction. A stored `'false'` came back truthy, the box showed checked, and clicking it submitted the wrong value with nothing visibly broken. We restore version 6's rule for a standalone checkbox only. Upstream changed this deliberately, so it is a permanent divergence and we are filing the issue.
- **`FieldArray` follows upstream's narrow subscription.** Our first version kept version 3's subscribe-to-everything default for compatibility. Then we looked: nobody needed it, and the only consumers reading array `meta` are three nested arrays, which is the exact case upstream narrowed it for. A default that protects nobody and slows down the one consumer it hurts is not compatibility, so those three name the key they read instead.
- **`useFormState` changes types, not behaviour.** Its booleans are always filled but typed optional, so every call site defaulted them by hand. We fixed the types and left the runtime untouched. `initialValues` stays optional so the hook agrees with `FormSpy` and `form.getState()` in the same form, rather than being right in one place and different in two others.
- **`DatePicker` dropped a value that arrived late.** Fields now register in an effect, so an autofocused picker received its value a frame after mount, showed an empty input, and submitted `null`. We let an empty input the user never touched accept that value, and read focus from the DOM, because React's focus state lags a render behind the browser.

#### What landed

| Area | Change |
| --- | --- |
| `picasso-forms` types | `FieldArray` and `useFieldArray` are Picasso-typed wrappers over `react-final-form-arrays`. `<FieldArray>` is a valid JSX element on `@types/react` 17 and 18 (upstream declares it as returning `ReactNode`), the item type parameter and `fields.update` are back, and the array field keeps the default subscription `react-final-form-arrays@5` narrowed to `length`, `value` and `error`; an array that reads another `meta` key passes a `subscription`. Both are upstream's runtime behind Picasso's types: `FieldArray/types.ts` declares the props and render props, and the component and the hook are typed re-exports rather than wrappers, so nothing of ours runs between the consumer and upstream. `useFormState` called without a `subscription` is typed as the state a fully subscribed form has: the booleans and records are non-optional. `initialValues` stays optional, as `FormSpy` and `form.getState()` report it, and the hook's runtime is upstream's. With a `subscription`, or when a wrapper forwards one whose presence is only known at runtime, the hook keeps final-form's optional types. `OnChange`, `OnFocus`, `OnBlur` and `ExternallyChanged` have real types now; `react-final-form-listeners@3.0.1` points `types` at a file it does not publish, so consumers had `any`. |
| `picasso-forms` runtime | A field rendered without a `name` throws a descriptive `[picasso]` error in development, where react-final-form 7 threw `Cannot call setIn() with undefined key`. A workaround for final-form/react-final-form#1095 (the fix in #1096 is still open): a field that unmounts and remounts, or that mounts over a value seeded with `form.change()`, keeps that value instead of the initial one that 7.0.1 reseeds. `Field` registers a transient final-form subscriber in the layout phase around `useField` and releases it in the next effect, so nothing outlives the field and `destroyOnUnregister` forms are untouched. The claim also covers radios and checkboxes inside their groups. The workaround is one module, `Field/use-claimed-field-state.ts`, to delete once a release contains the fix. |
| `picasso-forms` submit hooks | That claim is the call that creates final-form's field entry, and final-form applies a field's `beforeSubmit`, `afterSubmit`, `data` and `validateFields` only then: for an entry that already exists it re-applies `isEqual`, the validators, `initialValue` and `defaultValue`, and nothing else. So the claim carries those four. Without them react-final-form's own registration added nothing, and its `beforeSubmit`, the wrapper that applies `format()` to a `formatOnBlur` field at submit time, was dropped: such a field submitted its raw value, and a consumer `afterSubmit` never ran. |
| `picasso-forms` checkbox state | `Form.Checkbox`, `Form.ButtonCheckbox` and `Form.Switch` keep version 6's meaning of `checked` for a checkbox that has no `value` of its own. `react-final-form@7` derives `checked` from `parse(value)` where 6 used `format(value)`, so the documented pair `format={value => value === 'true'}` with `parse={checked => (checked ? 'true' : 'false')}` read a stored `'false'` through `parse`, got the truthy `'true'` back, and rendered an unchecked box as checked; clicking it then submitted the wrong value. A checkbox that carries a `value` belongs to a group, where `checked` is array membership, and `Form.Radio` compares `parse(value)` to the radio's own value. Both keep upstream's semantics. |
| DatePicker | The input shows a value that arrives while it is focused and empty. It skips an incoming `value` during focus so as not to overwrite what someone is typing, but an empty input the user has not typed into has nothing to protect. react-final-form 7 registers its fields in an effect and delivers the first value after mount, so an autofocused picker kept an empty input while the calendar showed the right date, and the form submitted `null`. Typed text, and an input the user cleared, stay protected for as long as the focus lasts. Two guards close the mount-frame window a test runner can hit: focus is read from the DOM as well as from React state, and a focus change never fills an empty focused input. |
| TimePicker | Safari gets the native `<input type="time">` like every other browser. The Safari-only `react-input-mask` branch called `ReactDOM.findDOMNode`, which React 19 removed, and threw on mount. `react-input-mask` and `detect-browser` are no longer dependencies of the package. |
| Release notes | The picasso-forms changesets describe the type changes as consumer actions (`FormProps` index signature, optional `initialValues` and state booleans, `FieldArray` generics) and the upstream behaviour that shifts tests (two renders on mount, synchronous `FieldArray` mutators, the narrowed `FieldArray` subscription). |

#### Design decisions worth a reviewer's time

- The claim stays transient. Holding the placeholder subscriber for the life of the form also fixes the remount bug, and final-form clears a field's error only when its last subscriber goes, so a hidden `required` field would then block submit with nothing on screen to explain it. The claim is released in the effect after `useField`.
- `FormSpy` and `Form` are the upstream components. An earlier alpha wrapped `FormSpy` and overloaded `Form`'s function child on `subscription`. Measured on Staff Portal, that did not pay for itself: no `Form` has a function child there, and 4 of 46 `FormSpy` sites read a key the typing made non-optional. The cost was a fork of upstream's render dispatch, a `FormSpy as jest.Mock` compile break in 17 files, and two exported types that could not be spread back into the components. The `useFormState` hook keeps the non-optional typing for the booleans and records as a type-only wrapper; `initialValues` stays optional, as `FormSpy` and `getState()` report it, so Staff Portal keeps its `?? {}` guards at the 22 call sites that read it.
- The `FieldArray` default subscription is upstream's, not version 3's. An earlier revision kept the old full subscription for compatibility. Then we looked at what consumers actually read: in Staff Portal the only array render props touching `meta` are three hooks in the recording-audits schema form, they read one key, `submitError`, and they nest three levels deep, which is the case upstream narrowed the default for. A default that protected nobody and taxed the one consumer the fix was written for is not compatibility, so the wrappers follow upstream and those three hooks name `submitError` explicitly.
- The checkbox fix is scoped to a custom `format`. The default `format` and `parse` agree on every value, so restricting the change to a checkbox that supplies its own `format` leaves every other checkbox on upstream's path.

#### Validation

- Picasso, this head (`ec3022416`): Static checks green, so `pnpm typecheck` and the React 18 jest suite both ran. The three heads before `371c9b9bb` never reached jest, because typecheck failed first in the same job. React 19 unit tests and `pnpm typecheck:react19` green. Two of the three React 19 Cypress shards green; the third was cancelled after 40 minutes inside `Carousel.spec.tsx`, the shard-level hang this repo also hits on unrelated specs, not a test failure. Happo Storybook reports no diffs; Happo Cypress reports five, all hover or width variants of `Button`, `PageTopBarMenu` and `Tabs`, none of which this PR touches. Locally `Form.spec.tsx` and `DatePicker.spec.tsx` pass 20 of 20 in Electron. Every regression test added here fails without its fix, checked by disabling each in turn; the two DatePicker tests were written first and watched fail.
- Staff Portal, React 18 (toptal/staff-portal#16703): validated on the `ec3022416` alpha with the topkit `132b4a9e` line, rebased on master, with the lockfile pinned to a single Picasso suffix. Alpha tags compare the commit sha as a string, so earlier heads had installed two Picasso trees and their numbers did not count. Typecheck reports 0 errors and the unit suites of every package consuming picasso-forms pass; the purchase-order and rubric-editor Cypress specs pass locally. Its specs found three of the defects fixed here (`PurchaseOrderEditModal`, `PurchaseOrderLineDetails`, `DeliveryModelModal` / `CrowdARButton` / `PreviewOutreachEmail`). The narrowed `FieldArray` subscription needed one change there: the three nested schema-form arrays now name `submitError`.
- Consumer tests that count renders have to change: a field registers in an effect, so a component calling `useField` renders twice on mount and its children render twice with it. That is upstream's behaviour, measured with react-final-form's own hook and no Picasso in the path. Staff Portal had six such assertions; they now assert what rendered rather than how often.
- React 19 Staff Portal validation is toptal/staff-portal#16704 and is not summarised here.

### How to test

- [Temploy](https://picasso.toptal.net/pf-2262-react-19-improvements-2)
- `pnpm test:unit -- packages/picasso-forms` and `pnpm test:react19 -- packages/picasso-forms`
- In Storybook, under Forms: edit a field, hide it and show it again (a conditional field or a `RadioGroup`); the edit stays. `useFormState` consumers keep guarding `initialValues` and need no `Boolean()` on the state booleans.
- Open the TimePicker story in Safari: the native time control renders, with no console error on React 19.

### Screenshots

No visual change on Chrome or Firefox. On Safari, TimePicker renders the native time control where the masked text field was.

### Development checks

- [x] Self reviewed

**Breaking change**

Yes, `@toptal/picasso-forms`, which is already major on this branch for the final-form 5 upgrade. Each changeset states the consumer action.

- [x] codemod is created and showcased in the changeset (not needed: the breaks are type-level, and each changeset gives the replacement)

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>


packages/picasso-forms/src/FieldArray/types.ts
```diff
@@ -0,0 +1,78 @@
...
+  /** Array-level validation; return an error, an array of per-item errors, or nothing */
+  validate?: (
+    value: FieldValue[] | undefined,
+    allValues: Record<string, unknown>,
+    meta: FieldState | undefined
+  ) => unknown
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface FieldArrayProps<FieldValue = any>
```
azebich, line 50: should we mention that there is a change regarding types this carries?

- `FieldArrayProps` no longer carries react-final-form-arrays@3's `[otherProp: string]: any` index signature, so extra props passed through `<FieldArray>` to its render component now error — read them from the enclosing scope instead (they still reach the render component at runtime)

packages/picasso-forms/src/Field/test.tsx
```diff
@@ -0,0 +1,414 @@
...
+import { Button } from '@toptal/picasso-button'
+import type { FormApi } from 'final-form'
+import arrayMutators from 'final-form-arrays'
+
+import { FormCompound as Form } from '../FormCompound'
+import { FieldArray } from '../FieldArray'
+import { scrollTo } from '../utils/scroll-to'
+
+jest.mock('../utils/scroll-to', () => ({ scrollTo: jest.fn() }))
+
```
azebich, line 13: warning Arrow function too many lines here

packages/picasso-forms/src/FormState/test.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+      expect(screen.getByText('{"submitting":false}')).toBeInTheDocument()
+    })
+  })
+
+  describe('types', () => {
+    // Type-checked by `pnpm typecheck:react19`: tsconfig.react19.json lists
+    // this file under `files` and builds the packages first. `pnpm typecheck`
+    // cannot resolve package imports without a build, so @types/react 17 is
+    // not checked.
```
azebich: looks like the reason for this is not correct, please double check.

```
// Type-checked by `pnpm typecheck:react19`, which lists this file in
// tsconfig.react19.json's `files`
```
maybe we can do it in a follow up if needed?

packages/picasso-forms/src/Field/Field.tsx
```diff
@@ -131,11 +151,31 @@ const Field = <
...
+  // string-boolean pair `format={value => value === 'true'}` with
+  // `parse={checked => (checked ? 'true' : 'false')}` reads a stored `'false'`
+  // through `parse`, gets the truthy `'true'` back and renders an unchecked box
+  // as checked; clicking it then submits the wrong value. For a checkbox
+  // without its own `value` the field's `input.value` is already
+  // `format(value)`, so this restores the 6.x meaning: a deliberate divergence,
+  // kept until upstream settles that case. A checkbox that carries a `value`
+  // belongs to a group, where `checked` is array membership and upstream's
+  // semantics stand.
+  // TODO: [PF-2262] link the upstream issue for the string-boolean checkbox
```
azebich: should we remove this todo if it's done through this ticket? (same as PR name)

packages/picasso-forms/src/Field/use-claimed-field-state.ts
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * The claim is also not re-established when a field's `data`, `defaultValue`
+ * or `initialValue` prop changes identity while mounted: react-final-form lists
+ * them as dependencies of its registration effect, so such a change unregisters
+ * and re-registers the field, and 7.0.1 reseeds it again. Picasso's wrappers
+ * forward those props unchanged, so a consumer passing a fresh object or array
+ * on every render re-registers the field on every render, with or without this
+ * module; that case keeps upstream's behaviour.
+ *
+ * TODO: [PF-2262] delete this file and its calls once a `react-final-form`
```
azebich: same comment for todo

tsconfig.react19.json
```diff
@@ -30,7 +32,14 @@
   },
   // The sanity test is the one test written against @types/react 19; `files`
   // is not subject to `exclude`, so it is checked while the rest stay out.
-  "files": ["react19/sanity.test.tsx"],
+  // Test files carrying type-level assertions that must stay true against
+  // @types/react 19 as well; `files` is not subject to `exclude`.
```
azebich: check if these comments are overlapping

packages/picasso-forms/src/FormState/index.ts
```diff
@@ -0,0 +1,2 @@
+export { useFormState } from './use-form-state'
+export type { DefaultFormValues, FullFormState } from './full-form-state'
```
azebich: looks like we don't need to export `DefaultFormValues ` from here as we're using it directly from the file. Maybe we can use it from here instead of file? wdyt?

packages/base/DatePicker/src/DatePicker/test.tsx
```diff
@@ -356,6 +356,272 @@ describe('DatePicker', () => {
       expect(handleChange).toHaveBeenCalledWith(new Date(2020, 6, 25))
     })
 
+    describe('when the value arrives after the input is focused', () => {
```
azebich, line 359: we could render 1 helper for `<DatePicker ... />` here which would save a lot of space as it's repeated few times. wdyt?

packages/picasso-forms/src/FieldArray/test.tsx
```diff
@@ -0,0 +1,125 @@
...
+    {fields.map((name, index) => (
+      <li key={name}>{fields.value[index].name}</li>
+    ))}
+  </ul>
+)
+
+const getSkills = () =>
+  screen.getAllByRole('listitem').map(item => item.textContent)
+
+describe('FieldArray', () => {
```
azebich, line 39: should we have 1 top level describe per test file?

.changeset/final-form-v5-upgrade.md
```diff
@@ -4,9 +4,15 @@
 
 ### Form
 
-- upgrade the final-form ecosystem to the versions that support React 19: `final-form@^5.0.1`, `final-form-arrays@^4.0.1`, `react-final-form@^7.0.1`, `react-final-form-arrays@^5.0.0` and `react-final-form-listeners@^3.0.1`. The majors are upstream's Flow-to-TypeScript rewrite; the re-exported `FieldArray`, `useFieldArray`, `OnChange`, `OnFocus`, `OnBlur` and `ExternallyChanged` surfaces are unchanged
+- upgrade the final-form ecosystem to the versions that support React 19: `final-form@^5.0.1`, `final-form-arrays@^4.0.1`, `react-final-form@^7.0.1`, `react-final-form-arrays@^5.0.0` and `react-final-form-listeners@^3.0.1`. The type churn is upstream's Flow-to-TypeScript rewrite; the runtime differences this package absorbs or documents are listed below: the checkbox `checked` derivation and the mount reseed, both new in `react-final-form@7.0.1`, and the synchronous `FieldArray` mutators, the narrowed `FieldArray` default subscription and the two renders on mount, which `7.0.0` already had
```
azebich: this looks like a lot for changeset - can it be shorter maybe? :D

azebich (commented):
check if any of the comments can be removed/trimmed as well

packages/picasso-forms/src/FieldArray/types.ts
vedrani: Adde note to changeset to read them from enclosing scope because they are passes into render component at runtime.

packages/picasso-forms/src/Field/test.tsx
vedrani: Should be reduced.

packages/picasso-forms/src/Field/Field.tsx
vedrani: https://toptal-core.atlassian.net/browse/PF-2522

packages/base/DatePicker/src/DatePicker/test.tsx
vedrani: Added `renderLatePicker(props)`

packages/picasso-forms/src/FieldArray/test.tsx
vedrani: Fixed

packages/picasso-forms/src/FormState/test.tsx
```diff
@@ -0,0 +1,153 @@
...
+        return <span>{`${initialValues} ${submitting}`}</span>
+      }
+
+      renderInForm(<Probe />)
+
+      expect(screen.getByText('undefined undefined')).toBeInTheDocument()
+    })
+
+    it('falls back to the optional keys when the subscription is only known at runtime', () => {
+      const Probe = ({ params }: { params: { subscription?: undefined } }) => {
```
azebich, line 135: double check if this test does what it should. From the look it never reaches the thing it should

azebich (approved):
the branch looks good, comment isn't something blocking

PR #5115, opened 2026-09-23
[PF-2262] Extend the remount fix to picasso-forms' final-form exports

vedrani:
[PF-2262]

### Description

#5111 worked around react-final-form 7.0.1 refilling a remounted field from `initialValues` ([react-final-form#1095](https://github.com/final-form/react-final-form/issues/1095)), but only inside Picasso's own fields. picasso-forms also re-exports react-final-form's `useField` and `Field` (as `FinalField`), and react-final-form-arrays' `FieldArray` and `useFieldArray`. Those still dropped the user's edit when they remounted.

Client Portal hit it first. The next-gen wizard's `use-form-local-state` reads its fields through picasso-forms' `useField`, and `MarketplaceNextGenWizard.cy.tsx` fails two of its three tests on the current alpha. The exposure is wide: 248 production files in Staff Portal and 48 in Client Portal import at least one of the four exports.

This PR puts the same claim behind all four.

#### What changed

- `FinalField` and `useField` live in a new `FinalField/` folder and wrap upstream with the claim. Their public types stay upstream's: `useField` is typed as `typeof` upstream's hook, and `FinalField` keeps upstream's three type parameters and still forwards `ref`. It uses picasso-utils' `forwardRef` and `documentable`, the repo's pattern for generic `forwardRef` components. React's own `forwardRef` does not compile here on `@types/react` 19: it hands the render function `Omit<FieldProps, 'ref'>`, and `Omit` over the index signature in react-final-form's `FieldProps` loses `name`.
- `FieldArray` and `useFieldArray` hold the claim as well and keep their Picasso types.
- The claim moved from `Field/` to `FinalField/use-claimed-field-state.ts` and is one hook now instead of a claim and release pair. All layout effects of a commit run before any passive effect, so the claim's position relative to upstream's hook does not matter. Only its release has to run after upstream's mount effect, and calling the hook after upstream's, or from a parent of the component that calls it, guarantees that. Picasso's `Field` uses the wrapped `useField` and drops its hand-written pair.
- Outside a `Form`, each export throws upstream's own message: `useField must be used inside of a <Form> component`, or `useFieldArray` for the arrays.

#### Decisions worth a reviewer's time

- Grouped `Form.Radio` and `Form.Checkbox` stay on react-final-form's plain `Field`, and `InputField` stays on its plain `useField`. Claims run child-first in the layout phase, in the same order the real registrations follow in the passive phase. A claim on each grouped radio would create the field entry before the group's claim, with the radio's empty config. final-form applies `data`, `validateFields`, `beforeSubmit` and `afterSubmit` only when it creates the entry, so the group's config would be lost. The group's claim already holds the entry for its children, and `InputField` renders a `Field` that holds one.
- The claim carries no `isEqual`. final-form 5 re-applies it on every registration, unlike those four keys.

Still not covered, as before: `Field` and `useField` imported from `react-final-form` itself, `FieldArray` and `useFieldArray` from `react-final-form-arrays`, and a field whose `data`, `defaultValue` or `initialValue` prop changes identity while mounted. [PF-2522](https://toptal-core.atlassian.net/browse/PF-2522) tracks deleting the workaround once upstream ships the fix, and now lists the wrappers.

#### Validation

- Four new remount tests, one per export. Each fails against the raw upstream export (a copy of the test importing upstream directly) and passes with the wrapper. The `ref` test fails when the wrapper stops forwarding.
- picasso-forms jest: 107 of 107 on React 18 and on React 19. The package builds clean on the React 17 types and `tsc -p tsconfig.react19.json` is clean. `FinalField/test.tsx` joins that config's `files`, so CI type-checks the wrappers' public types. Lint adds no warnings, madge finds no cycles, and the Cypress form specs pass 10 of 10.
- Neither app mocks `react-final-form` itself, the one setup where the claim could run against a mocked upstream. Their `@toptal/picasso-forms` mocks (269 in Staff Portal, 32 in Client Portal) either replace an export, which skips the claim, or keep the real one, which behaves as it does in production.
- Staff Portal and Client Portal on this PR's alpha: pending. The Client Portal wizard spec is the acceptance check.

### How to test

- Temploy
- `pnpm test:unit -- packages/picasso-forms` and `pnpm test:react19 -- packages/picasso-forms`
- On this PR's alpha in Client Portal, run `MarketplaceNextGenWizard.cy.tsx`, or step forward through the next-gen wizard and back: the earlier steps keep their values.

### Screenshots

No visual change.

### Development checks

- [x] Add changeset according to [guidelines](https://github.com/toptal/picasso/blob/master/docs/contribution/changeset-guidelines.md) (if needed): the unreleased `forms-keep-field-alive` changeset now lists the four exports
- [x] Self reviewed

**Breaking change**

No. The four exports keep their public types, and the changeset stays a patch.

<details>
<br />

</details>

<details>
<br />

#### When to approve? ✅

3. Code is dirty.

3. You have questions.

</details>

[PF-2522]: https://toptal-core.atlassian.net/browse/PF-2522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/picasso-forms/src/FormListeners/form-listeners.ts
```diff
@@ -37,23 +40,41 @@ export interface ExternallyChangedProps {
...
+// state the same way a field does
+const keepListenerFieldState = <Props extends { name: string }>(
+  Listener: ComponentType<Props>,
+  displayName: string
+) => {
+  const KeptListener = (props: Props) => {
+    useKeptFieldState(props.name)
+
+    return createElement(Listener, props)
+  }
```
azebich: there might be an edge case where when an listener is placed before fields hook or when FinalField is a group child. Double check that everything is as it should be

packages/picasso-forms/src/FinalField/FinalField.tsx
```diff
@@ -0,0 +1,31 @@
...
+import React from 'react'
+import type { FieldProps } from 'react-final-form'
+import { Field as FinalFormField } from 'react-final-form'
+import { documentable, forwardRef } from '@toptal/picasso-utils'
+
+import { useKeptFieldState } from './keep-field-state'
+
+/** react-final-form's `Field`, keeping its value when it remounts */
+export const FinalField = documentable(
+  forwardRef(
```
azebich: do we need to name the forwardref function?

azebich (commented):
nothing blocking as I can see, trim comments where possible

azebich (approved):

